PR #7, opened 2018-06-05
Adding timing and other execution info to ActionResult

ola-rozenfeld:
This change was originally proposed by Daniel Wagner-Hall. In the proposal, ActionResult will contain the metadata of the corresponding execution so that it is available on cached actions as well.

### DIFF
--- a/build/bazel/remote/execution/v2/remote_execution.proto
+++ b/build/bazel/remote/execution/v2/remote_execution.proto
@@ -612,6 +612,12 @@ message ExecutedActionMetadata {
 
   // When the worker completed executing the action command.
   google.protobuf.Timestamp execution_completed_timestamp = 8;
+
+  // When the worker started uploading action outputs.
+  google.protobuf.Timestamp output_upload_start_timestamp = 9;
+
+  // When the worker finished uploading action outputs.
+  google.protobuf.Timestamp output_upload_completed_timestamp = 10;
 }
 
 // An ActionResult represents the result of an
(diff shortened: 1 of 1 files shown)


build/bazel/remote/execution/v0.2/remote_execution.proto
```diff
@@ -580,6 +581,33 @@ message Digest {
...
+
+  // When the worker finished fetching action inputs.
+  google.protobuf.Timestamp input_fetch_completed_timestamp = 6;
+
+  // When the worker started executing the action command.
+  google.protobuf.Timestamp execution_start_timestamp = 7;
+
+  // When the worker completed executing the action command.
+  google.protobuf.Timestamp execution_completed_timestamp = 8;
+}
```
dgassaway: Should we also include start/completed timestamps for the output upload phase of the execution?

ola-rozenfeld: Good idea, thank you!

build/bazel/remote/execution/v0_2/remote_execution.proto
```diff
@@ -580,6 +581,39 @@ message Digest {
   int64 size_bytes = 2;
 }
 
+// ExecutedActionMetadata contains details about a completed execution.
+message ExecutedActionMetadata {
+  // The name of the worker which ran the execution.
+  string worker = 1;
+
+  // When was the action added to the queue.
+  google.protobuf.Timestamp queued_timestamp = 2;
```
alercah: Would this make more sense as a key-value map? Not all systems may expose the same metrics. We could provide a standard list of keys if desired.

ola-rozenfeld: These particular metrics seem general enough, though. Also, the map would be much heavier on the wire. How about we add a map for server-specific times later if/when we need it?

PR #10, opened 2018-06-07
Put Action messages into the CAS.

alercah:
Experience has shown that having access to a copy of the Action message
available makes debugging much easier, as not doing so risks the client
simply throwing it away. It also results in duplication, with the same
or similar Actions being sent repeatedly, wasting bandwidth.

Putting the Action messages into the CAS ensure that they will remain
accessible, and additionally allows servers to perform access control on
UpdateActionResult based on the contents of an Action (for instance, to
prevent an individual user from poisoning actions that might be executed
by a continuous integration system).

If #9 is integrated, the largest performance advantages will likely be lost, but the Action being in the CAS is still valuable. In the case of Execute, we could choose not to impose this requirement on the client, but just expect a server that wishes the Action to be in the CAS to put it there on its own. This would create a discontinuity in the API where clients executing directly do not need to upload the Action, but clients using the results cache directly would have to do so, or we would have to include the Action message inline in the UpdateActionResultRquest. Requiring the client to always upload the action to the CAS leads to the cleanest API, and there will still be some savings on repeated Actions (such as those executed with `do_not_cache` set) to offset much of the additional cost. I hope to release a more detailed analysis soon, based on our observed usage data in RBE.

Theoretical performance analysis/tradeoffs [document](https://docs.google.com/document/d/1ECd0I7aen8CCk3-RUuBjOsBvRmWcsRMrF9bQftcDzpw/edit?usp=sharing).

### DIFF
--- a/build/bazel/remote/execution/v2/remote_execution.proto
+++ b/build/bazel/remote/execution/v2/remote_execution.proto
@@ -42,12 +42,12 @@ service Execution {
   // Execute an action remotely.
   //
   // In order to execute an action, the client must first upload all of the
-  // inputs, as well as the
-  // [Command][build.bazel.remote.execution.v2.Command] to run, into the
+  // inputs, the
+  // [Command][build.bazel.remote.execution.v2.Command] to run, and the
+  // [Action][build.bazel.remote.execution.v2.Action] into the
   // [ContentAddressableStorage][build.bazel.remote.execution.v2.ContentAddressableStorage].
-  // It then calls `Execute` with an
-  // [Action][build.bazel.remote.execution.v2.Action] referring to them.
-  // The server will run the action and eventually return the result.
+  // It then calls `Execute` with an `action_digest` referring to them. The
+  // server will run the action and eventually return the result.
   //
   // The input `Action`'s fields MUST meet the various canonicalization
   // requirements specified in the documentation for their types so that it has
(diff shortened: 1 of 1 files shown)


build/bazel/remote/execution/v0.2/remote_execution.proto
```diff
@@ -760,15 +767,16 @@ message ExecuteRequest {
...
-  // The action to be performed.
-  Action action = 2;
-
   // If true, the action will be executed anew even if its result was already
   // present in the cache. If false, the result may be served from the
   // [ActionCache][build.bazel.remote.execution.v0_2.ActionCache].
   bool skip_cache_lookup = 3;
 
-  reserved 4, 5; // Used for removed fields in an earlier version of the API.
+  reserved 2, 4, 5; // Used for removed fields in an earlier version of the API.
```
ola-rozenfeld: I suggest removing all these from this version. Maybe in a different patch.

build/bazel/remote/execution/v0.2/remote_execution.proto
```diff
@@ -760,15 +767,16 @@ message ExecuteRequest {
...
+
+  // The digest of the [Action][build.bazel.remote.execution.v0_2.Action] to
+  // execute.
+  Digest action_digest = 6;
```
ola-rozenfeld: We also need to update the comment on UpdateActionResult to say that the server may require the action and its part to be uploaded into the CAS.

alercah: I did, it's above.

build/bazel/remote/execution/v0.2/remote_execution.proto
alercah: I think it would be good to do in in a different patch, yes. If we're comfortable saying that these are Completely Different Protos and should never be interchanged, then I think it's fine to renumber fields or at least reuse numbers.. I am comfortable saying that personally but it should probably be done when all the dust is settled anyway.

PR #11, opened 2018-06-08
Adding Results Cache Policy To Execution Requests.

ola-rozenfeld:
Not all action results and CAS blobs are born equal. Some services will have overall space constraints, and throwing out CI execution results in favor of more recent developer build results would not be a good idea for overall performance. Meaning, some implementations may want to granulate cache policy on a per execution basis, and not just on a per-customer basis.

This value will be passed through Bazel via a flag, something like --remote_cache_priority=X, with a default value of 0. The particular value will only make sense for a given server, and even then, not all users may have permissions on setting high cache retention priority for their content.
If a user tries to set a value for which they don’t have permission, the server will return a PERMISSION_DENIED error.

We will soon add a section on a proposed Capabilities API, which will allow, among other things, to fetch the valid priority range from a server for a particular user programmatically.

### DIFF
--- a/build/bazel/remote/execution/v2/remote_execution.proto
+++ b/build/bazel/remote/execution/v2/remote_execution.proto
@@ -774,9 +774,10 @@ message ExecutionPolicy {
 // outputs are stored in the CAS and Action Cache.
 message ResultsCachePolicy {
   // The priority (relative importance) of this content in the overall cache.
-  // Generally, a higher value means a longer retention time or other advantage,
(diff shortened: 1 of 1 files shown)


ola-rozenfeld:
Good question (). That reminds me, we should probably add the ResultsCachePolicy to UpdateActionResult as well (PTAL). On one hand, yes, it could be seen as part of overall ExecutionPolicy; on the other hand, you could say that ExecutionPolicy contains everything related to running the action, and that they are orthogonal. Nesting will look a bit weird, given that for now these are just the priorities.

PR #13, opened 2018-06-11
Adding a Capabilities service.

ola-rozenfeld:
Its goal is to allow clients to configure themselves based on the capabilities of the remote execution service, or at least to give users better actionable error messages when client and server are not compatible.

### DIFF
--- a/build/bazel/remote/execution/v2/BUILD
+++ b/build/bazel/remote/execution/v2/BUILD
@@ -7,6 +7,7 @@ proto_library(
     srcs = ["remote_execution.proto"],
     deps = [
         "@com_google_protobuf//:duration_proto",
+        "@com_google_protobuf//:timestamp_proto",
         "@googleapis//:google_api_annotations_proto",
         "@googleapis//:google_api_http_proto",
         "@googleapis//:google_longrunning_operations_proto",
--- a/build/bazel/remote/execution/v2/remote_execution.proto
+++ b/build/bazel/remote/execution/v2/remote_execution.proto
@@ -279,12 +279,12 @@ service ContentAddressableStorage {
 // The Capabilities service may be used by remote execution clients to query
 // various server properties, in order to self-configure or return meaningful
 // error messages.
+//
 // The query may include a particular `instance_name`, in which case the values
 // returned will pertain to that instance.
 service Capabilities {
   // GetCapabilities returns the server capabilities configuration.
-  rpc GetCapabilities(GetCapabilitiesRequest)
-    returns (GetCapabilitiesResponse) {
+  rpc GetCapabilities(GetCapabilitiesRequest) returns (ServerCapabilities) {
     option (google.api.http) = {
       get: "/v1test/{instance_name=**}/capabilities"
     };
@@ -1107,7 +1107,7 @@ message GetCapabilitiesRequest {
 
 // A response message for
 // [Capabilities.GetCapabilities][google.devtools.remoteexecution.v1test.Capabilities.GetCapabilities].
-message GetCapabilitiesResponse {
+message ServerCapabilities {
   // Capabilities of the remote cache system.
   CacheCapabilities cache_capabilities = 1;
 
--- a/build/bazel/remote/execution/v2/remote_execution.proto
+++ b/build/bazel/remote/execution/v2/remote_execution.proto
@@ -276,6 +276,21 @@ service ContentAddressableStorage {
   }
 }
 
+// The Capabilities service may be used by remote execution clients to query
+// various server properties, in order to self-configure or return meaningful
+// error messages.
+// The query may include a particular `instance_name`, in which case the values
+// returned will pertain to that instance.
+service Capabilities {
+  // GetCapabilities returns the server capabilities configuration.
+  rpc GetCapabilities(GetCapabilitiesRequest)
+    returns (GetCapabilitiesResponse) {
+    option (google.api.http) = {
+      get: "/v1test/{instance_name=**}/capabilities"
+    };
+  }
+}
+
 // An `Action` captures all the information about an execution which is required
 // to reproduce it.
 //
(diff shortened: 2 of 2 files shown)


build/bazel/remote/execution/v0.2/remote_execution.proto
```diff
@@ -269,6 +269,21 @@ service ContentAddressableStorage {
   }
 }
 
+// The Capabilities service may be used by remote execution clients to query
+// various server properties, in order to self-configure or return meaningful
+// error messages.
+// The query may include a particular `instance_name`, in which case the values
```
alercah: Nit: paragraph break should be a blank line, or else the lines should be wrapped together.

build/bazel/remote/execution/v0.2/remote_execution.proto
```diff
@@ -269,6 +269,21 @@ service ContentAddressableStorage {
...
+// returned will pertain to that instance.
+service Capabilities {
+  // GetCapabilities returns the server capabilities configuration.
+  rpc GetCapabilities(GetCapabilitiesRequest)
+    returns (GetCapabilitiesResponse) {
```
alercah: GET requests should return a resource message, not a separate response message (https://cloud.google.com/apis/design/standard_methods).

build/bazel/remote/execution/v0.2/remote_execution.proto
```diff
@@ -994,6 +1009,78 @@ message GetTreeResponse {
...
+
+  // Capabilities of the remote execution system.
+  ExecutionCapabilities execution_capabilities = 2;
+}
+
+// The digest function used for converting values into keys for CAS and Action
+// Cache.
+enum DigestFunction {
+  UNKNOWN = 0;
+  SHA256 = 1;
```
alercah: nit: Should these be SHA_256 and SHA_1 for style?

build/bazel/remote/execution/v0.2/remote_execution.proto
```diff
@@ -994,6 +1009,78 @@ message GetTreeResponse {
...
+  // support multiple instances of the execution system (with their own workers,
+  // storage, caches, etc.). The server MAY require use of this field to select
+  // between them in an implementation-defined fashion, otherwise it can be
+  // omitted.
+  string instance_name = 1;
+}
+
+// A response message for
+// [Capabilities.GetCapabilities][google.devtools.remoteexecution.v1test.Capabilities.GetCapabilities].
+message GetCapabilitiesResponse {
```
alercah: In general, the heavy nesting of messages here makes it a bit difficult for me to follow. Is there a reason not to inline all these messages into the top-level one, except perhaps for PriorityCapabilities?

build/bazel/remote/execution/v0.2/remote_execution.proto
```diff
@@ -994,6 +1009,78 @@ message GetTreeResponse {
...
+
+// Describes the server/instance capabilities for updating the action cache.
+message ActionCacheUpdateCapabilities {
+  bool update_enabled = 1;
+}
+
+// Allowed values for priority in
+// [ResultsCachePolicy][google.devtools.remoteexecution.v1test.ResultsCachePolicy]
+// Used for querying both cache and execution valid priority ranges.
+message PriorityCapabilities {
```
alercah: These are related to #11 and #12; and it strikes me as possibly problematic that they use inverse conventions for priotity (for the cache, higher priority is stronger, but for execution, lower priority is stronger).

ola-rozenfeld: Oh, that was a mistake! I have now inverted the priority for cache to make lower=better for both for consistency.

build/bazel/remote/execution/v0.2/remote_execution.proto
ola-rozenfeld: I don't think so, I've seen this style used with hash function names much more often than not (results may be biased as they are based primarily on internal code searches :-)).

build/bazel/remote/execution/v0.2/remote_execution.proto
ola-rozenfeld: Yes, the service provides remote caching and remote execution and I wanted to separate the properties of each into their own messages, as, for example, a cache-only server will simply not provide any ExecutionCapabilities and that will be clear (I thought). Also, they both have digest function, and it doesn't have to be the same one (there may be more than one for the cache, but just one for execution).
One thing I was not sure in the beginning is having UpdateActionCacheCapabilities as a message -- Eric suggested that (IIRC), because there will probably be more fields related to only that and it would make sense to encapsulate them, and if the message missing, that means no such capability entirely.

build/bazel/remote/execution/v0.2/remote_execution.proto
ola-rozenfeld: Done.

build/bazel/remote/execution/v0.2/remote_execution.proto
ola-rozenfeld: Done.

ola-rozenfeld: Wait, now that I think about it -- should it always be GetX returns X? If so, I'm confused as to how to name this -- the service can't have the same name as the message, right?

alercah:
Based on discussion today, we want to add a version field allowing the server to indicate which version (in particular, minor/patch versions of the API) it supports. My suggestion would be:

```protobuf
message SemVer {
  int32 major = 1;
  int32 minor = 2;
  int32 patch = 3;
  string prerelease = 4;
}
```

I think that the message probably should reside outside the execution API since it will likely be reused if/when more APIs are added to the repository.

ola-rozenfeld:
On second thought, let's add the API version field in a separate PR. It is independent, and I want to discuss the exact placement some more. Also, maybe we want to return the server version as well as the API version?

PR #14, opened 2018-06-11
Making GetTree streaming while still allowing skip ahead.

ola-rozenfeld:


build/bazel/remote/execution/v0.2/remote_execution.proto
```diff
@@ -256,6 +256,10 @@ service ContentAddressableStorage {
   // [ContentAddressableStorage][build.bazel.remote.execution.v0_2.ContentAddressableStorage]
   // (CAS). The server will enumerate the `Directory` tree recursively and
   // return every node descended from the root.
+  // The GetTreeRequest.page_token parameter can be used to skip ahead in
```
alercah: Nit: either include a blank line to separate paragraphs, or wrap the lines together.

build/bazel/remote/execution/v0.2/remote_execution.proto
```diff
@@ -264,7 +268,7 @@ service ContentAddressableStorage {
   // portion present and omit the rest.
   //
   // * `NOT_FOUND`: The requested tree root is not present in the CAS.
-  rpc GetTree(GetTreeRequest) returns (GetTreeResponse) {
+  rpc GetTree(GetTreeRequest) returns (stream GetTreeResponse) {
```
alercah: It occurs to me that calling this GetTree, even though it was intended as a custom method (https://cloud.google.com/apis/design/custom_methods), is confusing with the style for Get*. We should probably rename it while we are here to avoid this.

Issue #16, opened 2018-06-13
Should we be making version 2.0.0?


alercah:
The initial API version we released was labeled as v1test, i.e. a prerelease version of v1. The name was based on Google practice while we were still largely following Google's API processes, representing a pre-release version of v1. But in practice, the API has evolved faster.

I picked v0.2 for the API when publishing this repository because we needed a number and that seemed like a good choice---what [Semver](https://semver.org/) would recommend for the second in-development version. Semver says "If you have a stable API on which users have come to depend, you should be 1.0.0. If you’re worrying a lot about backwards compatibility, you should probably already be 1.0.0." Given the immense amount of dancing around backwards compatibility that we are currently doing, this implies to me that the current API is effectively already 1.0.0.

So I think that perhaps we should be calling the new API 2.0.0, and commit to no major breaking changes. In terms of the proto package path, we should just go with "v2" with no minor number, as unlike with prerelease versions, the minor number changing cannot introduce breaking changes.

PR #17, opened 2018-06-13
Changing version to v2, fixing all references.

ola-rozenfeld:
Fixes https://github.com/bazelbuild/remote-apis/issues/16


alercah:
I think README.md references v0.2 too, can you please update that?

PR #20, opened 2018-07-08
Adding batch downloads method and minor naming tweaks.

ola-rozenfeld:
@alercah did a performance analysis that showed significant improvement when batching is used for uploading small blobs. We suspect that batched downloads will have similar effect. This is primarily intended to be used by the bots, rather than clients, but could be used by the client as well.


ola-rozenfeld:
Note: I suspect we should leave the exact number of MiB allowed for the batching requests out of the API, in general. This is up to gRPC and server limits. I put 4Mib for now, because that is the current gRPC limitation, but we probably should remove the comment.

PR #26, opened 2018-08-22
Fixing OutputFile comment: path is relative to working directory.

### LINKED ISSUE #25: Clarify OutputFile.path and OutputDirectory.path relativeness
In REAPI v1 `OutputFile.path` and `OutputDirectory.path` where both documented as:

 > message OutputFile {
 >   // The full path of the file relative to the input root, [...]

 > message OutputDirectory {
 >   // The full path of the directory relative to the input root, [...]

But now, in REAPI v2, `OutputDirectory.path` documentation changed for :

 > message OutputDirectory {
 >   // The full path of the file relative to the working directory. [...]

There doesn't seem to be any good reason for `OutputFile.path` and `OutputDirectory.path` to be relative to different root, neither there is to depend on a root (working directory) that is define outside of the current message scope.

Any reasons why `OutputDirectory.path` relativeness had been changed?

I think that this modification has been introduced by mistake.

ola-rozenfeld:
Fixes #25.

### DIFF
--- a/build/bazel/remote/execution/v2/remote_execution.proto
+++ b/build/bazel/remote/execution/v2/remote_execution.proto
@@ -821,7 +821,7 @@ message ActionResult {
 //
 // `OutputFile` is binary-compatible with `FileNode`.
 message OutputFile {
-  // The full path of the file relative to the input root, including the
+  // The full path of the file relative to the working directory, including the
   // filename. The path separator is a forward slash `/`. Since this is a
   // relative path, it MUST NOT begin with a leading forward slash.
   string path = 1;


sstriker:
This would actually be a spec change.  I don't think we can chalk this up as a comment fix.

ola-rozenfeld:
It's not a spec change. You can see it being documented correctly in line https://github.com/bazelbuild/remote-apis/blob/master/build/bazel/remote/execution/v2/remote_execution.proto#L431, and the current line just straight-up contradicts that one, because I forgot to change it.

sstriker:
I clearly overlooked something about the proposal on [Paths + Whole directory special case](https://docs.google.com/document/d/12c3oAPgedckLpue2yj0xGgJTEOyUm4mXWWBJ157J-8I/edit#heading=h.sg3la2izb6yt).  I assumed wrongly that absolute pats would refer to the input_root.
For the BuildStream client we have a use case where we do want to retrieve files relative to the input_root.  Shall I file a new issue and proposal for that?

ola-rozenfeld:
But working directory is equal to the input root by default? This is more
related to the https://github.com/bazelbuild/remote-apis/pull/8 change.

sstriker:
@ola-rozenfeld that sounds about right.  Now, to refer to the input_root for output_*, should I open up a new proposal that allows passing of absolute paths that are rooted in input_root?  Or should I pass in ../../../subdir_of_input_root for output_directory?  The latter is legit I think according to spec, but a bit awkward.

ola-rozenfeld:
You are talking about the case where the working directory is a subdir of input root, right? For that one, you pass a ../other_subdir_of_input_root/myoutput path. I think when we discussed "Paths + Whole directory special case" we decided that using absolute paths to refer to something else, like input root, was too confusing.

PR #27, opened 2018-09-18
Fixing comment erroneously mentioning 10MB batch limit.

ola-rozenfeld:


orrsella:
Hi Ola, can you give a bit context on this? Was it ever observed that 10MB is an effective limit or just put as a magic number in place? What’s the current thinking around a limit?

ola-rozenfeld:
Exactly, it was just a magic number. I mean, we did run some experiments that showed batching is a very good idea compared to no batching, but we don't know of any particular number. Hence, mentioning a number in the API was a mistake. We will let it be determined on a per-server basis, and passed back in the max_batch_update_size Capabilities field.

PR #28, opened 2018-09-19
Allowing file symlink outputs.

ola-rozenfeld:
This is a missing case from the existing API: all symlinks that are part of larger
output directories are already supported. However, if the output
file or output directory is itself a symlink, then the current API does not distinguish that, and instead just provides the content Digest.

A motivating case for distinguishing the symlink case is efficiency,
when the client did not want the action to create and download a copy of
an existing file.

TESTED=bazel build

	modified:   build/bazel/remote/execution/v2/remote_execution.proto

### DIFF
--- a/build/bazel/remote/execution/v2/remote_execution.proto
+++ b/build/bazel/remote/execution/v2/remote_execution.proto
@@ -901,8 +901,7 @@ message OutputDirectory {
 
 // An `OutputSymlink` is similar to a
 // [Symlink][build.bazel.remote.execution.v2.SymlinkNode], but it is used as an
-// output in an `ActionResult`. It allows a full file path rather than
-// only a name. It is only used with symlinks pointing to files, not directories.
+// output in an `ActionResult`.
 //
 // `OutputSymlink` is binary-compatible with `SymlinkNode`.
 message OutputSymlink {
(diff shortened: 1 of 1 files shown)


ola-rozenfeld:
@bergsieker, @buchgr, PTAL as well. Thank you! This is a corner case that I think very few people should be using, but at least one customer is using, so I think it's better to add it now.

juergbi:
The changes look good to me. However, I'm a bit confused by your comment but maybe I'm just misreading it:

> symlink output directories are already supported, because the SymlinkNodes are part of the Directory message. Similarly, all file symlinks that are part of larger output directories are already supported as well.

Symlinks **in** output directories indeed work already (independent of the target, i.e., they can point to files, directories or even be dangling). However, if the client specifies 'foo' in 'Command.output_directories' and 'foo' turns out to be a symlink, it won't work. This will also require the changes in this branch and 'foo' specified in 'Command.output_files'.

ola-rozenfeld:
Oh, you're right, sorry! I thought that would work, but you're right, it won't. PTAL.

juergbi:
I was wondering what happens when 'foo' is specified in 'Command.output_files' but it's a symlink to a directory (or dangling). However, I expect this to be very similar to the case where it's specified in 'Command.output_files' but the action creates it as a regular directory. I.e., this shouldn't introduce new (significant) special cases.

In hindsight it might have been simpler to have a single list of output paths for files and directories (and symlinks) in the Command. However, it's probably too late for this now. The proposed changes look fine to me as extension to the current API.

ola-rozenfeld:
> In hindsight it might have been simpler to have a single list of output paths for files and directories (and symlinks) in the Command.

Yeah, that's what it used to be in v0, IIRC, and I remember we discussed this change at length. We thought it's nice to check for these edge cases explicitly on the server whenever we can.

Thank you!

PR #29, opened 2018-10-11
Removing the no-parent restriction in output files and directories

ola-rozenfeld:
Currently, the output files and directories of an action have the following restrictions:
- an output file cannot be a child of an output directory
- an output directory cannot be a child of another output directory

I propose to remove both these restrictions. Reasons:
- In Bazel, it is difficult to generate an output directory and both refer to the entire directory in child action A and refer only to a particular file out of that directory in a child action B. In fact, it's not possible -- the only workaround the situation would be to manually throw these cases out when we build the Command proto.
- This complicates code unnecessarily on both clients and servers! Example of client code complication (in addition to above): it's very hard to check whether an Action produced all the required outputs with the restriction in place. We have to go inside the output directories (and follow symlinks!) to make sure a particular output file or child directory was present -- instead of simply checking in the ActionResult itself.

### DIFF
--- a/build/bazel/remote/execution/v2/remote_execution.proto
+++ b/build/bazel/remote/execution/v2/remote_execution.proto
@@ -438,9 +438,8 @@ message Command {
   // MUST be sorted lexicographically by code point (or, equivalently, by UTF-8
   // bytes).
   //
-  // An output file cannot be duplicated, be a parent of another output file, be
-  // a child of a listed output directory, or have the same path as any of the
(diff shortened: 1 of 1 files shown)


ola-rozenfeld:
@bergsieker, FYI.

ola-rozenfeld:
OK, @buchgr , since you proposed this originally, I assume you're okay with this. I'll go ahead and merge.

PR #30, opened 2018-10-11
Clarify semantics of CancelOperation

ola-rozenfeld:
The Operations API specifies the use of the error field on the Operation proto in case the operation was cancelled. However, for our API, we want to remove all uses of this field, and use the ExecuteResponse.status instead. Therefore, we need to explicitly specify what should happen in case the server implements CancelOperation (which it doesn't have to!) and it was called for the given exeution.


ola-rozenfeld:
@t-chaik, FYI.

ola-rozenfeld:
Also: @buchgr and @bergsieker

PR #45, opened 2019-01-07
Add reserved keywords to instance name segments. Fixes #41.

### LINKED ISSUE #41: Ambiguity extracting instance names from resource names
`remote_execution.proto` describes instance names but there appears to be a small hole in the specification:

```
the `instance_name` is an identifier, possibly containing multiple path segments,
```

There are no further restrictions, so the following would appear to be a valid instance name: `uploads/6c92172c-8064-4351-93a2-640d5e8761fe/blobs/187d384348c73a2c0246a42fd061167039e551c1fe8c24a51d9538f4536fa72c/1034`

In which case, if I sent this resource name:

`/uploads/6c92172c-8064-4351-93a2-640d5e8761fe/blobs/187d384348c73a2c0246a42fd061167039e551c1fe8c24a51d9538f4536fa72c/1034/uploads/6c92172c-8064-4351-93a2-640d5e8761fe/blobs/187d384348c73a2c0246a42fd061167039e551c1fe8c24a51d9538f4536fa72c/1034/uploads/6c92172c-8064-4351-93a2-640d5e8761fe/blobs/187d384348c73a2c0246a42fd061167039e551c1fe8c24a51d9538f4536fa72c/1034`

What is the actual instance name and what is the hash of the object requested? I think the instance name could be either one or two repetitions of the first string I posted.

Because anything after the 'size' field is ignored, you cannot pattern match from the right.

I ask because we have an implementation of a remote execution server, and it attempts to determine an optional instance name from the resource name. I'm wondering how to get this code exactly correct.

ola-rozenfeld:


build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -209,6 +209,8 @@ service ActionCache {
 // by the server. For servers which do not support multiple instances, then the
 // `instance_name` is the empty path and the leading slash is omitted, so that
 // the `resource_name` becomes `uploads/{uuid}/blobs/{hash}/{size}`.
+// To simplify parsing, a path segment cannot equal any of the following
+// keywords: `blobs`, `uploads`, `actions`, `actionResults`, and `size`.
```
bergsieker: I don't think it's necessary to reserve 'size' (because it's only ambiguous when preceeded by 'blobs') but we probably should reserve 'operations' and 'capabilities'.

ola-rozenfeld: Operations, agreed, done. But why capabilities? You think there's a chance they will have a resource name some time? I'm not sure about that.

bergsieker: I think it's unlikely they would end up with a resource name, but given that the string 'capabilities' already appears in the URI, there's at least a chance. I don't think reserving it realistically hurts anyone, and it's easier to drop a restriction in the future than to add one.

ola-rozenfeld: Sure, done.

Issue #43, opened 2018-12-23
Lack of description on what the permissions on files in the input root are


EdSchouten:
(Note: This is a continuation of a remark I made in #40)

It seems to be the case that the protocol doesn't document what the permissions of files and directories in the input root are in relation to the credentials of the build action.

- Are build actions permitted to overwrite input files? When using Sandboxfs, this would be easy to achieve. Without using Sandboxfs, it's also possible, but has the downside that optimizations built around caching input files and hardlinking them is out of the question.
- Related to the previous question, is it permitted to hardlink input files to some other location? Recent versions of Linux have `fs.protected_hardlinks` enabled by default. This implies that if input files are read-only (due to the use of hardlinking caches), we cannot guarantee that the kernel will allow the creation of hardlinks to input files.
- In the general sense: is there even any guarantee that hardlinks can be created between two distinct directories in the input root? For example, may a worker place the output directories on a separate file system (`tmpfs`)? If so, this means you can't hardlink input files into an output directory.
- What are permissions on directories? Where may the build action create temporary files? In any directory in the input root, or only inside of a directory containing one or more outputs?

Buildbarn's workers don't use anything like FUSE (yet!), for the reason that I initially aimed at using Buildbarn on Kubernetes, where you can't simply make mounts inside of containers. After giving it enough tweaks, I eventually concluded that:

- All directories in the input root need to be writable to appease the build rules out there. Build rules should be allowed to rename and remove any file in the input root, and to create files in any directory in the input root.
- Input files may be read-only. They may be replaced by removing them first and creating a new one with the old name.
- Disallowing input files to be hardlinked (`fs.protected_hardlinks == 1`) causes a very small number of build rules to break, but those may be easy to fix.

Do we want to document this in the `.proto` file somewhere?

ola-rozenfeld:
Does https://github.com/bazelbuild/remote-apis/issues/90 fully address this, or is there anything remaining?

juergbi:
> Does #90 fully address this, or is there anything remaining?

We still need to document the default permissions, i.e., when `UnixMode` is not specified as node property.

EricBurnett:
[...]

PR #54, opened 2019-02-01
Document what the intended file permissions of the input root is.

EdSchouten:
At the build event at Bloomberg, several users of REv2 (Bazel,
Buildbarn, BuildGrid) discussed what file permissions of the input root
ought to be. Basically, there were two camps:

- Camp A: reduce the number of writable files and directories to the
  absolute minumum (Bazel and Buildbarn).

- Camp B: keep files read-only to allow caching based on hardlinks,
  while making all directories writable (BuildGrid). This is desired to
  support the execution of very large build actions that invoke
  traditional build systems that tend to clobber the input root all over
  the place.

It was eventually decided to support the latter. In the long run a
separate effort could be started to lock this down further, if it turns
out this causes significant performance overhead.

Until recently, Buildbarn also followed the second approach. A recent
change caused Buildbarn to only make directories with one or more
outputs to be writable and all of the parent directories. This change
will be reverted to match the semantics proposed by this commit.

Fixes: #43

### DIFF
--- a/build/bazel/remote/execution/v2/remote_execution.proto
+++ b/build/bazel/remote/execution/v2/remote_execution.proto
@@ -382,7 +382,7 @@ message Action {
   // by the build action at any place within the input root.
   //
   // In addition to that, the entire input root must be exposed through
-  // a single mount, meaning that calls to the rename() on files and
+  // a single mount, meaning that calls to rename() on files and
   // directories within the input root may not fail due to cross-device
   // links. If the operating system executing the build action supports
   // hard links, the same applies to link(). Whether directories outside
--- a/build/bazel/remote/execution/v2/remote_execution.proto
+++ b/build/bazel/remote/execution/v2/remote_execution.proto
@@ -385,12 +385,15 @@ message Action {
   // a single mount, meaning that calls to the rename() on files and
   // directories within the input root may not fail due to cross-device
   // links. If the operating system executing the build action supports
-  // hard links, the same applies to link().
+  // hard links, the same applies to link(). Whether directories outside
+  // of the input root (e.g., /tmp) reside on the same file system is
+  // unspecified.
   //
   // Due to recent versions of Linux disallowing the creation of hard
   // links to files without having write access to them (sysctl
-  // fs.protected_hardlinks=1), the build action may only be permitted
-  // to create hard links to files that were not part of the input root.
+  // fs.protected_hardlinks=1) and input files may not be writable, the
+  // build action may only be permitted to create hard links to files in
+  // the input root that were created at runtime.
   Digest input_root_digest = 2;
 
   reserved 3 to 5; // Used for fields moved to [Command][build.bazel.remote.execution.v2.Command].
--- a/build/bazel/remote/execution/v2/remote_execution.proto
+++ b/build/bazel/remote/execution/v2/remote_execution.proto
@@ -379,7 +379,7 @@ message Action {
   // input root are available for reading (i.e., at least 'r--' or 'r-x'
   // in terms of UNIX permissions). All directories are available for
   // writing ('rwx'), meaning that files and directories may be created
-  // by the build action at any place within the input root.
+  // and deleted by the build action at any place within the input root.
   //
   // In addition to that, the entire input root must be exposed through
   // a single mount, meaning that calls to rename() on files and
--- a/build/bazel/remote/execution/v2/remote_execution.proto
+++ b/build/bazel/remote/execution/v2/remote_execution.proto
@@ -374,6 +374,23 @@ message Action {
   // directory, as well as every subdirectory and content blob referred to, MUST
   // be in the
   // [ContentAddressableStorage][build.bazel.remote.execution.v2.ContentAddressableStorage].
+  //
+  // The build action may assume that all files that are part of the
+  // input root are available for reading (i.e., at least 'r--' or 'r-x'
+  // in terms of UNIX permissions). All directories are available for
+  // writing ('rwx'), meaning that files and directories may be created
+  // by the build action at any place within the input root.
+  //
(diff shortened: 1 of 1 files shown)


build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -374,6 +374,23 @@ message Action {
...
+  // The build action may assume that all files that are part of the
+  // input root are available for reading (i.e., at least 'r--' or 'r-x'
+  // in terms of UNIX permissions). All directories are available for
+  // writing ('rwx'), meaning that files and directories may be created
+  // by the build action at any place within the input root.
+  //
+  // In addition to that, the entire input root must be exposed through
+  // a single mount, meaning that calls to the rename() on files and
+  // directories within the input root may not fail due to cross-device
+  // links. If the operating system executing the build action supports
```
ulfjack: Maybe note explicitly that the input root does not include /tmp?

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -374,6 +374,23 @@ message Action {
...
+  // hard links, the same applies to link().
+  //
+  // Due to recent versions of Linux disallowing the creation of hard
```
ulfjack: This entire paragraph is unclear to me.

build/bazel/remote/execution/v2/remote_execution.proto

build/bazel/remote/execution/v2/remote_execution.proto
EdSchouten: What do you think of the current phrasing? Is it any better?

ulfjack: LGTM

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -374,6 +374,26 @@ message Action {
...
+  // directories within the input root may not fail due to cross-device
+  // links. If the operating system executing the build action supports
+  // hard links, the same applies to link(). Whether directories outside
+  // of the input root (e.g., /tmp) reside on the same file system is
+  // unspecified.
+  //
+  // Due to recent versions of Linux disallowing the creation of hard
+  // links to files without having write access to them (sysctl
+  // fs.protected_hardlinks=1) and input files may not be writable, the
+  // build action may only be permitted to create hard links to files in
```
EricBurnett: If I'm reading this right, it's a warning about something that may or may not be a constraint? That feels like "too much information" in this context, since there are many other constraints that different execution systems may or may not impose not documented in the API. (Random examples: limits on numbers of inputs, whether or not inputs or input directories may be symlinks, whether or not output symlinks will be "flattened", available disk space).

If there is a minimum requirement here that should be specified (such as "inputs are readable", above), can we pivot the paragraph to that? Otherwise I'd suggest removing it entirely, and documenting it as limitations of a particular implementation rather than API-level limitations.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -374,6 +374,26 @@ message Action {
...
   // be in the
   // [ContentAddressableStorage][build.bazel.remote.execution.v2.ContentAddressableStorage].
+  //
+  // The build action may assume that all files that are part of the
+  // input root are available for reading (i.e., at least 'r--' or 'r-x'
+  // in terms of UNIX permissions). All directories are available for
+  // writing ('rwx'), meaning that files and directories may be created
+  // and deleted by the build action at any place within the input root.
+  //
+  // In addition to that, the entire input root must be exposed through
```
EricBurnett: I have a couple concerns with this paragraph as written.

 1. It's very linux-centric right now. We've defined the API in a system agnostic way (see e.g. handling of platform separators), and I consider that an important requirement to keep. So the above paragraph specifying "files are readable" and "directories are writable" is good, and I think the examples reasonable to keep since they're clearly examples; this one is defined in terms of mounts and rename() and whatnot that are *nix specific though.

Can this be generalized to more universal properties that hold across OS? If not, is it really a system-specific constraint and thus not actually appropriate for the API spec?

 2. Does the API require specifying this at all? Similar to my other comment, I'm not actually sure it's a limitation that all execution systems must comply with.

This gets into somewhat philosophical territory that is not perfectly settled to me yet, but at a general level it's necessary for some execution systems to allow things others do not, and also, for execution systems to have room to optimize within the constraints of the API. Take RBE's environment controls, for example: https://cloud.google.com/remote-build-execution/docs/changing-docker-config (if this 404's for you you'll need to join the https://groups.google.com/forum/#!forum/rbe-alpha-customers group to get access to our Alpha docs). Users may opt in to various environment features, each of which may also come with a performance pessimization or other tradeoffs. I could see doing the same for actions requiring the input FS to be a single mount point, or where all input files are distinct (and not hardlinks of each other), but *not* adhering to those constraints by default.

What do you think - is this restriction a must-have constraint of the API? Or more practically terms, is *not* adhering to it going to lead to execution systems that are incompatible for "common case" actions? That's a minimal bar I think I'd draw: if "most" actions can run across implementations of the API, that seems good, even if there are corner cases where certain actions need to be changed before they can be ported. But if the majority of actions require porting, clearly we haven't done a good enough job defining a core standard.

ola-rozenfeld: Ping?

bergsieker:
I believe this will be largely superseded by https://github.com/bazelbuild/remote-apis/issues/90.

Once we've finalized the more general solution of handling file permissions in the context of that issue, we can loop back here to update what the default expectations are.

bergsieker:
I don't think this is resolvable right now given the requirement to support different platforms with different permissions schemas. NodeProperty allows users to explicitly set permissions if necessary.

PR #55, opened 2019-02-06
Clarify semantics around do_not_cache and skip_cache_lookup

EricBurnett:
Clarify semantics for do_not_cache and skip_cache_lookup fields to cover the semantics of when in-flight actions may be "merged", and to clarify when results are and aren't eligible for caching.

The use-cases of interest I'm considering are:
  - Minimizing total number of executions for idempotent actions. (The default: neither field is set; only one total execution is required)
 - Bypassing and overwriting the Action Cache when its results are stale or corrupt. (skip_cache_lookup should be set and the result should still be cached).
 - Actions with side-effects that should always be re-executed (do_not_cache should be set, multiple in-flight requests should not be merged)
 - Actions intentionally being (re)run, e.g. for performance analysis across multiple attempts. (do_not_cache should be set; multiple in-flight requests should not be merged).

I took a look at Bazel and I believe its usage to be consistent with these semantics:
 - --runs_per_test implies uncacheable: https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/analysis/test/TestRunnerAction.java#L373
 - Outputs found to be missing after an action cache entry is found result in re-execution with skip_cache_lookup set so the action gets re-executed: https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/remote/RemoteSpawnRunner.java#L209

Any sanity-checks that the comments here are clear and understandable would be great, and/or validation that this is consistent with the semantics expected by other client tools.

### DIFF
--- a/build/bazel/remote/execution/v2/remote_execution.proto
+++ b/build/bazel/remote/execution/v2/remote_execution.proto
@@ -396,7 +396,8 @@ message Action {
   // immediately, rather than whenever the cache entry gets evicted.
   google.protobuf.Duration timeout = 6;
 
-  // If true, then the `Action`'s result cannot be cached.
+  // If true, then the `Action`'s result cannot be cached, and in-flight
+  // requests for the same `Action` may not be merged.
   bool do_not_cache = 7;
 }
 
@@ -982,9 +983,19 @@ message ExecuteRequest {
   // omitted.
   string instance_name = 1;
 
-  // If true, the action will be executed anew even if its result was already
-  // present in the cache. If false, the result may be served from the
-  // [ActionCache][build.bazel.remote.execution.v2.ActionCache].
+  // If true, the action will be executed even if its result is already
+  // present in the [ActionCache][build.bazel.remote.execution.v2.ActionCache].
+  // The execution is still allowed to be merged with other in-flight executions
+  // of the same action, however - semantically, the service MUST only guarantee
+  // that the results of an execution with this field set were not visible
+  // before the corresponding execution request was sent.
+  // Note that actions from execution requests setting this field set are still
+  // eligible to be entered into the action cache upon completion, and services
+  // SHOULD overwrite any existing entries that may exist. This allows
+  // skip_cache_lookup requests to be used as a mechanism for replacing action
+  // cache entries that reference outputs no longer available or that are
+  // poisoned in any way.
+  // If false, the result may be served from the action cache.
   bool skip_cache_lookup = 3;
 
   reserved 2, 4, 5; // Used for removed fields in an earlier version of the API.


EricBurnett:
@EdSchouten @juergbi FYI.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -396,7 +396,8 @@ message Action {
   // immediately, rather than whenever the cache entry gets evicted.
   google.protobuf.Duration timeout = 6;
 
-  // If true, then the `Action`'s result cannot be cached.
+  // If true, then the `Action`'s result cannot be cached, and in-flight
+  // requests for the same `Action` may not be merged.
```
EdSchouten: This is hard to realise in the case of Buildbarn.

Buildbarn currently treats the action as a black box. It merely queues an execute request and forwards it to a worker. The worker is responsible for loading the action from the CAS. If this change is pushed through, Buildbarn's scheduler would need to be extended to access the CAS, which it currently doesn't.

Maybe we should move this into the ExecuteRequest to make it more readily available?

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -982,9 +983,19 @@ message ExecuteRequest {
...
+  // that the results of an execution with this field set were not visible
+  // before the corresponding execution request was sent.
+  // Note that actions from execution requests setting this field set are still
+  // eligible to be entered into the action cache upon completion, and services
+  // SHOULD overwrite any existing entries that may exist. This allows
+  // skip_cache_lookup requests to be used as a mechanism for replacing action
+  // cache entries that reference outputs no longer available or that are
+  // poisoned in any way.
+  // If false, the result may be served from the action cache.
   bool skip_cache_lookup = 3;
```
EdSchouten: It may be noteworthy to mention that Buildbarn currently ignores this flag altogether; it behaves as if it's always set to true. We currently assume that if a client actually calls into Execute() it already checked the AC beforehand. Bazel seems to do that at least.

Is it really worth the effort to have this flag?

ulfjack: I think it's a mistake for Bazel to do two calls. This just adds another round-trip for no benefit. I think at some point Bazel will move to doing a single call to the remote executor, and expect to be able to get a cache hit back.

ola-rozenfeld: +1 to Ulf, I'm pretty sure Bazel can already switch to doing the single Execute call and get cache hits back (from RBE), since it already has code in place to do all the uploads if a missing input status is returned. That wasn't the case in the beginning, when we wanted to enforce the order of cache check -> get missing -> upload missing -> execute. But note that while doing that may be a bit simpler, I'm not sure that it will save a round trip often. The saving will only happen when the action is (a) not in the cache, and (b) can be re-executed without any additional uploads. This seems like an unlikely case.
Nevertheless, I think it's important to have the flag to that servers *may* return cached results, or merge actions in flight, as an optimization.

build/bazel/remote/execution/v2/remote_execution.proto
ola-rozenfeld: I recall that we wanted to make it part of the Action in order to ensure that actions that differ on this field alone have different action keys. But now that I think about it, I don't really see the point -- I don't see the different keys preventing any bad scenarios in a way that a field in ExecuteRequest wouldn't. So I wouldn't be opposed to moving this (even though it would be a breaking change, would need to be done in steps, etc).
Am I forgetting some reason?

EricBurnett: Ed, just to make sure I understand: is the semantic clarification that in-flight actions may not be deduped problematic for you, or more generally that some actions are uncacheable at all (as defined within the Action)?

I agree with Ola that moving this field out of the Action would be reasonable, so if its existence is problematic for you that seems fine to me. (But yeah, breaking change). I'd prefer to not block this change on it however if this doesn't make things _harder_ for you.

build/bazel/remote/execution/v2/remote_execution.proto
EricBurnett: Ulf: I think it's equivalent right now (as Ola says) - either:
 GetActionResult (NOT_FOUND) -> upload action & blobs -> Execute, or
 Execute(NOT_FOUND) -> upload actions & blobs -> Execute

You could pre-upload actions & blobs _before_ calling Execute, but you'd only want to do so on incremental compilation paths where you know the action is going to be a miss anyways (at which point whether or not Execute checks the action cache is moot); on clean executions where the action is most likely a cache hit calling any upload APIs before calling Execute would be net worse than calling GetActionResult or Execute and uploading only on miss.

So bazel could call one fewer distinct API if it wanted to (Execute*2 instead of Get...Execute), but still needs to call GetActionResult for cache-only flows. So not really better? On the service owner side, Ed's point holds: it's simpler if a service doesn't _have_ to check the action cache on Execute, and doesn't _have_ to be able to quickly reject actions where blobs are missing. 

On the other side, one feature I'd like to eventually add to the API is inline uploads in the Execute call itself. (Probably gated on something in the Capabilities API, so services with separation between CAS and Execute don't have to implement it). At that point build tools can make an optimization when executions are _probably_ incremental or are likely to be critical-path bound of using a single-RPC fast-path:
   Execute(inc. inline upload of changed files; merkle tree nodes; action)
You could still maybe argue that skip_cache_lookup is still unnecessary (just set it to true always), as in Ed's case though then the failure mode of bazel calling the "fast" path when there is actually a cache entry is a lot worse.

ulfjack: We already have code in Blaze to upload locally modified files eagerly (we detect locally modified files very early in the build), and we've found that to be a significant performance improvement for highly incremental builds at the time, so it seems likely that that's going to happen here as well. (Caveat: it's possible that Bazel's internals have changed sufficiently that eager uploading is no longer a win.)

EricBurnett: +1 for eagerly uploading files - I think it'd still be a boon - but note that merkle trees and action protos also need upload, and can only be constructed at around the time the action is ready for execution anyways.

illicitonion:
FWIW in Pants we don't use this mechanism at all because it's not flexible enough for us. We use a unique value, which we call a seed (that gets set as an environment variable, because there's nowhere else convenient to put it) - rather than using the caching flags; we change the seed if we want to bypass the cache.

The reason we do this is because we want to be able to do performance benchmarks, and we want to include caching effects there. One of our benchmark runs is:
Do a build
Change one file
Flush any in-memory caches (so we hit the remote cache)
Do a build again

We do this with a unique cache-seed for that benchmark run, because it means we know that the cache contains exactly what we expect. If we just used booleans, we would have no convenient way of isolating the cache per-benchmark-run; we only have the flexibility to either entirely disable the cache (where we don't get to test that cache fetches are as fast as we expect), or entirely enable it (where we need to trust that the change we made is globally unique).

In particular, at Twitter, we use this mechanism to do things like "reply the last n commits' CI runs" when validating new Pants releases to detect regressions, where we _know_ the cache has otherwise been populated, and which we wouldn't be able to do with just booleans.

build/bazel/remote/execution/v2/remote_execution.proto
EdSchouten: > Ed, just to make sure I understand: is the semantic clarification that in-flight actions may not be deduped problematic for you, or more generally that some actions are uncacheable at all (as defined within the Action)?

The latter. Right now Buildbarn's scheduler always deduplicates actions. It would be possible to disable this logic, but then it would be disabled globally; not on a per-action basis. Buildbarn's scheduler has no access to the CAS and thus not to the Action message.

Shall I send out a pull request to move this flag after your pull request lands?

EricBurnett: Yes please! Though I suggest syncing with @ola-rozenfeld on _how_ ... I think this will need to be a change destined for V3? It's a breaking change that I'm not sure offhand how to make in a forwards and backwards compatible way.

EricBurnett:
@illicitonion thanks for the comment, and the idea of using it to verify new tool builds! That does sound like a useful pattern.

We use "seeds" in some contexts as well (as an additional key+value in the Platform properties rather than an env var - might work for you too?), and I think you're right that it could entirely replace do_not_cache if so desired. (@edbaunton , more food for thought). I don't think it could replace skip_cache_lookup though, as that flag supports use-cases where the same action cache entry needs to be replaced for some reason, but caching is still otherwise desired for this and future build tools.

ulfjack:
We've internally been using a flag (outside of the action proto which is used for canonicalization) to indicate that we're not willing to accept a cached response, but are ok with the rerun being cached. We've used this to overwrite bad cache entries, as well as to make sure tests are rerun (e.g., to detect flakiness; note that --runs_per_test also adds a key to the action proto).

I don't think we've ever added a truly random number to an action request proto. That prevents any caching, but can also fill up the cache with non-reusable entries.

build/bazel/remote/execution/v2/remote_execution.proto
EdSchouten: I think it'll be manageable. We can add the field to `ExecuteRequest` as well, stating that:

- The one in `ExecuteRequest` is the canonical one going forward.
- For compatibility, clients should set both to true and false.
- For compatibility, servers should 'or' both of them together.

EricBurnett: With the idea being that Buildbarn will then be noncompliant with the spec (you'll only actually check the one in ExecuteRequest), but you think that'll be benign?

ola-rozenfeld: I like this plan, but let's do this change in a separate PR, and have the discussion there as well, for records.

EdSchouten: Exactly. Buildbarn can easily check both of them to determine whether something needs to be inserted into the cache (which it already does for the current flag), but will only respect the flag in the ExecuteRequest to prevent merging of in-flight requests.

EricBurnett: SGTM then! And glad to hear we don't have to consider V3 just yet :).

EricBurnett:
Since there are no objections that I can see (just discussions that don't seem blocking), anyone want to approve and merge this change? Or is there anything else we're waiting on before that's appropriate.

Thanks!

PR #59, opened 2019-02-22
Define OS and ISA as standard platform properties

juergbi:
As discussed at the build meetup in London, here is my proposal to define an initial set of standard platform properties.

I'm happy to trim down the initial set of standard values, if some values require further discussion.

I'm unsure about capitalization. I've used all lower case for ISA values to keep it simple but I'm open to change this, of course.

### DIFF
--- a/build/bazel/remote/execution/v2/platform.md
+++ b/build/bazel/remote/execution/v2/platform.md
@@ -1,0 +1,120 @@
+# Platform Lexicon
+
+This lexicon defines standard platform property names and values that servers
+MAY support in the `Platform` message to facilitate interoperability.
+
+The following standard property `name`s are defined:
+
+* `OSFamily`: This describes the operating system family. Multiple values
+  are not allowed and an exact match is required.
+
+  The following standard values are defined:
+
+    - `aix`
+    - `freebsd`
+    - `linux`
+    - `macos`
+    - `sunos`
+    - `windows`
+
+  Additional values may be defined by the server. For other POSIX systems
+  the recommendation is to use the output of `uname -s` in lower case.
+
+* `ISA`: This describes the instruction set architecture including
+  instruction set extensions and versions. Multiple values are allowed. If
+  multiple values are specified, they are AND-ed together: the worker is
+  required to support all of the listed values.
+
+  The following standard instruction set architecture values are defined:
+
+    - `arm-a32` (little endian)
+    - `arm-a32-be` (big endian)
+    - `arm-a64` (little endian)
+    - `arm-a64-be` (big endian)
+    - `arm-t32` (little endian)
+    - `arm-t32-be` (big endian)
+    - `power-isa-be` (big endian)
+    - `power-isa-le` (little endian)
+    - `sparc-v9` (big endian)
+    - `x86-32`
+    - `x86-64`
+
+  The following standard instruction set extension and version values are
+  defined:
+
+    - `arm-neon`
+    - `arm-sve`
+    - `arm-vfpv3`
+    - `arm-vfpv4`
+    - `armv6`
+    - `armv7`
+    - `armv8`
+    - `x86-avx`
+    - `x86-avx2`
+    - `x86-avx-512`
+    - `x86-sse4.1`
+    - `x86-sse4.2`
+
+  Additional values may be defined by the server. Vendor-neutral names are
+  recommended.
+
+  Clients SHOULD NOT request instruction set extensions or versions without
+  requesting an instruction set architecture.
+
+  Examples with multiple values:
+
+  ```json
+  // (Platform proto)
+  {
+    "properties": [
+      {
+        "name": "ISA",
+        "value": "x86-64"
+      },
+      {
+        "name": "ISA",
+        "value": "x86-avx2"
+      }
+    ]
+  }
+  ```
+
+  ```json
+  // (Platform proto)
+  {
+    "properties": [
+      {
+        "name": "ISA",
+        "value": "arm-a64"
+      },
+      {
+        "name": "ISA",
+        "value": "armv8"
+      },
+      {
+        "name": "ISA",
+        "value": "arm-sve"
+      }
+    ]
+  }
+  ```
+
+  ```json
+  // (Platform proto)
+  {
+    "properties": [
+      {
+        "name": "ISA",
+        "value": "arm-a32"
+      },
+      {
+        "name": "ISA",
+        "value": "armv7"
+      },
+      {
+        "name": "ISA",
+        "value": "arm-vfpv4"
+      }
+    ]
+  }
+  ```


juergbi:
I'd like to define the same standard properties in the RWAPI. However, as that's not yet in this repository, I'm focusing on the REAPI for now.

ola-rozenfeld:
+@bwkimmel

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -520,6 +520,49 @@ message Platform {
   // The server MAY use the `value` of one or more properties to determine how
   // it sets up the execution environment, such as by making specific system
   // files available to the worker.
+  //
+  // The following standard `name`s are defined:
+  //
+  // * `OS`: This describes the operating system. Only a single value is allowed
```
EricBurnett: I'd vote to lowercase everything - keys ("OS" -> "os"), values, etc.

EricBurnett: Is your intention that all servers *must* support OS? If not, how about adding a caveat:
  "Only a single value..." -> "If supported, only a single value..." ?

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -520,6 +520,49 @@ message Platform {
...
+  //   and an exact match is required. The following standard values are
+  //   defined:
```
EricBurnett: Do you want to proscribe what should happen with non-standard values? One option is that clients and servers are free to make up their own; another is that anyone wanting to support a value not in the canonical list should namespace it in some way (e.g. '<service>_<os>').

Depends if we think that adding new standard values will be benign or problematic if some implementers are already using their own custom ones. (I'm leaning towards thinking it'll be benign, and so allowing free use of non-standard values, but interested in your take).

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -520,6 +520,49 @@ message Platform {
...
+  //     - AIX
+  //     - FreeBSD
+  //     - Linux
+  //     - MacOS
+  //     - SunOS
+  //     - Windows
+  //
+  // * `ISA`: This describes the instruction set architecture including
+  //   instruction set extensions. Multiple values are allowed. The worker is
+  //   required to support all of the listed values. The following standard
```
EricBurnett: > "The worker is required to support all of the listed values." 

Maybe phrase as "If multiple values are specified, they are AND-ed together: the worker is required to support all of the listed values".

Otherwise, it sounds like the comment is saying "if a value is specified, the server is required to ensure the worker supports it", rather than specifying the logic for combining multiple, which contradicts the higher level comments that it's up to the server how to handle keys they don't understand (one allowed option being to ignore them entirely).

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -520,6 +520,49 @@ message Platform {
...
+  //   values are defined:
```
EricBurnett: Do you know if anyone has an existing list of ISA that we could point to? Maintaining our own list of values seems like it may be a maintenance burden in this case, if there's an easier way we can have an up-to-date list.

build/bazel/remote/execution/v2/remote_execution.proto
edbaunton: Is it worth saying that this is, when possible, the result of running `uname -s` or is that not a correct approximation? For example my Mac reports Darwin which is not on the list but Linux, AIX and SunOS report as listed.

build/bazel/remote/execution/v2/remote_execution.proto
edbaunton: Is it simply analogous to `uname -i` + extensions?

build/bazel/remote/execution/v2/remote_execution.proto
juergbi: That's fine with me. Will change it unless there are objections.

juergbi: I wouldn't mandate it, however, I think it would make sense for most servers to support it. The only exception being servers that don't accept any OS-specific input.

From a client request perspective, "If supported, only a single value..." seems a bit odd, in my opinion. Is this just me? How about "Multiple values are not allowed." as alternative?

EricBurnett: "Multiple values are not allowed" sounds better, yeah, thanks!

And yeah, it being _recommended_ especially for multi-OS services sounds appropriate to me, but this way there's a nice carveout for services backing homogeneous fleets and other such use-cases.

agoulti: "it would make sense for most servers to support it" is great in general. Unfortunately, I'm not sure if we'll get there just by adding a comment to the API.

Actually, do we know the state of the world wrt this?

I understand BuildBarn has no immediate plans to support routing based on Platform and it would be a lot of work to bring it to conformance; RBE has no near-term plans for this type of matching.
Do we know the state of support/near-term plans for the other servers?

EricBurnett: @bwkimmel to comment on RBE's interest and timeline in supporting OS specifically.

build/bazel/remote/execution/v2/remote_execution.proto
juergbi: I'm also thinking that namespacing non-standard values is not required. It may be different for other platform properties but for `os` and also `isa` I consider the risk of conflicting value definitions relatively low. There is a chance that implementations use different non-standard values for the same OS or ISA, however, namespacing non-standard values wouldn't help in that case.

> Is it worth saying that this is, when possible, the result of running `uname -s` or is that not a correct approximation? For example my Mac reports Darwin which is not on the list but Linux, AIX and SunOS report as listed.

I think it makes sense to match `uname -s` in most cases, but there are some exceptions. Besides Windows as special case without `uname`, I intentionally used 'MacOS' instead of 'Darwin' as I think the kernel alone is insufficient in that case. To my knowledge, Darwin doesn't have a stable syscall ABI and the common case would be for input binaries to depend on macOS userspace parts as well. Linux is different in this regard.

build/bazel/remote/execution/v2/remote_execution.proto
juergbi: Sounds good, will include this.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -520,6 +520,49 @@ message Platform {
   // The server MAY use the `value` of one or more properties to determine how
   // it sets up the execution environment, such as by making specific system
   // files available to the worker.
+  //
+  // The following standard `name`s are defined:
```
bergsieker: I'm wondering if this really belongs in the core API file, or in a separate lexicon file. That would allow the API to remain generic and flexible, but clients and servers could choose to specify that they are using a particular platform lexicon (which could then have well-defined semantics for how particular fields are matched).

build/bazel/remote/execution/v2/remote_execution.proto
juergbi: > Do you know if anyone has an existing list of ISA that we could point to?

Unfortunately, I'm not aware of an existing list that would be suitable. The list in my patch aims to match official documentation where reasonable, but in a vendor-neutral and consistent form.

> Is it simply analogous to `uname -i` + extensions?

Do you mean `uname -m`? `uname -i` just prints 'unknown' here on Linux.

Unfortunately, `uname -m` is OS-specific. E.g., various BSDs use 'amd64' while many other operating systems use 'x86_64'. And Linux doesn't use anything generic for 32-bit x86, it reports 'i386', 'i486', 'i586' or 'i686', while SunOS/Solaris reports 'i86pc'. There are also inconsistencies for ARM.

build/bazel/remote/execution/v2/remote_execution.proto
bergsieker: My preference is lower-dash-separated (I don't know if there's a formal name for that style). Of course, for "os" and "isa" there are no dashes. :)

juergbi: > I understand BuildBarn has no immediate plans to support routing based on Platform and it would be a lot of work to bring it to conformance

A possible option for minimal support would be to at least reject incompatible requests, even if only a single worker platform is supported per instance.

bwkimmel: For worker attributes specified via the REAPI platform variables, RBE has been Pascal casing for "standard" attributes and camel casing for non-standard attributes, with initialisms maintaining fixed casing throughout (i.e., all lower-case if they appear at the beginning of a non-standard attribute name, all upper-case otherwise).

For example, we have been using `OSFamily` (note that "OS" is all-caps) and `Pool` as "standard" attributes and `gceMachineType` (note that "gce" is all lowercase) as a non-standard attribute. My preference would be to maintain this convention.

bwkimmel: Perhaps rename to `OSFamily`? `OS` seems a bit ambiguous regarding how specific this ought to be. I.e., should it only be the high-level `OS` "family" as you've specified here? Or should it represent the full specification (`uname -a`), or something in between.

Internally, RBE has been using `OSFamily` to represent this value (i.e., just `Linux`, `Windows`, etc.) -- though as Eric indicated, this is not yet supported externally. We could later standardize other variables to indicate more specific values, e.g., `OSDistribution` might take on values like `Ubuntu`, `Debian`, etc., `OSDistributionVersion` for the version number of that distribution (e.g., `18.04`).

juergbi: I'd be fine either way for the keys, all lower case or Pascal case (the abbreviations `OS` and `ISA` would remain all-caps, I presume). However, for the values Pascal case would not be ideal, in my opinion, especially for ISA values I'd prefer all lower case to keep it clear and simple. Does this work for you?

juergbi: `OSFamily` sounds fine to me.

build/bazel/remote/execution/v2/remote_execution.proto
juergbi: Would that separate lexicon be in a separate repository or remain here and thus, still be considered (an optional) part of the official API specification?

I don't have any objections moving this to a separate file. However, I think we should only have one upstream platform lexicon. It may make sense for special purpose or internal deployments of remote API infrastructure to have a different platform lexicon. However, I don't see a point in defining multiple incompatible lexicons for general purpose build infrastructure.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -520,6 +520,49 @@ message Platform {
...
+  //
+  //     - AIX
+  //     - FreeBSD
+  //     - Linux
+  //     - MacOS
+  //     - SunOS
+  //     - Windows
+  //
+  // * `ISA`: This describes the instruction set architecture including
+  //   instruction set extensions. Multiple values are allowed. The worker is
```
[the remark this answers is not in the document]
juergbi: I don't see a good option how we could fully address this with the current string-based key-value structure. I'd consider this an edge case we could skip for now.

Either the FreeBSD worker could claim to be a Linux-only system or if running a mix of FreeBSD and Linux binaries is desired, the Linux compatibility layer support could be specified in a separate platform property (just like other kernel/OS config options).

Or do you have another idea how to cover all of this nicely?

build/bazel/remote/execution/v2/remote_execution.proto
juergbi: I've kept the keys Pascal case in line with RBE but I also wouldn't have any objections to change this to lower-dash-separated.

The `ISA` values are lower-dash-separated and I think that's sensible, especially as there isn't a clear Pascal case equivalent for all values.

I've changed the `OSFamily` values to lower-dash-separated to be in line with the `ISA` values. However, Pascal case would work ok for `OSFamily` values as well.

Is this acceptable to everyone?

juergbi: I've changed this to `OSFamily`. Does this look ok now?

build/bazel/remote/execution/v2/remote_execution.proto
juergbi: I've added a brief comment about non-standard values. Does this suffice or any further suggestions on this?

build/bazel/remote/execution/v2/remote_execution.proto
juergbi: I've split the list into instruction set architectures and instruction set extensions/versions. Is the suggested list acceptable as a starting point or does anyone have an alternative suggestion?

build/bazel/remote/execution/v2/remote_execution.proto
juergbi: I've moved this to a separate file called `platform.md`. I'm open for better suggestions on name or format.

build/bazel/remote/execution/v2/remote_execution.proto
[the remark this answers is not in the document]
juergbi: Are you happy not covering such edge cases with the standard platform lexicon for the time being?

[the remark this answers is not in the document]
EdSchouten: Sure, but I think this could be solved relatively easy, though. What if the OS+ISA were announced through a single label (key or value)? "I want the build to run on a system that supports FreeBSD/amd64." This is, in fact, also closer towards the way OS kernels model their ISA support. For example, on FreeBSD, every OS+ISA pair has its own kernelspace `struct sysentvec` that points to callbacks on how to deal with syscalls, faults, signals, etc.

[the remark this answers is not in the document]
juergbi: This wouldn't allow adding more detailed properties such as `OSKernelVersion` or `OSDistribution` in the future, and already ISA extension handling might require extra effort. I.e., we'd essentially have to ignore the key/value structure for anything OS/ISA-related and define a new text-based format that covers these parts. That's definitely possible but seems like a workaround, not how the spec was designed.

Clients issuing actions that require execution on multiple operating system ABIs seems pretty niche to me. Workers that support multiple operating system ABIs might be interesting in more cases but maybe the better approach for that would be to allow workers to advertise multiple `Platform` instances, each with `OSFamily` and `ISA`.

ola-rozenfeld:
Looks like all the questions have been resolved -- can any stake holders please approve, if so? @EdSchouten, @EricBurnett , @bergsieker ? I will merge after two approvals. Thank you!

build/bazel/remote/execution/v2/platform.md
```diff
@@ -0,0 +1,120 @@
...
+    - `x86-avx`
+    - `x86-avx2`
+    - `x86-avx-512`
+    - `x86-sse4.1`
+    - `x86-sse4.2`
+
+  Additional values may be defined by the server. Vendor-neutral names are
+  recommended.
+
+  Clients SHOULD not request instruction set extensions or versions without
```
EricBurnett: Nit: "SHOULD not" -> "SHOULD NOT"

juergbi: Thanks, fixed.

PR #64, opened 2019-03-13
Adding C++ gRPC libraries. Also fixing build for latest Bazel.

ola-rozenfeld:
I don't know why it was necessary to move the BUILD.googleapis file to external, but it wouldn't compile otherwise no matter what I tried.


ola-rozenfeld:
Dang it, it took my other email somehow. Don't know how this could have happened. Fixing...

PR #69, opened 2019-04-11
support inlining of output files in the action result

buchgr:
Allow the client to request the inlining of action outputs,
stdout and stderr in the action result message. We only add
this capability to the GetActionResult call for now as its
typically on the latency sensitive fast path, while Execute
is not.

Also make the stdout and stderr inlining decisions client
controlled and recommend for the server to keep the serialized
size of the ActionResult message below 1 MiB [1].

See https://groups.google.com/d/msg/remote-execution-apis/bmUuI2wUcwA/p0yrYq7VBgAJ

[1] https://developers.google.com/protocol-buffers/docs/techniques#large-data

### DIFF
--- a/build/bazel/remote/execution/v2/remote_execution.proto
+++ b/build/bazel/remote/execution/v2/remote_execution.proto
@@ -738,8 +738,8 @@ message ActionResult {
   // The output files of the action. For each output file requested in the
   // `output_files` field of the Action, if the corresponding file existed after
   // the action completed, a single entry will be present either in this field,
-  // or the `inlined_output_files` field or the `output_file_symlinks` field if the
-  // file was a symbolic link to another file.
+  // or the `output_file_symlinks` field if the file was a symbolic link to
+  // another file.
   //
   // If the action does not produce the requested output, or produces a
   // directory where a regular file is expected or vice versa, then that output
@@ -845,7 +845,8 @@ message ActionResult {
   // The standard output buffer of the action. The server SHOULD NOT inline
   // stdout unless requested by the client in the
   // [GetActionResultRequest][build.bazel.remote.execution.v2.GetActionResultRequest]
-  // message.
+  // message. The server MAY omit inlining, even if requested, and MUST do so if inlining
+  // would cause the response to exceed message size limits.
   bytes stdout_raw = 5;
 
   // The digest for a blob containing the standard output of the action, which
@@ -856,7 +857,8 @@ message ActionResult {
   // The standard error buffer of the action. The server SHOULD NOT inline
   // stderr unless requested by the client in the
   // [GetActionResultRequest][build.bazel.remote.execution.v2.GetActionResultRequest]
-  // message.
+  // message. The server MAY omit inlining, even if requested, and MUST do so if inlining
+  // would cause the response to exceed message size limits.
   bytes stderr_raw = 7;
 
   // The digest for a blob containing the standard error of the action, which
@@ -889,6 +891,8 @@ message OutputFile {
   // The contents of the file if inlining was requested. The server SHOULD NOT inline
   // file contents unless requested by the client in the
   // [GetActionResultRequest][build.bazel.remote.execution.v2.GetActionResultRequest]
+  // message. The server MAY omit inlining, even if requested, and MUST do so if inlining
+  // would cause the response to exceed message size limits.
   bytes contents = 5;
 }
 
--- a/build/bazel/remote/execution/v2/remote_execution.proto
+++ b/build/bazel/remote/execution/v2/remote_execution.proto
@@ -845,8 +845,7 @@ message ActionResult {
   // The standard output buffer of the action. The server SHOULD NOT inline
   // stdout unless requested by the client in the
   // [GetActionResultRequest][build.bazel.remote.execution.v2.GetActionResultRequest]
-  // message and instead return a digest in `stdout_digest` that points to
-  // the buffer.
+  // message.
   bytes stdout_raw = 5;
 
   // The digest for a blob containing the standard output of the action, which
@@ -857,8 +856,7 @@ message ActionResult {
   // The standard error buffer of the action. The server SHOULD NOT inline
   // stderr unless requested by the client in the
   // [GetActionResultRequest][build.bazel.remote.execution.v2.GetActionResultRequest]
-  // message and instead return a digest in `stderr_digest` that points to
-  // the buffer.
+  // message.
   bytes stderr_raw = 7;
 
   // The digest for a blob containing the standard error of the action, which
--- a/build/bazel/remote/execution/v2/remote_execution.proto
+++ b/build/bazel/remote/execution/v2/remote_execution.proto
@@ -732,10 +732,6 @@ message ExecutedActionMetadata {
 
 // An ActionResult represents the result of an
 // [Action][build.bazel.remote.execution.v2.Action] being run.
-//
-// The server MUST NOT make any inlining decisions on his own but should
-// do so when requested by the client. The total serialized message size
-// SHOULD NOT exceed 1 MiB.
 message ActionResult {
   reserved 1; // Reserved for use as the resource name.
 
@@ -750,12 +746,6 @@ message ActionResult {
   // will be omitted from the list. The server is free to arrange the output
   // list as desired; clients MUST NOT assume that the output list is sorted.
   repeated OutputFile output_files = 2;
-
-  // The output files of the action whose contents are inlined into the request.
-  // A server MUST NOT inline outputs unless requested by the client in the
-  // [GetActionResultRequest][build.bazel.remote.execution.v2.GetActionResultRequest]
-  // message.
-  repeated InlinedOutputFile inlined_output_files = 12;
 
   // The output files of the action that are symbolic links to other files. Those
   // may be links to other output files, or input files, or even absolute paths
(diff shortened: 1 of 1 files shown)


buchgr:
@werkt @ishikhman @EricBurnett @juergbi @edbaunton

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -731,21 +731,31 @@ message ExecutedActionMetadata {
 
 // An ActionResult represents the result of an
 // [Action][build.bazel.remote.execution.v2.Action] being run.
+//
+// The server MUST NOT make any inlining decisions on his own but should
+// do so when requested by the client. The total serialized message size
+// SHOULD NOT exceed 1 MiB.
```
edbaunton: Where does this 1MiB limit come from? A gRPC message size limit?

buchgr: See commit message :-). It's not a limit but advice from protobuf https://developers.google.com/protocol-buffers/docs/techniques#large-data

EricBurnett: I'd prefer to drop this recommendation, personally - we use >1MB messages all over the place today, and have seen no issues with doing so. My read is also that it wasn't intended to be a hard limit - the linked doc reads more like guidance not to use 100MB messages than it does a concrete recommendation that e.g. 4MB would be too much.

gRPC itself provides an upper bound on the size of messages - 4MB by default, if I remember correctly. I don't think we need a documented recommendation beyond that.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -731,21 +731,31 @@ message ExecutedActionMetadata {
 
 // An ActionResult represents the result of an
 // [Action][build.bazel.remote.execution.v2.Action] being run.
+//
+// The server MUST NOT make any inlining decisions on his own but should
```
EricBurnett: Can we remove this comment? It's documented on the fields below, which seems like the appropriate place for it; I'm not sure that message-level comment adds anything.

Also, the server is free to *not* inline, so this one as written seems overly strong.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -731,21 +731,31 @@ message ExecutedActionMetadata {
...
   //
   // If the action does not produce the requested output, or produces a
   // directory where a regular file is expected or vice versa, then that output
   // will be omitted from the list. The server is free to arrange the output
   // list as desired; clients MUST NOT assume that the output list is sorted.
   repeated OutputFile output_files = 2;
 
+  // The output files of the action whose contents are inlined into the request.
+  // A server MUST NOT inline outputs unless requested by the client in the
+  // [GetActionResultRequest][build.bazel.remote.execution.v2.GetActionResultRequest]
```
EricBurnett: MUST NOT->SHOULD NOT?

I'm hesitant to make it a MUST, in that it'll be mostly benign if it happens - no correctness guarantees will be violated due to over-inlining, just some amount of bandwidth wastage. As such, SHOULD NOT seems more correct (from the descriptions in https://www.ietf.org/rfc/rfc2119.txt).

Same below.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -841,12 +851,13 @@ message ActionResult {
...
-  // on the size of the buffer, whether to return it in raw form or to return
-  // a digest in `stdout_digest` that points to the buffer. If neither is set,
-  // then the buffer is empty. The client SHOULD NOT assume it will get one of
-  // the raw buffer or a digest on any given request and should be prepared to
-  // handle either.
+  // The standard output buffer of the action. The server MUST NOT inline
+  // stdout unless requested by the client in the
+  // [GetActionResultRequest][build.bazel.remote.execution.v2.GetActionResultRequest]
+  // message and instead return a digest in `stdout_digest` that points to
+  // the buffer. The client SHOULD NOT assume it will get one of the raw
```
EricBurnett: Somewhat orthogonal to your proposal, but I think we err'd in saying that stdout_raw may be set and stdout_digest then left empty. I'd prefer the semantics be that stdout_digest is always set, and stdout_raw may *also* be set if inlining is requested. Otherwise if a client both wants to display the stdout and save a reference to it for later, the client needs to *upload* it to the CAS for persistence themselves, which seems backwards.

If we're changing the semantics of inlining anyways, and no implementation we know of relies on the semantics of today, could we change this?

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -894,6 +906,24 @@ message OutputFile {
   bool is_executable = 4;
 }
 
+// `InlinedOutputFile` is similar to [OutputFile][build.bazel.remote.execution.v2.OutputFile]
+// but has its contents inlined in the message.
+message InlinedOutputFile {
```
EricBurnett: Why not just add an optional 'contents' field to OutputFile? I'm not a huge fan of the duplication this currently requires, and don't really see why it would be necessary to segregate.

Tangential question: Anyone know why "`OutputFile` is binary-compatible with `FileNode`." ? I assume that was motivated by something in the past, but I can't figure out what. It's probably the wrong time to change that (should be a V3 feature request), but can still be satisfied by reserving the field number in FileNode, as is done with 3 now.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1126,6 +1156,19 @@ message GetActionResultRequest {
   // The digest of the [Action][build.bazel.remote.execution.v2.Action]
   // whose result is requested.
   Digest action_digest = 2;
+
+  // A hint to the server to inline stdout in the
+  // [ActionResult][build.bazel.remote.execution.v2.ActionResult] message.
```
EricBurnett: inline -> request inlining ? I think that helps reinforce the message that the server may not comply.

build/bazel/remote/execution/v2/remote_execution.proto
ulfjack: I agree that it would be better to report the bytes and the digest.

build/bazel/remote/execution/v2/remote_execution.proto
ola-rozenfeld: +1 to that. I think we should just revert https://github.com/bazelbuild/remote-apis/pull/5 to start with.

I also can't remember for the life of me why OutputFile should be binary compatible with FileNode either. I don't think it should.

build/bazel/remote/execution/v2/remote_execution.proto
buchgr: done

build/bazel/remote/execution/v2/remote_execution.proto
buchgr: I created the extra message because of binary compatibility requirement with FileNode. I agree to adding the field to `OutputFile` instead.

build/bazel/remote/execution/v2/remote_execution.proto
buchgr: done

build/bazel/remote/execution/v2/remote_execution.proto
buchgr: Great idea. Done.

build/bazel/remote/execution/v2/remote_execution.proto
buchgr: Done, also below.

buchgr:
Apologies for the delay in responding to comments. I have been quite busy with other things and all of a sudden a month had passed :-). I believe I have addressed all comments. I ll squash and update the commit message after getting an LGTM.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -841,32 +841,28 @@ message ActionResult {
...
-  // on the size of the buffer, whether to return it in raw form or to return
-  // a digest in `stderr_digest` that points to the buffer. If neither is set,
-  // then the buffer is empty. The client SHOULD NOT assume it will get one of
-  // the raw buffer or a digest on any given request and should be prepared to
-  // handle either.
+  // The standard error buffer of the action. The server SHOULD NOT inline
+  // stderr unless requested by the client in the
+  // [GetActionResultRequest][build.bazel.remote.execution.v2.GetActionResultRequest]
+  // message and instead return a digest in `stderr_digest` that points to
+  // the buffer.
```
ola-rozenfeld: The comments still make it look like the digest is optional.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -892,6 +886,11 @@ message OutputFile {
 
   // True if file is executable, false otherwise.
   bool is_executable = 4;
+
+  // The contents of the file if inlining was requested. The server SHOULD NOT inline
+  // file contents unless requested by the client in the
+  // [GetActionResultRequest][build.bazel.remote.execution.v2.GetActionResultRequest]
+  bytes contents = 5;
```
ola-rozenfeld: I know that usually we don't reuse removed tags, but this is a rare case of the exact same field being removed and then ressurrected. I think this justifies using bytes contents = 3, removing the "reserved".

build/bazel/remote/execution/v2/remote_execution.proto
buchgr: done.

build/bazel/remote/execution/v2/remote_execution.proto
buchgr: Meh. I'd prefer to stick to the proto guidelines if you don't mind.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -892,6 +884,11 @@ message OutputFile {
 
   // True if file is executable, false otherwise.
   bool is_executable = 4;
+
+  // The contents of the file if inlining was requested. The server SHOULD NOT inline
```
EricBurnett: Should we be explicit somewhere that the server MAY omit inlining, even if requested, and MUST do so if inlining would cause the response to exceed message size limits? Possibly here (and on other output fields), possibly on the relevant fields in GetActionResultRequest instead.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -737,8 +737,8 @@ message ActionResult {
   // The output files of the action. For each output file requested in the
   // `output_files` field of the Action, if the corresponding file existed after
   // the action completed, a single entry will be present either in this field,
-  // or in the output_file_symlinks field, if the file was a symbolic link to
-  // another file.
+  // or the `inlined_output_files` field or the `output_file_symlinks` field if the
```
EricBurnett: Out of date? inlined_output_files no longer exists, I don't think.

build/bazel/remote/execution/v2/remote_execution.proto
buchgr: done.

build/bazel/remote/execution/v2/remote_execution.proto
buchgr: done.

buchgr:
Good to go?

ishikhman:
LGTM ;)

EricBurnett:
LGTM as well, thanks Jakob.

Issue #65, opened 2019-03-15
Add generated *.pb.go


ukai:
google/go-genproto#138 and googleapis/google-cloud-go#1322

could you add generated *.pb.go to make it easier for Go developers to use that API?

ola-rozenfeld:
Closed via 5556e9c6153f41bf8f2ebcf6272cfc1ec201b279

PR #72, opened 2019-05-03
Add go_proto_library plus generated files for "go build".

jasharpe:
Generating Go code like this matches the pattern precedented by https://github.com/google/go-genproto for Google APIs.

Generated files are placed in:
- v2/ for remote execution, matching the style used by go-genproto for versioned APIs, e.g. https://github.com/google/go-genproto/blob/master/googleapis/appengine/v1/appengine.pb.go#L4
- semver/ for semver library, matching the style used by go-genproto for unversioned APIs, e.g. https://github.com/google/go-genproto/blob/master/googleapis/api/distribution/distribution.pb.go#L4

Unlike go-genproto, we use Bazel to generate the Go code, which is much simpler, since it handles calling protoc for us.

go.mod and go.sum generated by:

```
go mod init github.com/bazelbuild/remote-apis
go build ./...
```

I tested this by creating a Go project that imports this proto and confirmed that I can build it with Bazel and go build.

This addresses issue #65.


buchgr:
I understand that this generator has to be rerun with every API change?

jasharpe:
Yes. I'm not sure a good way to automate this. Maybe there is some git hook that could require it?

buchgr:
A git hook sounds like a fine choice to me i.e. a pre-commit hook that runs the code generation steps. Would you mind adding this? Should be quite straightforward?

ola-rozenfeld:
That's a great idea, thank you!

jasharpe:
I agree technically it should be simple (I'll try it out), though I have some concerns:
1. It would force a "bazel build" on every commit which might be slow/annoying. Although it may also be useful for checking correctness.
2. The instructions only work on linux_amd64 machines currently (I don't know a good way to generate this part of the path, though I'll try using a `find` to avoid needing to hardcode a path).

jasharpe:
I updated to include a pre-commit hook in the `hooks` folder. Unfortunately there's no way to enable this by default, but I included instructions for setting up the hooks for developers.

There could be some quibbling about whether to "git add" these files by default or not. It is most painless to just "git add" them regardless since it doesn't require multiple commits (or amending) to get the generated files into the same commit as the changed proto files.

Issue #61, opened 2019-02-27
Capabilities in multi-endpoints configuration


t-chaik:
The REAPI describes four services: `Execution` (EXEC), `ActionCache` (AC), `ContentAddressableStorage` (CAS) and `Capabilities`. The services do have inter-dependencies but the specification doesn't restrict nor advocate on how separated a server implementation can or should expose them. Clients tend to support only some possible configurations:
* Bazel has support for two (as far as I know):
    - EXEC + CAS + AC available at one end-point.
    - EXEC separated from CAS + AC.
* BuildStream has support for any: three different endpoints can be specified for the three main services.
* RECC has support for two (goal is to support any):
    - EXEC + CAS + AC available at one end-point.
    - EXEC + AC separated from CAS.

The `Capabilities` allow requesting server capabilities at a given endpoint. The `ServerCapabilities` messages contains two sets of capabilties:
* `CacheCapabilities` relevant for CAS and AC.
* `ExecutionCapabilities` relevant for EXEC.

The specification currently doesn't mention anything about how the server should advertise capabilities. Disparities on client expectations already exists. For example Bazel expects all capabilities to be served at the EXEC end-point (event if CAS + AC are separated) while BuildStream expects a `Capabilities` service to be available at every endpoint but only advertising relevant capabilities for the services hosted at that endpoint .

I think we should discuss and agree on how capabilities should be advertised by server implementations and how client implementations should query them depending on the endpoint configuration. Sensible conclusions should probably be part of the specification.

ola-rozenfeld:
Yes, agreed, we should be more explicit in the spec! One thing I wanted to avoid is to spec out the full system configuration in the ServerCapabilities -- which endpoints there are altogether, what services do they support, and with what properties. I think it's useful to distinguish between *properties of a particular endpoint* vs *properties of the service as a whole*, and I chose to have ServerCapabilities represent the service as a whole, because that is a lot easier to describe (common elements, inter-dependencies), and also more useful to clients. But I see how that can be misleading -- should all endpoints in the system return all the capabilities, then? My idea was this:

If an execution endpoint is present in the system, it should return the ServerCapabilities of the whole system. All other endpoints should return the same response, except for ExecutionCapabilities.

It is a bit ugly idea, but it is simple.
One alternative I see would be to formalize the `EndpointServices` idea and add something like:

```
message EndpointServices {
  boolean action_cache = 1;
  boolean cas = 2;
  boolean execution = 3;
[...]

PR #76, opened 2019-05-09
Split the Capabilities by endpoints

ola-rozenfeld:
Fixes https://github.com/bazelbuild/remote-apis/issues/61.
Will need to fix Bazel to match!

### DIFF
--- a/build/bazel/remote/execution/v2/remote_execution.proto
+++ b/build/bazel/remote/execution/v2/remote_execution.proto
@@ -332,7 +332,14 @@ service ContentAddressableStorage {
 // The query may include a particular `instance_name`, in which case the values
 // returned will pertain to that instance.
 service Capabilities {
-  // GetCapabilities returns the server capabilities configuration.
+  // GetCapabilities returns the server capabilities configuration of the
+  // remote endpoint.
+  // Only the capabilities of the services supported by the endpoint will
+  // be returned:
+  // * Execution + CAS + Action Cache endpoints should return both
+  //   CacheCapabilities and ExecutionCapabilities.
+  // * Execution only endpoints should return ExecutionCapabilities.
+  // * CAS + Action Cache only endpoints should return CacheCapabilities.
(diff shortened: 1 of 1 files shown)


build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -332,7 +332,14 @@ service ContentAddressableStorage {
 // The query may include a particular `instance_name`, in which case the values
 // returned will pertain to that instance.
 service Capabilities {
-  // GetCapabilities returns the server capabilities configuration.
+  // GetCapabilities returns the server capabilities configuration of the
+  // remote endpoint.
+  // Only the capabilities of the services supported by the endpoint will
+  // be returned. For example:
```
buchgr: Is 'For example' necessary here? Couldn't we simply specify an exhaustive list?

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1301,9 +1308,13 @@ message GetCapabilitiesRequest {
 // [Capabilities.GetCapabilities][build.bazel.remote.execution.v2.Capabilities.GetCapabilities].
 message ServerCapabilities {
   // Capabilities of the remote cache system.
+  // This field will be returned by endpoints that support CAS and Action
+  // Cache services.
```
buchgr: Does this comment add additional value? If not, could we remove it? Same below.

build/bazel/remote/execution/v2/remote_execution.proto
ola-rozenfeld: Hmm, I guess this is already an exhaustive list :-)

build/bazel/remote/execution/v2/remote_execution.proto
ola-rozenfeld: Sure.

edbaunton:
Maybe I missed this in a separate discussion but is it the case that we cannot host the ActionCache separately from the CAS?

ola-rozenfeld:
Correct, these two are linked.

PR #77, opened 2019-05-10
Add a reference to the new platform.md from proto

edbaunton:
So that readers are aware of the relevance of platform.md


buchgr:
LGTM

PR #79, opened 2019-05-22
clarify consistency requirements of GetActionResult

buchgr:
- move the statement from the service documentation to the call.
- add a section about extending TTLs

related discussions:
- https://groups.google.com/d/msg/remote-execution-apis/zmcEV8kubRc/5PPcHOVZAgAJ
- bazelbuild/bazel#6862

@EricBurnett @werkt

### DIFF
--- a/build/bazel/remote/execution/v2/remote_execution.proto
+++ b/build/bazel/remote/execution/v2/remote_execution.proto
@@ -128,10 +128,7 @@ service Execution {
 //
 // The lifetime of entries in the action cache is implementation-specific, but
 // the server SHOULD assume that more recently used entries are more likely to
-// be used again. Additionally, action cache implementations SHOULD ensure that
-// any blobs referenced in the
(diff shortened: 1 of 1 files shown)


ulfjack:
I think that looks fine. I seem to remember a discussion of this on the Bazel issue tracker - can we link that from the change description or at least from this thread for future reference?

buchgr:
@ulfjack done

ulfjack:
Let me add that my approving this is not a statement on what the right long-term strategy here is. There has been work on supporting missing remote artifacts in Bazel and 'rewinding' execution to regenerate them. However, it does add a lot of complexity, and it would probably be simpler for us if we could rely on things not silently going blubb.

buchgr:
@ulfjack Right. I'd think this is orthogonal to supporting 'rewind' action support. We can still decide to do it in Bazel if we find it necessary. Either way, this change seems like an improvement to me.

EricBurnett:
LGTM

PR #81, opened 2019-05-23
Add a way for users of remote-apis to only require dependencies for the rules of languages they actually use

### LINKED ISSUE #78: loads in //build/bazel/remote/execution/v2 induce dependencies
The loads of grpc and rules_go starlark methods mandate that any workspace that references the //build/bazel/remote/execution/v2 package - even to non-language targets - provide both the rules_go and grpc repositories. This will break implementors that do not have these dependencies already, and require that wholly unrelated implementations (i.e. buildfarm in Java) include declarations for these go and grpc repos, that must match by name.

If remote-apis is going to continue down this path of providing language implementations (C++ and go so far), it should provide an extension library for initializing the repo with these dependencies activated, and personally I would like to have these available in separate packages by language, if only to make it possible not to import the growing list of supported language rules toolchains.

jasharpe:
Add a way for users of remote-apis to only require dependencies for the rules of languages they actually use.

We reuse the solution developed for googleapis, see "switched_rules_by_language": https://github.com/googleapis/googleapis/blob/master/repository_rules.bzl

Basically the way this works is that we generate a Bazel repository that wraps the enabled language rules, and implements them as no-ops if they are not specifically enabled.

I've tested with with tools_remote, and builds there now work when this snippet is added to the workspace, even without adding go workspace dependencies:

```
load("@remoteapis//:repository_rules.bzl", "switched_rules_by_language")
switched_rules_by_language(
    name = "bazel_remote_apis_imports",
    java = True,
)
```

Fixes #78.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,26 @@ packages from that repository in order to generate code using this API. If you
 build the repository using the included `BUILD` files, Bazel will fetch the
 protobuf compiler and googleapis automatically.
 
+## Selecting language support
+
+The repository provides a rule that you can use in your `WORKSPACE` file to
+select which languages you are using. This allows you to depend on this
+repository without pulling in Go rules, for instance, if you aren't using Go
+rules.
+
+For instance, if you only use Java in your project, add the following to your
+`WORKSPACE` file:
+
+```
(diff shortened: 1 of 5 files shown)


buchgr:
Sweet ... why did we not think of this ourselves :)

PR #85, opened 2019-06-03
Wrapping all enums in messages for clarity in C++/Py.

### LINKED ISSUE #56: SymlinkAbsolutePathStrategy enum should be wrapped in a message
Protobuf's enum scoping follows traditional C/C++ rules, where the enum name is not a new level of namespacing. This means the current definition of `SymlinkAbsolutePathStrategy`:

```proto
message CacheCapabilities {
  enum SymlinkAbsolutePathStrategy {
    UNKNOWN = 0;
    DISALLOWED = 1;
    ALLOWED = 2;
  }
  SymlinkAbsolutePathStrategy symlink_absolute_path_strategy = 5;
```

will define `CacheCapabilities::ALLOWED`, `CacheCapabilities::DISALLOWED`, etc. This is very confusing because the symbols indicate that caching itself is allowed/disallowed, instead of just this one particular feature.

I recommend the following structure instead:

```proto
message SymlinkAbsolutePathStrategy {
  enum Enum {
    UNKNOWN = 0;
    DISALLOWED = 1;
    ALLOWED = 2;
  }
}

message CacheCapabilities {
  SymlinkAbsolutePathStrategy.Enum symlink_absolute_path_strategy = 5;
```

ola-rozenfeld:
Fixes #56.


buchgr:
Is this change backwards compatible?

ola-rozenfeld:
Yes, none of the enum values changed, so it should be. The rollout plan is:
1. sync Bazel to latest remote-apis NOT including this commit (for future rollback-ability)
2. same internally, for RBE
3. prepare & submit an external Bazel patch syncing remote-apis to this commit, and fixing compile
4. internally, submit a change that includes three things:
  - syncing remote-apis to this commit
  - fixing Bazel compile
  - fixing RBE

WDYT?

buchgr:
Why is there a need for a roll out plan if this change is wire compatible?

ola-rozenfeld:
Because it breaks compile on clients and servers alike, and in our particular case we have a rather weird dependency tree: RBE and internal Bazel depend on internal remote-apis, while external Bazel depends on its own copy of external remote-apis. So it's hardest for us, but for example for Buildfarm upgrading to this commit it should be a very simple single patch.

buchgr:
LGTM

buchgr:
I think you need to sync it because I merged the VSO hash

PR #86, opened 2019-06-05
add ios and android as standard values for OSFamily

buchgr:


buchgr:
@juergbi @edbaunton @EdSchouten

juergbi:
Shouldn't we use `OSFamily` `linux` for Android and then distinguish between Android and other Linux distributions/variants by a new property such as `OSDistribution`?

A similar approach may or may not make sense for macOS/iOS. I'm not familiar enough with the commonalities and differences (besides the ISA).

buchgr:
I wouldn't consider Android a Linux distribution. My intention was for `android` to indicate the worker running the action on an android device (i.e. a phone).

sstriker:
We should drive this to closure.  @juergbi, does @buchgr 's comment address your question/concern?

juergbi:
> I wouldn't consider Android a Linux distribution. My intention was for `android` to indicate the worker running the action on an android device (i.e. a phone).

If Android is not considered a Linux distribution, it's not clear to me which systems should and which shouldn't use `OSFamily` `linux`. Actions with input roots that only depend on the Linux kernel syscall interface should work in any case. And if the action depends on specific userspace libraries, tools, or services, `OSFamily` alone is anyway not specific enough (whether Android is considered a separate `OSFamily` or not). I.e., `OSFamily` `linux` alone should not indicate that it's a glibc-based environment.

Using `OSDistribution` `android` also matches the withdrawn proposal [Platforms for Bazel Remote Execution](https://docs.google.com/document/d/1VemwPeyZh2rRkTGVFV18TQbyxB3FPc9iTZKpshLpNGU/edit#).

Do you have an alternative suggestion how to define `OSFamily` `linux`?

bergsieker:
@buchgr Any further thoughts here?

bergsieker:
It seems that no one cares enough about this to drive it forward in 6+ months, and we didn't have agreement on what we should do. Closing for now. If someone submits an alternative proposal we can consider that separately.

PR #87, opened 2019-07-01
Add SHA384 and SHA512 to the set of supported digest functions.

jmillikin-stripe:


build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -670,7 +670,7 @@ message SymlinkNode {
 
 // A content digest. A digest for a given blob consists of the size of the blob
 // and its hash. The hash algorithm to use is defined by the server, but servers
-// SHOULD use SHA-256.
+// SHOULD use at least SHA-256.
```
buchgr: I think this sentence could be more clear by saying servers should not use SHA-1.

buchgr:
LGTM

build/bazel/remote/execution/v2/remote_execution.proto
bergsieker: I would actually suggest dropping this altogether. Hash functions aren't strictly "linear," so "at least" doesn't really make sense. Implementors should be able to choose whatever hash function(s) they want to, and must be aware of the pros and cons of doing so.

jmillikin-stripe: Would you like me to remove that phrase in this PR, or will you do it separately?

bergsieker:
I'd go ahead and remove it in this PR, please.

jmillikin-stripe:
Done.

PR #88, opened 2019-07-01
 Add .proto specification for the remote repository cache API.

jmillikin-stripe:
This API is currently a proposal under review: https://docs.google.com/document/d/1aagKnM8kRmMvLqSXasShsI-NYOBZYtgNrbYWLPwBoBg/edit

Mailing list thread: https://groups.google.com/forum/#!topic/remote-execution-apis/Gyzvz7WkQPE

Discussion on the doc is mostly focused on whether it should be expanded to support non-file caching, such as the output of a Git clone. Since it seems like the core functionality of "download and cache a single file" is not controversial, I'd like to get that part merged into the remote APIs proto specification.

cc @EdSchouten


build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1188,6 +1212,34 @@ message UpdateActionResultRequest {
...
+  // support multiple instances of the execution system (with their own workers,
+  // storage, caches, etc.). The server MAY require use of this field to select
+  // between them in an implementation-defined fashion, otherwise it can be
+  // omitted.
+  string instance_name = 1;
+
+  // List of mirror URLs referencing the same file.
+  repeated string urls = 2;
+
+  // Expected checksum of the file downloaded, in Subresource Integrity format.
```
illicitonion: There's an interesting tension here between using SRI (which is nicely standard, and which e.g. maven .sha1 files can be used to populate even though they don't have known size, and the digest function may be different), vs using a `Digest` (which fits in nicely with the rest of the remote execution ecosystem)... If we really want to have SRI here, maybe this could become a `OneOf` field so that if you _do_ know the digest, that information can be used as an optimisation?

jmillikin-stripe: We can't use a `Digest` for the input because that would require knowing the size, which (at least in Bazel) is not something users are expected to provide for all external files.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -172,6 +172,30 @@ service ActionCache {
   }
 }
 
+// The repository cache API is used to download and cache external repositories.
```
EricBurnett: Meta comment, probably for @ola-rozenfeld : is it time to start splitting this API into multiple files? this one in particular feels like a reasonable candidate to be in its own file.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -172,6 +172,30 @@ service ActionCache {
...
+  // Errors:
+  // * `INVALID_ARGUMENT`: One or more request fields are invalid.
+  // * `NOT_FOUND`: The cache could not find content for the given request.
+  // * `RESOURCE_EXHAUSTED`: There is insufficient CAS capacity to store the
+  //   downloaded content.
+  // * `FAILED_PRECONDITION`: The cache found content at the given URLs that it
+  //   believes is valid content, but it did not match the content integrity
+  //   checksum.
+  //
+  // Repository caches may, but are not required to, verify content integrity.
```
EricBurnett: See my comment below for an alternate requirement (if any, at least one must match). I don't feel strongly about *those* semantics - another option is for the server to say which integrity formats they matched explicitly - but I do think clients need something stronger than "may" here. If a client provides an integrity string but cannot tell if it matched or was ignored, what can they do with that information?

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1188,6 +1212,34 @@ message UpdateActionResultRequest {
...
+message DownloadRequest {
+  // The instance of the execution system to operate against. A server may
+  // support multiple instances of the execution system (with their own workers,
+  // storage, caches, etc.). The server MAY require use of this field to select
+  // between them in an implementation-defined fashion, otherwise it can be
+  // omitted.
+  string instance_name = 1;
+
+  // List of mirror URLs referencing the same file.
+  repeated string urls = 2;
```
EricBurnett: Open question: Should this be URLs or URIs? Or either? Aka should this be 'https://some.host/some.resource', or 'some.host/some.resource'? 

On one hand, not all URIs will be URLs - git repos are an example where the protocol may reasonable be decoupled from the identifier. And if one knows the host, URIs should theoretically be easily mapped to URLs. On the other hand, if you want to support truly host-agnostic fetching, URLs may be required, or we'll just end up with ad-hoc standards like "for unknown hosts, HTTPS is assumed".

I'm leaning towards saying "URL" is correct, and servers may rewrite it if needed. But I'd be interested in what you all think.

build/bazel/remote/execution/v2/remote_execution.proto
EricBurnett: I think what Daniel's suggesting here makes sense, and in fact generalizes further - broadly speaking a DownloadRequest may ultimately be used for all sorts of resources, with all sorts of integrity schemes. SRI makes sense for some purposes. Digest makes sense for others (including, I'll note, as a way to request server-side cross-CAS asset population). But for a generic API we should assume that others will ultimately be desired as well, and prepare to gracefully handle that.

I see three main options for addressing that:

 1.  Make the 'integrity' field optional, and allow other options as they come as their own optional top-level fields.
 2. Use a OneOf to express the same, but be clear that they're mutually exclusive.
 3. Use an Any to express the same.

I don't really like 1) - multiple fields - because it requires a proliferation of fields, and if anyone tries to use it privately and then open up later we risk field collisions.
I don't really like 2) - oneof type - either, because integrity formats are not actually exclusive; a client that knows multiple should be able to specify both, e.g. to ease graceful transitions.
Which leads me to 3) - Any type - as my preferred option. I.e. something like:

```
message DownloadRequest {
  ...
  // Optional integrity message, for doing server-side validation that the resource to be download was
  // correctly fetched.
  // Servers MAY support any formats they wish, or none. Common formats include `Digest` and 
  // `SubresourceIntegrity`.
  // If any messages are specified, servers MUST validate at least one matched or return an error.
  // <what error specifically?>
  repeated google.protobuf.Any integrity = 3;
}

message SubresourceIntegrity {
  string integrity = 1;
}
```

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1188,6 +1212,34 @@ message UpdateActionResultRequest {
...
+  // If set, restrict cache hits to those cases where the file was added to the
+  // cache with the same canonical id
+  string canonical_id = 4;
+}
+
+// A response message for
+// [RepositoryCache.Download][build.bazel.remote.execution.v2.RepositoryCache.Download].
+message DownloadResponse {
+  // The blob digest of the downloaded file.
+  Digest digest = 1;
```
EricBurnett: In a similar vein, this assumes the downloaded asset can be expressed as a single file, and that that's the desired format. Is that a reasonable assumption?

I personally believe we'll want to support packed and unpacked outputs before too long - e.g. for URLs that point at tarballs, the Digest of the tarball itself, and/or some index of its contents pre-unpacked.

Would it then also make sense to symmetrically express this as `repeated google.protobuf.Any contents = 1`, and enable the server to support any format(s) it wishes?

I'd suggest also privileging Digest (e.g. making it its own field, with a second field for alternates), except I don't know what to do with the case where the contents are not reasonably expressed by a single digest. For that reason, I suggest we mention Digest in comments but *not* require that every download include a single Digest amongst its response types. Does that make sense?

build/bazel/remote/execution/v2/remote_execution.proto
EricBurnett: Actually, perhaps neither URL nor URI are broad enough? A file can be expressed by a URL. A git repo is probably repo+ref. Either might be augmented by a minimum fetch timestamp. A monorepo CL is potentially just a CL, but in reality could be some complicated combination of CL, branch, diffs, workspace, and who knows what else. And then a file *in* a repo is a path on top of one of those.

I'm torn between saying that a URI can encode any packed data it wants, and so suffices for all of those use-cases, and that we'll just end up inventing ugly and broken packing schemes if we go that route, making strongly typed protos preferred.

I'd be interested in a client perspective - for Bazel or others, which would be more usable in practice? Typed protos would probably require you to know the types you want to support, but maybe you need to anyways; untyped URIs probably require you to get those directly from clients, or embed URI-encoding logic. I don't know which is worse.

build/bazel/remote/execution/v2/remote_execution.proto
jmillikin-stripe: If a future PR adds support for features like cache prefill, I'd rather have that in its own RPC rather than try to multiplex a half-dozen behaviors into this one. For example, cross-CAS asset distribution would use an RPC in `ContentAddressableStorage`.

Regarding specifically `Digest` vs SRI: the purpose of the repository cache is to handle ingestion of files that are not yet present in the build cluster and, therefore, do not have a `Digest`. If the client already has a `Digest` then it does not need to call `RepositoryCache.Download()`, it can simply pass that `Digest` to `ContentAddressableStorage`.

build/bazel/remote/execution/v2/remote_execution.proto
jmillikin-stripe: I can't think of any cases in which the client would want to change its behavior based on the presence of a response field with a content checksum in it.

As I understand the security model of a distributed build cluster, either:
* The client does not trust the repository cache, and thus the client must re-validate downloaded files regardless of what the cache says. A response field with the cache's opinion on integrity would be ignored.
* The client does trust the repository cache, and does not re-validate downloaded files. A response field with the cache's opinion on integrity would be ignored.

Either way the behavior is exclusively in the domain of the client, controlled by client options like Bazel's `--noremote_verify_downloads`.

build/bazel/remote/execution/v2/remote_execution.proto
jmillikin-stripe: I think URLs. A client that requests `https://example.com/some/archive.tar.gz` is being clear about what it expects, and all of the obvious ways to download a file (HTTP, FTP, S3/GCS, Git, SCP) either identify resources by URL natively or are trivially adaptable to do so.

The `RepositoryCache.Download()` RPC concerns itself only with downloading single files. Downloading of multi-file content such a directory trees is out of scope -- so we don't need to solve the issue of encoding arbitrary revision identifiers into a URL. If there is some future person who would like to identify downloadable content by diff, then that person will have enough context on their use case to write a reasonable API.

build/bazel/remote/execution/v2/remote_execution.proto
jmillikin-stripe: This API is specifically designed for the single-file use case, such as Bazel's `repository_ctx.download()` or Buck's `http_archive()`. I want to use it as a distributed version of Bazel's `--repository_cache` flag.

There are potentially many ways that new caching behavior could be added in the future, but I think that discussion belongs in separate proposals and separate RPCs.

build/bazel/remote/execution/v2/remote_execution.proto
illicitonion: In Pants, we're looking to use `Digest` as our remote file unit of truth, and that may involve things not already in the `ContentAddressableStorage`. Our primitive here would be something like:
```
remote_file(
  digest = Digest(fingerprint=..., length=...),
  urls = [...],
)
```

but we want to be able to pass this to the remote execution system to fetch so that if the client is a mac and the execution platform is linux, we don't need to round-trip a linux-only clang via the mac which can't possibly use it, just to upload it to the remote execution server.

jmillikin-stripe: @illicitonion If you have a `Digest` then it can be converted to an SRI hash without much trouble. Bazel doesn't require users to specify sizes of remote files, so if this API was specified in terms of `Digest` it would save you a very small amount of code at the cost of not working at all for my Bazel use case.

EricBurnett: @jmillikin-stripe maybe I should take a step back. In the narrow form of _"only fetch single files"_ and _"only address content by URL+SRI"_, I don't think I'd support having this interface added here. The concerns I have boil down to:

   * In this form it'd have the effect of entrenching what I consider to be an antipattern. For bazel to take a dependency on a git repo right now it must:
      1. Be given the hash of a tarball archive of that repo. _(Cost: user experience...they're being forced to reference something derived from what they wanted to express, which was repo+hash. Also, of tangential interest: https://github.com/easybuilders/easybuild-easyconfigs/issues/5151)_
      1. Download that entire archive. _(Cost: network and time. This phase already takes whole minutes for some repos, and scales with the size of the dependencies rather than the size of the built subset.)_
      1. Unpack that entire archive. _(Cost: disk IO. Puts a floor on the minimum resources required locally at the build host.)_
      1. Load the BUILD and .bzl files contained in that archive.
      1. Re-hash everything, to calculate its digest. (Cost: disk IO and CPU).

     That's a lot of unscalable work, most of it unnecessary, and causing real-world performance bottlenecks today. I'm happy to take incremental steps towards improving it - repo fetching via a local cache would reduce the time needed for the download step, for example - but only if they enable us to keep iterating further. A single-purpose API as you describe would actually get in the way of that iteration: it wouldn't make it easier to change either end of (how dependencies are expressed; how they're returned), and using something else would then require clients to support _multiple_ APIs and pick between them.
   * In this context - this repo, and set of APIs - it feels like an awkward fit. Neither URLs nor SRI are native concepts for a gRPC API, or this ecosystem in particular. They make sense as a compatibility shim specifically for downloading files from websites, but that in turn makes it less suited for any other purpose. And this repo is all about having a small set of flexible, powerful APIs that cross clients and servers and use-cases.

But that all said, I _do_ believe we can add it in a form that both supports your purposes - indirecting downloads of repository cache files in bazel - and mine - iterating towards APIs clients can generally use to download and access remote files. For you it does come with the cost of added indirection and generality, but I'm hoping that's acceptable enough.

build/bazel/remote/execution/v2/remote_execution.proto
EricBurnett: If a client wants to ensure that either it or the server validated the files, it presumably needs to know whether the server did. So this is the "client trusts the server" case.

build/bazel/remote/execution/v2/remote_execution.proto
jmillikin-stripe: @EricBurnett OK, do you think I should move this new service spec to bazelbuild/bazel then?

This spec is not designed to solve issues with slow commands executed during repository setup, so even if I changed it to fit your goals, I would have to submit a separate proposal to get this same download-oriented API in for my purposes.

Note that for me, blocking network access from arbitrary command execution is an explicit goal and I do not want to support patterns like `git_repository()` here.

jmillikin-stripe:
Closing to move into `bazelbuild/bazel`

build/bazel/remote/execution/v2/remote_execution.proto
ulfjack: From a client perspective, I think URL-encoding would be delegated to Starlark, which avoids requiring Bazel to be aware of the possible formats. +1 for URLs.

build/bazel/remote/execution/v2/remote_execution.proto
sstriker: I agree.  Once more it raises the question of separating out CAS as well, as it and the Digest, Directory, Tree types are probably the most reused in other contexts.

build/bazel/remote/execution/v2/remote_execution.proto
EricBurnett: I'm not sure I understand your concern - making the API more general does not require you to actually *implement* support for anything else, and certainly not git repos or multi-file returns. I imagined you'd still only implement support for accepting SRI and returning the file's Digest, as you desire, and in the first cut that's all bazel would implement either. I did not intend to suggest you change what features you implement (now or later), beyond how they're expressed in the on-the-wire protos.

But that all said, if you prefer this as a narrow bazel API to solve your specific pain point, I think bazelbuild/bazel is a fine home for it. Someone on the bazel side will have to ok it though, I cannot speak for them.

PR #91, opened 2019-08-15
Add NodeProperty message to filetype nodes

### LINKED ISSUE #90: Add properties for files
In order to preserve file properties such as timestamps and permission bits as metadata, I propose to add the concept of  `NodeProperties` as repeated key/values to FileNodes, DirectoryNodes, and SymlinkNodes. This would allow servers to enrich directory trees and is similar to the Platform properties in #38. For example in directory proto:
 ```json
{
  files: [
    {
      name: "bar",
      digest: {
        hash: "4a73bc9d03...",
        size: 65534
      },
      "properties": [
        {
          "mtime": "2019-08-12 10:00:00.000000000 +0100 "
        }
      ]       
    }
  ]
 }
```

### PR DESCRIPTION
**Pull request by traveltissues**
closes #90

If there's interest in this change it can be signed off under the Codethink Ltd cla

### CODE REVIEW

File: build/bazel/remote/execution/v2/nodeproperties.md [hunk 1]
```diff
@@ -0,0 +1,54 @@
+# Node Property Lexicon
+
+This lexicon defines standard property names and values that servers MAY
+support in the `NodeProperty` message.
+
+The following standard property `name`s are defined:
```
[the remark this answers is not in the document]
traveltissues: yes, I think so

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -628,6 +641,25 @@ message Directory {
   repeated SymlinkNode symlinks = 3;
 }
 
+// A single property for [FileNodes][build.bazel.remote.execution.v2.FileNode],
+// [DirectoryNodes][build.bazel.remote.execution.v2.DirectoryNode], and
+// [SymlinkNodes][build.bazel.remote.execution.v2.SymlinkNode]. The server is
+// responsible for specifying the property `name`s that it accepts. If
+// permitted by the server, the same `name` may occur multiple times.
+//
+// The server is also responsible for specifying the interpretation of
```
ola-rozenfeld: I think this paragraph is no longer true. All the properties are rigorously defined in the md file.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -640,6 +672,9 @@ message FileNode {
 
   // True if file is executable, false otherwise.
   bool is_executable = 4;
+
+  // The FileNode properties
```
ola-rozenfeld: Rename here and everywhere: node_properties.

ola-rozenfeld: Comment here and elsewhere: the node properties of the <thing>.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -912,6 +953,9 @@ message OutputFile {
   // message. The server MAY omit inlining, even if requested, and MUST do so if inlining
   // would cause the response to exceed message size limits.
   bytes contents = 5;
+
+  // The available properties of the file
```
ola-rozenfeld: Here and in other output node_properties: "The supported node properties of the <thing>, if requested by the Action."

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1449,6 +1499,9 @@ message ExecutionCapabilities {
 
   // Supported execution priority range.
   PriorityCapabilities execution_priority_capabilities = 3;
+
+  // List of supported node properties
+  repeated string node_property_capabilities = 4;
```
ola-rozenfeld: Rename: supported_node_properties.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -571,6 +575,9 @@ message Platform {
 // * The files, directories and symlinks in the directory must each be sorted
 //   in lexicographical order by path. The path strings must be sorted by code
 //   point, equivalently, by UTF-8 bytes.
+// * The [NodeProperties][build.bazel.remote.execution.v2.DirectoryNode] of
```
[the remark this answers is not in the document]
traveltissues: oops, thanks

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -410,6 +410,10 @@ message Action {
   // If true, then the `Action`'s result cannot be cached, and in-flight
   // requests for the same `Action` may not be merged.
   bool do_not_cache = 7;
+
+  // List of required supported [NodeProperty][build.bazel.remote.execution.v2.NodeProperty]
+  // keys. If a property is not recognized the server returns INVALID_ARGUMENT.
+  repeated string output_properties = 8;
```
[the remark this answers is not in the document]
traveltissues: yes, that's better

build/bazel/remote/execution/v2/remote_execution.proto
traveltissues: yes, i think so

build/bazel/remote/execution/v2/remote_execution.proto
traveltissues: yes, it's probably better to have a less generic name

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -412,6 +412,10 @@ message Action {
...
```
bergsieker: Is there a sorting requirement here? I don't necessarily think we need one, but we should be explicit about that if so.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -412,6 +412,10 @@ message Action {
   // If true, then the `Action`'s result cannot be cached, and in-flight
   // requests for the same `Action` may not be merged.
   bool do_not_cache = 7;
+
+  // List of required supported [NodeProperty][build.bazel.remote.execution.v2.NodeProperty]
+  // keys. If a property is not recognized the server returns INVALID_ARGUMENT.
```
bergsieker: Can we beef out this description a bit? In particular, I think the server should have leeway to interpret these properties how it sees fit. For example, some properties may be applicable to directories but not files (or vice-versa).

build/bazel/remote/execution/v2/remote_execution.proto
ola-rozenfeld: Oh, good catch, these absolutely have to be sorted. Note this is part of the Action, so order affects the cache key.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1449,6 +1501,9 @@ message ExecutionCapabilities {
 
   // Supported execution priority range.
   PriorityCapabilities execution_priority_capabilities = 3;
+
+  // Supported node properties
```
ola-rozenfeld: Period in the end of line, please.

traveltissues:
thanks @ola-rozenfeld @bergsieker

EdSchouten:
How does one specify properties on the root directory of the input root?

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -573,6 +583,9 @@ message Platform {
 // * The files, directories and symlinks in the directory must each be sorted
 //   in lexicographical order by path. The path strings must be sorted by code
 //   point, equivalently, by UTF-8 bytes.
+// * The [NodeProperties][build.bazel.remote.execution.v2.NodeProprty] of files,
```
EdSchouten: This contains a typo.

ola-rozenfeld:
> How does one specify properties on the root directory of the input root? Wouldn't it have been smarter to place the properties inside the Directory message itself?

/facepalm
You are 100% right, this should be in the Directory message! Will send a PR to fix.

> I also wonder how this will impact the size of directory objects, considering that all of this (even names of keys) is done through strings.

Yeah, I suspect it will have a big impact, my hope basically is that the need for this will be a rare case, and that clients will use these only as needed.

EdSchouten:
> Yeah, I suspect it will have a big impact, my hope basically is that the need for this will be a rare case, and that clients will use these only as needed.

Then my hope is that we don't follow through on what is requested in #99.

ola-rozenfeld:
https://github.com/bazelbuild/remote-apis/pull/102

PR #95, opened 2019-09-04
Clarify description of `GetTreeRequest.page_token`

santigl:
This changes the description of  `GetTreeRequest.page_token`, seeking to make it more clear that the `page_token` value represents an offset in the stream.


EricBurnett:
+1 from me.

PR #102, opened 2019-10-22
Moving node properties of a directory to the Directory message

ola-rozenfeld:


ola-rozenfeld:
@EdSchouten, @mostynb

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -689,9 +692,6 @@ message DirectoryNode {
   // represented. See [Digest][build.bazel.remote.execution.v2.Digest]
   // for information about how to take the digest of a proto message.
   Digest digest = 2;
-
-  // The node properties of the DirectoryNode.
-  repeated NodeProperty node_properties = 3;
```
EdSchouten: I think these properties may also be removed from one or two other places (e.g., output directories).

ola-rozenfeld: Right, thanks!

ola-rozenfeld:
@traveltissues, PTAL. Thank you!

traveltissues:
https://github.com/bazelbuild/remote-apis/pull/102/files#diff-4153f76ba92d8d30764c0251177105e8R655
doesn't seem to apply anymore

build/bazel/remote/execution/v2/remote_execution.proto
traveltissues: I understand why properties should be added to the directory message but why are these being removed?

juergbi: Otherwise all directories except the top-level one would have two `node_properties` lists, and we should only need one per directory.

traveltissues: ah, i see

PR #106, opened 2019-11-12
point out that servers may modify ActionResults, and mention one reason why

mostynb:
When serialized, default ActionResult proto3 messages contain no data. This is unfortunate because there is no way to distinguish this valid data from from file truncation errors on the server (which are perhaps the most common form of file-based data corruption).

As a workaround, server implementations can modify incoming ActionResult messages to ensure that at least one field has a non-default value, and then consider any default/zero-length ActionResult messages retrieved from storage as corrupt. Let's point this out in the documentation.

The simplest field that servers may consider modifying is ActionResult.execution_metadata.worker, which (if empty) can safely be set to the IP address or hostname of the creator/uploader without modifying the semantic meaning of the ActionResult.

### DIFF
--- a/build/bazel/remote/execution/v2/remote_execution.proto
+++ b/build/bazel/remote/execution/v2/remote_execution.proto
@@ -161,6 +161,9 @@ service ActionCache {
   // [Command][build.bazel.remote.execution.v2.Command], into the
   // `ContentAddressableStorage`.
   //
+  // Server implementations MAY modify the
+  // `UpdateActionResultRequest.action_result` and return an equivalent value.
+  //
   // Errors:
   //
   // * `INVALID_ARGUMENT`: One or more arguments are invalid.
@@ -853,6 +856,11 @@ message ExecutedActionMetadata {
 
 // An ActionResult represents the result of an
 // [Action][build.bazel.remote.execution.v2.Action] being run.
+//
+// It is advised that at least one field (for example
+// `ActionResult.execution_metadata.Worker`) have a non-default value, to
+// ensure that the serialized value is non-empty, which can then be used
+// as a basic data sanity check.
 message ActionResult {
   reserved 1; // Reserved for use as the resource name.
 


EricBurnett:
I'm not sure I follow the semantics here - can you clarify for me?

- If a server is defending against their own corruptions (they both add the field, and then validate that the message is non-empty), couldn't they just wrap the message in a non-empty wrapper of their own? I think that'd be equivalent - preventing themselves from wrapping an empty message would be equivalent to preventing themselves from injecting a field into an empty message.
 - If the client is the one who should reject empty messages, when is it safe to do so?

I'm wondering if we can go further now and simply say something to the effect of "clients SHOULD reject empty protos; in the event that an ActionResult is legitimately empty - (containing no outputs) - servers MUST set at least one field, for example `execution_metadata.worker`."

This is one of those technically-backwards-incompatible changes that likely affects no implementations in reality; I _think_ the policy we were going with is that we'd allow such semantic tightenings as long as nobody objected but @ola-rozenfeld to confirm.

If not, I'm not sure we can even suggest that clients SHOULD reject such messages.

mostynb:
> I'm not sure I follow the semantics here - can you clarify for me?
> 
> * If a server is defending against their own corruptions (they both add the field, and then validate that the message is non-empty), couldn't they just wrap the message in a non-empty wrapper of their own? I think that'd be equivalent - preventing themselves from wrapping an empty message would be equivalent to preventing themselves from injecting a field into an empty message.

That would be equivalent from the point of view of the server, yes.

But there are a few reasons why I think it makes sense to use existing fields:
* Clients & execution implementations can choose to set some of these optional fields themselves now, since they might become required in REAPIv3 anyway.
* This metadata is probably useful info (null digests were suggested in the meeting yesterday, but I think this is useless info and larger than the alternative suggestions).
* Wrapping protobufs would make debugging cache contents slightly more difficult.

> * If the client is the one who should reject empty messages, when is it safe to do so?

Clients should not reject empty messages in REAPIv2.

If clients start rejecting empty messages now, then I think we will see interoperability issues, since different ActionCache and Execution services might not implement this recommendation.

This advice is purely for the server side- when receiving ActionResult messages store them in a way that has non-zero size, and when processing GetActionResult requests from clients check that stored ActionResult messages have non-zero size before returning them to the client. If the server finds a zero-size ActionResult in its storage, it should be assumed to be corrupt and a NOT_FOUND result should instead be sent to the client.

This is not as robust as if we could assume the same behaviour on the client side- it only protects against server-side bugs (at least two separate servers seem to have suffered from this symptom). In REAPIv3 if non-zero ActionResult protos are enforced on the client side too, then we can also catch some client-side and transmission bugs.

> I'm wondering if we can go further now and simply say something to the effect of "clients SHOULD reject empty protos; in the event that an ActionResult is legitimately empty - (containing no outputs) - servers MUST set at least one field, for example `execution_metadata.worker`."
> This is one of those technically-backwards-incompatible changes that likely affects no implementations in reality; I _think_ the policy we were going with is that we'd allow such semantic tightenings as long as nobody objected but @ola-rozenfeld to confirm.
> 
> If not, I'm not sure we can even suggest that clients SHOULD reject such messages.

I don't think we can do this until REAPIv3.

sstriker:
@mostynb This has been open for a while, are you still interested in inclusion of the language?

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -160,6 +160,12 @@ service ActionCache {
   // [Command][build.bazel.remote.execution.v2.Command], into the
   // `ContentAddressableStorage`.
   //
+  // Server implementations are advised to modify the uploaded ActionResult to
```
sstriker, line 163: I do struggle with encouraging modification of ActionResult.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -786,6 +792,14 @@ message ExecutedActionMetadata {
...
 // An ActionResult represents the result of an
 // [Action][build.bazel.remote.execution.v2.Action] being run.
+//
+// It is advised that at least one non-default field value is set, to
+// differentiate the serialized value from no data. Suitable fields include
+// `ActionResult.execution_metadata.Worker` which could be set to the IP
+// address or hostname of the client that created the ActionResult, any other
+// fields of `ActionResult.execution_metadata`, or the null Digest value
+// for either `ActionResult.stdout_digest.hash` or
+// `ActionrResult.stderr_digest.hash` if all other fields are default values.
```
mostynb, line 802: There's an "ActionrResult" typo here.

build/bazel/remote/execution/v2/remote_execution.proto
mostynb: It's implicit that ActionResults can be modified, otherwise why does UpdateActionResult return an ActionResult?

Would it help if we suggest execution_metadata.worker here too, as a safe and convenient value to modify?

build/bazel/remote/execution/v2/remote_execution.proto
mostynb, line 802: Given the discussion around empty blob digests in #134, I would rather not suggest using the null/empty blob digest values for this purpose.

build/bazel/remote/execution/v2/remote_execution.proto
sstriker: > It's implicit that ActionResults can be modified, otherwise why does UpdateActionResult return an ActionResult?

Actually this raises a good point.  To me it isn't implicit at all that ActionResults can be modified.  It could be returning the ActionResult that was replaced by the new ActionResult.

Regarding modification, I would rather that the ActionCache service is explicitly not allowed to modify ActionResult.  When when we start considering signing of ActionResult, we won't have to worry about ActionCache.

build/bazel/remote/execution/v2/remote_execution.proto
mostynb: Removed.

build/bazel/remote/execution/v2/remote_execution.proto
mostynb: Updated the text, so that it's simpler and softer, but still hopefully useful to server implementors.

I was also confused by the fact that `UpdateActionResult` returns an `ActionResult`, then I stumbled upon this (note the discussion about mutable resource properties): https://cloud.google.com/apis/design/standard_methods#update

EricBurnett: > Regarding modification, I would rather that the ActionCache service is explicitly not allowed to modify ActionResult. When when we start considering signing of ActionResult, we won't have to worry about ActionCache.

If you mean workers signing what they did, I don't think signing ActionResult as-is is the way to go. Some fields - like stdout_raw, execution_metadata.worker, execution_metadata.queued_timestamp - are explicitly intended for servers to set, either in general or conditional on the cache hit request. And that will likely expand, as ActionResult is a communication message used directly in RPC responses. But more concerningly, ActionResult doesn't describe the action itself, and any use-case for worker signing where the server is untrusted presumably requires the worker sign _what they ran_, not just the result of it, otherwise you're subject to replay attacks of various sorts.

Whatever workers sign should be

a.    self-contained: no extraneous fields, no optional fields introducing ambiguity, well-defined serialization, etc
b.    complete: cover everything material to execution, including inputs, outputs, command, environment changes, etc
c.    passed through the API as-is without modification alongside the corresponding signature.

I'm not sure how exactly to achieve this, especially with the additional complication that REAPI isn't even a worker-facing API (the APIs in use to communicate with workers may not reuse any messages from it - RWAPI reuses the CAS but none of Execution, for example - but regardless, I'm pretty sure signing ActionResult as-is is not going to be the way to go, so it being flagged as server-modifiable is probably useful in not misleading anyone in that direction.

sstriker: @EricBurnett I don't think we are disagreeing on signing ActionResult as-is, and yes, at a minimum the Action digest should be included in the ActionResult.  This would also solve the defaults-only ActionResult problem.  We can take the ActionResult signing discussion to a separate issue (eg. #133) though :)

To bring it back to this issue, I would prefer to keep ActionCache as a simple store/retrieve service.

I would be fine with a description along the lines of:
"The server MAY return an equivalent ActionResult as a response.  The client SHOULD use this equivalent ActionResult for inputs to dependent Actions remaining to be executed."

This does not imply that the ActionCache service munges ActionResults on the way in.  It does imply that ActionCache may return a "better" result for the client to use moving forward.  Now, how this "better" result comes to be is an implementation detail.  It may be selective based on which process produced the result for the action, or when, or...

What I like about this phrasing is that there is something useful for the client in the response as well.

Hope that clarifies my thinking a little bit.

mostynb: > at a minimum the Action digest should be included in the ActionResult. This would also solve the defaults-only ActionResult problem

Many Actions can produce the same ActionResult, so it would have to be a repeated field I guess.

> I would be fine with a description along the lines of:
> "The server MAY return an equivalent ActionResult as a response. The client SHOULD use this equivalent ActionResult for inputs to dependent Actions remaining to be executed."

This part sounds fine to me. I would also like to hint that zero-valued ActionResults should be avoided (though I have not yet found another cache implementor willing to comment here).

sstriker: @mostynb I think that language allows for server implementations to not store zero-valued ActionResults as-is.

Would you like to add something like:
- In UpdateActionResult "Clients SHOULD not send zero-sized ActionResults"
- In GetActionResult "Servers SHOULD not return zero-sized ActionResults"

Or do you have something completely different in mind?

mostynb: Currently on master, there is no mention of the UpdateActionResult return value here. So I assume that the "Server implementations MAY modify the ActionResult and return an equivalent value" part is uncontroversial.

The second part of this PR is merely a hint for server implementors as to why they might want to do so- as a way to catch the most common form of data corruption while the ActionResult is in storage (and ActionResult.execution_metadata.Worker is the simplest field to use for this purpose). How about something like:

`Servers MAY modify fields such as ActionResult.execution_metadata.Worker, to ensure that  the serialized value of the ActionResult is non-empty during processing and storage.`

But even "MAY" feels too strong IMO. Would you prefer that I drop this part entirely?

PR #111, opened 2020-01-15
Move the deps handling, for go explicitly. Though the api woudln't ch…

### LINKED ISSUE #110: switched_rules_by_language doesn't fully work in not requiring per-platform deps
If you try use this without go/having go locally it will fail that gazelle isn't defined. I think because the load definition at 

https://github.com/bazelbuild/remote-apis/blob/master/remote_apis_deps.bzl#L3

ianoc-stripe:
…ange for more languages.

fixes #110


repository_rules.bzl
```diff
@@ -6,9 +6,55 @@ This is adapted from
 https://github.com/googleapis/googleapis/blob/master/repository_rules.bzl
 """
 
-load("//:remote_apis_deps.bzl", "remote_apis_go_deps")
 
 def _switched_rules_impl(ctx):
+    go_deps = """
```
ola-rozenfeld: Is there a way to make this an actual bzl file that we include on demand, rather than a string inside this one? That way, it formats/lints nicely in various IDEs.

bergsieker: I concur with Ola: Ideally this would be a standalone bzl file, rather than something generated from a string.

ianoc-stripe: Sorry about the delay on this -- missed these comments somehow. Updated the PR not to inline the content

sstriker:
@bergsieker could you give this another look?

mostynb:
I think this PR has been made obsolete by #307.

PR #112, opened 2020-01-19
Add Remote Asset API

sstriker:
Based on the [design document](https://docs.google.com/document/d/10ari9WtTTSv9bqB_UU-oe2gBtaAA7HyQgkpP-RFP80c)
and updates in bazelbuild/proposals#160.


sstriker:
@jmillikin-stripe @EricBurnett please have a look.  Alternatively we fold it all in the remote_execution.proto.

build/bazel/remote/fetch/v1/remote_fetch.proto
```diff
@@ -0,0 +1,442 @@
...
+  //
+  // Servers *MAY* fetch content that they do not already have cached, for any
+  // URLs they support.
+  //
+  // Servers *SHOULD* ensure that referenced files are present in the CAS at the
+  // time of the response, and (if supported) that they will remain available
+  // for a reasonable period of time. In the event that a client receives a
+  // reference to content that is no longer present, it *MAY* re-issue the
+  // request with `oldest_content_accepted` set to a more recent timestamp than
+  // the original attempt, to induce a re-fetch from origin.
```
sstriker: @jmillikin-stripe, I believe this is one of the few places where the proposals differ, as per https://github.com/bazelbuild/proposals/pull/160/files#diff-3bdfd5c462602da5c73b9b00c1696223R116

Is there a reason you made the guarantee stronger?

jmillikin-stripe: The fetch API isn't very useful if the server is allowed to reply "OK I have added your data to the CAS" when it has, in fact, _not_ added that data. There's no way for the client to handle that situation gracefully, and in Bazel it causes a big stack trace if downloading a just-enCASed file fails.

sstriker: The text is actually inspired by [ActionCache.GetActionResult](https://github.com/bazelbuild/remote-apis/blob/master/build/bazel/remote/execution/v2/remote_execution.proto#L140).  I would like for both to be **MUST**, but realistically, this is going to be hard to guarantee without amending the [ContentAddressableStorage](https://github.com/bazelbuild/remote-apis/blob/master/build/bazel/remote/execution/v2/remote_execution.proto#L234) lifetime requirements.  As such we can only suggest a strong **SHOULD**.

Clients should always handle errors gracefully, at least more graceful than a big stack trace like a helpful error message to the user.  The client can also retry, or fall back to direct download if allowed.

sstriker: I've updated the language slightly, to include that the server should extend lifetime of referenced blobs where necessary and applicable, mirroring the language of ActionCache.GetActionResult more closely.

EricBurnett:
@googlebot I consent

jmillikin-stripe:
@googlebot I consent

buchgr:
Huge congrats @jmillikin-stripe @sstriker @EricBurnett for shipping the API!

jmillikin-stripe:
FYI: While implementing this in Bazel I discovered that some of the quotes in the comment blocks have been replaced with Unicode versions. This makes ASCII-only consumers, including the Bazel CI runners, very unhappy.

I'll fix this for Bazel in the initial vendor. Can someone with a commit bit fix it in this repo, so other clients don't hit the same problem?

Example error (from https://storage.googleapis.com/bazel-untrusted-buildkite-artifacts/beb93ad4-5b4f-43c0-925c-2a6a2b6d73d7/src/test/shell/bazel/bazel_bootstrap_distfile_test/attempt_1.log):
```
derived/src/java/build/bazel/remote/asset/v1/PushGrpc.java:168: error: unmappable character for encoding ASCII
     * and only return resources that can be Fetch???d from origin.
```

philwo:
@jmillikin-stripe Sure, I can merge a PR that fixes that.

PR #113, opened 2020-01-22
Clarify that servers should extend TLL for blobs referenced in FindMissingBlobs

sstriker:
* build/bazel/remote/execution/v2/remote_execution.proto 
  (ContentAddressableStorage.FindMissingBlobs): Clarify that servers
  should extend lifetime for blobs requested.

### DIFF
--- a/build/bazel/remote/execution/v2/remote_execution.proto
+++ b/build/bazel/remote/execution/v2/remote_execution.proto
@@ -246,6 +246,9 @@ service ContentAddressableStorage {
   // Clients can use this API before uploading blobs to determine which ones are
   // already present in the CAS and do not need to be uploaded again.
   //
+  // Servers SHOULD increase the TTLs of the referenced blobs if necessary and
+  // applicable.
+  //
   // There are no method-specific errors.
   rpc FindMissingBlobs(FindMissingBlobsRequest) returns (FindMissingBlobsResponse) {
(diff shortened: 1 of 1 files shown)


build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -246,6 +246,9 @@ service ContentAddressableStorage {
   // Clients can use this API before uploading blobs to determine which ones are
   // already present in the CAS and do not need to be uploaded again.
   //
+  // Servers SHOULD increase the TTLs of the referenced blobs if necessary and
```
mostynb: nitpick: "TTL" sounds like an implementation detail. Would it make sense to change "TTLs" to "lifetimes" (both here and the GetActionResult comment that already exists in this doc)?

buchgr: GetActionResult seems to use TTLs? Generally I agree that we should standardize on one wording. I don't have a preference either way.

sstriker: Can we fix terminology in a separate PR?  I would prefer to get the intent across regardless :).

mostynb: Yes, of course. This is just a "nitpick".

bergsieker: I prefer "lifetime" to "TTL" but agree that the terminology change can be made separately

PR #115, opened 2020-02-04
use ASCII apostrophes in remote_asset.proto

mostynb:
IIRC someone reported that unicode apostrophes break some software?


jmillikin-stripe:
> IIRC someone reported that unicode apostrophes break some software?

That was me -- it broke the Bazel CI build when I tried to use the newer version in https://github.com/bazelbuild/bazel/pull/10622.

jmillikin-stripe:
Also it looks like there's still a few left. Would it be possible to add a test to this repository that rejects PRs containing non-ASCII text in the .proto schemas?

mostynb:
I noticed the other non-ascii characters the other day: https://github.com/bazelbuild/remote-apis/pull/119

sstriker:
A check in hooks/pre-commit would have caught it.

Apologies for the fall-out.

PR #128, opened 2020-03-19
Field-based node properties

juergbi:
This adds support for field-based node properties to reduce the size of directory trees compared to string-based node properties, as discussed in the monthly meeting. `MTime` and `UnixMode` are promoted to fields and the node property lexicon is removed.

The change is not backward-compatible for clients/servers that already use string-based node properties. I've followed this approach for the following reasons:
- The API structure would be more complex/confusing if we kept backward compatibility (require separate fields for string-based and field-based node properties in many places, couldn't use `node_properties` as name for the new fields)
- Migrating clients from e.g. the string-based `MTime` node property to the `mtime` protobuf field is a breaking change anyway, so streamlining the structure at the same time shouldn't be a major concern.
- Node properties are not widely used yet


build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -686,7 +685,8 @@ message Directory {
   repeated SymlinkNode symlinks = 3;
 
   // The node properties of the Directory.
-  repeated NodeProperty node_properties = 4;
+  reserved 4;
```
ulfjack: What would be the harm in leaving node_properties as-is, and just adding another field for the field-based ones? That would allow clients that want to be compatible to set both, and servers that want to be compatible to accept both.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -702,6 +702,48 @@ message NodeProperty {
...
+// [SymlinkNodes][build.bazel.remote.execution.v2.SymlinkNode]. The server is
+// responsible for specifying the properties that it accepts.
+//
+message NodeProperties {
+  // A list of string-based
+  // [NodeProperties][build.bazel.remote.execution.v2.NodeProperty].
+  repeated NodeProperty properties = 1;
+
+  // The keys of node properties whose values are valid in this message.
+  // This is required as protobuf doesn't provide a way to distinguish between
```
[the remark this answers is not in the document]
juergbi: Am I misunderstanding the documentation?

> Note that for scalar message fields, once a message is parsed there's no way of telling whether a field was explicitly set to the default value (for example whether a boolean was set to false) or just not set at all: you should bear this in mind when defining your message types. For example, don't have a boolean that switches on some behaviour when set to false if you don't want that behaviour to also happen by default. Also note that if a scalar message field is set to its default, the value will not be serialized on the wire.

https://developers.google.com/protocol-buffers/docs/proto3#default

build/bazel/remote/execution/v2/remote_execution.proto
juergbi: We would need the same two fields (`repeated NodeProperty` and the new `NodeProperties`) in 5 places in the REAPI and more in downstream protos (e.g., BuildBox uses an additional proto for communication between local components). And the same issue for `NodePropertyKeys`. Code that handles node properties couldn't simply work with a single message. And what name would we choose for the new fields? I.e., it's definitely possible but it makes everything a bit less straight forward long term, in my opinion.

> That would allow clients that want to be compatible to set both, and servers that want to be compatible to accept both.

I don't think we should ever set both variants for the same property (non-canonical, even larger tree size, possibly conflicting redundant values). It would theoretically be possible to use server capabilities to use one or the other approach. I'm not convinced that this would work well in practice, though, as node properties are part of the directory tree, which may also be cached locally. I.e., you might not regenerate the whole tree for a RE action, so you also don't want to tie it to a particular RE server.

Do you already use node properties yourself and are concerned about migration?

build/bazel/remote/execution/v2/remote_execution.proto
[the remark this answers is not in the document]
ulfjack: Ah, I see. You can tell whether a non-scalar field (i.e., message field) is set or not. I didn't check for scalar fields. Below, that would apply to unix_mode, but not mtime.

I really don't like specifying the keys in addition to the fields - it seems that that will cancel out most of the expected size benefits. Would it be possible to always require these to be wrapped in a proto message instead, i.e., forbid scalar fields?

build/bazel/remote/execution/v2/remote_execution.proto
ulfjack: I just started to work on a Bazel change to specify node properties in some cases to pass additional information to the remote execution server. Technically, if I wait until this change goes in, I could use the new fields instead.

I think what you say makes sense. It would be good to promote some of your reply into the commit message (I think your wording here is clearer than what's there already).

build/bazel/remote/execution/v2/remote_execution.proto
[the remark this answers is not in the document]
juergbi: It would definitely be nice to avoid the need for the extra keys message with boolean fields. I just discovered `wrappers.proto` with `Int32Value` and similar wrapper messages. That seems like a perfect fit and I'll change this branch accordingly.

The `NodePropertyKeys` message will still be needed for `output_node_properties` and capabilities but not in the directory tree.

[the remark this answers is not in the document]
juergbi: I've updated the branch. Does this look good to you now?

build/bazel/remote/execution/v2/remote_execution.proto
juergbi: I've expanded the commit message.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -702,6 +703,38 @@ message NodeProperty {
...
+
+  // The file's last modification timestamp.
+  google.protobuf.Timestamp mtime = 2;
+
+  // The UNIX file mode, e.g., 0755.
+  google.protobuf.UInt32Value unix_mode = 3;
+}
+
+// A set of node property keys including string-based and field-based variants.
+message NodePropertyKeys {
```
juergbi: Now that the values of field-based properties are always embedded messages (to distinguish between the absence of a value and the default value), `NodePropertyKeys` is no longer needed in the directory trees and thus, space efficiency for specifying the set of node property keys is less of a concern.

We could consider dropping `NodePropertyKeys` and keep using `repeated string` fields for capabilities and `output_node_properties` even for field-based properties. This would match the approach taken by the protobuf `FieldMask` concept.

The only downside I see is that mixing field names and names of string-based node properties might be confusing. If field names always start with a lowercase letter and string-based node property names always start with an uppercase letter, there shouldn't be any real conflicts, though.

Any thoughts on this?

ulfjack: I think I would stick with the simpler approach (repeated string) for now.

juergbi: Ok, I've switched back to the simpler `repeated string` approach, dropping `NodePropertyKeys`.

sstriker:
@ulfjack if you have no remaining concerns after the changes, I'd like us to land this.

ulfjack:
It seems fine to merge this.

PR #129, opened 2020-04-03
List EngFlow as another implementation

ulfjack:
Also clearly mark which ones are open source, and which are commercial ones.

If you can't accept this for policy reasons, I'd strongly recommend clarifying the policy around which implementations are listed here and also *being transparent* in the documentation that this is the case.


EricBurnett:
LGTM!

PR #142, opened 2020-06-05
Drop canonicalization requirement for symlink targets

juergbi:
This allows use of symlink targets created on the filesystem by existing
build tools and build systems without an extra REAPI-specific
canonicalization step.

The same build tool/system is still expected to create identical symlink
targets when executing the same Action again. I.e., this should not
result in different digests for logically equivalent Actions in common
cases.

As servers may but are not required to reject non-canonical input trees,
removing this canonicalization requirement for symlink targets reduces
the allowed differences in behavior between server implementations,
potentially improving interoperability.

### DIFF
--- a/build/bazel/remote/execution/v2/remote_execution.proto
+++ b/build/bazel/remote/execution/v2/remote_execution.proto
@@ -772,8 +772,11 @@ message SymlinkNode {
   // The target path can be relative to the parent directory of the symlink or
   // it can be an absolute path starting with `/`. Support for absolute paths
   // can be checked using the [Capabilities][build.bazel.remote.execution.v2.Capabilities]
-  // API. The canonical form forbids the substrings `/./` and `//` in the target
-  // path. `..` components are allowed anywhere in the target path.
+  // API. `..` components are allowed anywhere in the target path as logical
+  // canonicalization may lead to different behavior in the presence of
+  // directory symlinks (e.g. `foo/../bar` may not be the same as `bar`).
+  // To reduce potential cache misses, canonicalization is still recommended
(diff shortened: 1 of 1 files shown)


juergbi:
I don't expect any implementation to actually benefit from the symlink target canonicalization requirement. If I'm mistaken, please comment.

ulfjack:
As discussed in the meeting, non-canonical paths can cause cache misses / cache segmentation (and maybe also add that `foo/../bar` may not be the same as `bar/` in the presence of symlinks). I think adding a warning about that here would be good. Otherwise I have no objection to this change.

I think there was a desire to get this in before tagging 2.1.

@sstriker

sstriker:
> As discussed in the meeting, non-canonical paths can cause cache misses / cache segmentation (and maybe also add that `foo/../bar` may not be the same as `bar/` in the presence of symlinks). I think adding a warning about that here would be good. Otherwise I have no objection to this change.

Thanks @ulfjack.  What did you have in mind?  "Non-canonical and non-deterministically produced paths may lead to cache misses / segmentation." ?
@juergbi would that work?
 
> I think there was a desire to get this in before tagging 2.1.

I think that is correct.

juergbi:
I've expanded the comment to:

> `..` components are allowed anywhere in the target path as logical canonicalization may lead to different behavior in the presence of directory symlinks (e.g. `foo/../bar` may not be the same as `bar`). To reduce potential cache misses, canonicalization is still recommended where this is possible without impacting correctness.

Is this clear enough or do either of you have any further suggestions?

ulfjack:
Sounds good to me!

PR #146, opened 2020-06-25
Add MurmurHash3 digest function

erikma:
Add the MurmurHash v3 variant as a digest function an RE implementation can choose to implement.

### DIFF
--- a/build/bazel/remote/execution/v2/remote_execution.proto
+++ b/build/bazel/remote/execution/v2/remote_execution.proto
@@ -1530,6 +1530,11 @@ message DigestFunction {
 
     // The SHA-512 digest function.
     SHA512 = 6;
+
+    // Murmur3 128-bit digest function, x64 variant. Note that this is not a
+    // cryptographic hash function and its collision properties are not strongly guaranteed.
+    // See https://github.com/aappleby/smhasher/wiki/MurmurHash3 .
+    MURMUR3 = 7;
   }
 }
 


erikma:
Background: The Microsoft Office and BuildXL teams got a significant performance win moving from VSO-Hash and SHA256 to Murmur3. Murmur is not a crypto hash but instead is intended specifically for content hashing and has a tight, fast inner loop.

mostynb:
There are three variants of MurmurHash3, should this refer to a specific one? (MurmurHash3_x64_128?)

EricBurnett:
LGTM. I'll note that in some contexts (I'd have assumed including MS Office CI) collision resistance is a necessary security property for the CAS to hold, for which Murmur hashes are not a good choice. But if you're happy that that's not required in your context, seems like it should at least be fast! LGTM either way - collision resistance is not required for inclusion here, as MD5 clearly shows :).

EricBurnett:
Actually, please also make the capitalization consistent with the other values. Thanks!

EdSchouten:
There is a risk that if people would blindly start using this in multi-tenant environments, they open the door to all sorts of spooky security issues.

Note that there are modern cryptographic hash functions that are faster than SHA-256, such as BLAKE3. Maybe those strike a better balance between performance and safety?

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1530,6 +1530,10 @@ message DigestFunction {
 
     // The SHA-512 digest function.
     SHA512 = 6;
+
+    // Murmur3 digest function.
+    // See https://github.com/aappleby/smhasher/wiki/MurmurHash3 .
+    Murmur3 = 7;
```
EdSchouten, line 1536: Aren’t enum values supposed to be capitalized? Which variant of Murmur3 is being used? Wikipedia says this:

> The current version is MurmurHash3,[4][5] which yields a 32-bit or 128-bit hash value. When using 128-bits, the x86 and x64 versions do not produce the same values, as the algorithms are optimized for their respective platforms. MurmurHash3 was released alongside SMHasher—a hash function test suite.

erikma: Latest update:
- Capitalized enum per multiple comments.
- Clarified the variant of Murmur in use.
- Highlighted the non-crypto nature of the hash.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1530,6 +1530,11 @@ message DigestFunction {
 
     // The SHA-512 digest function.
     SHA512 = 6;
+
+    // Murmur3 128-bit digest function, x64/64-bit variant. Note that this is not a
```
mostynb, line 1534: "x64/64-bit" is a bit rendundant, and might be slightly confusing given the "128-bit" in the first part of the sentence. Maybe just say "x64 variant" ?

erikma: Fixed

ulfjack:
I would tend towards BLAKE2b rather than BLAKE3. My impression was that the performance improvement BLAKE3 primarily comes from multi-threading, and I'm not sure if the additional complexity is worth the cost (and in some cases, worker machines are single-core machines, so they don't see much / any benefit from multi-threading).

I would also caution that the use of non-crypto hashes increases insider risk, making it easier for a single person at the company to take over other engineering accounts, and would therefore strongly advise companies not to do that.

EdSchouten:
> I would tend towards BLAKE2b rather than BLAKE3. My impression was that the performance improvement BLAKE3 primarily comes from multi-threading

Doesn't [this graph](https://github.com/BLAKE3-team/BLAKE3/blob/master/media/speed.svg) on the BLAKE3 main page show that it's faster than BLAKE2b, even for the single-threaded case?

ulfjack:
I'm not sure I trust that graph, but after looking into it some more, that
appears to be what they claim. I find it confusing that they'd run it on a
c5.metal, which IIUC is a 96 core machine. Maybe because it is the newest
Intel arch? There's also blake2bp and blake2bs, which they did *not*
compare with, which seem like they would be similar in performance to
BLAKE3?

I like that BLAKE2 seems to be more mature with multiple conformant
implementations and an official RFC.

EdSchouten:
> I'm not sure I trust that graph, but after looking into it some more, that appears to be what they claim. I find it confusing that they'd run it on a c5.metal, which IIUC is a 96 core machine. Maybe because it is the newest Intel arch?

Probably because they simply needed a bare-metal server to rule out virtualisation overhead. Amazon doesn't offer single-core bare-metal servers.

> There's also blake2bp and blake2bs, which they did *not* compare with, which seem like they would be similar in performance to BLAKE3? I like that BLAKE2 seems to be more mature with multiple conformant implementations and an official RFC.

I like that BLAKE3 uses a Merkle tree construction instead of Merkle-Dåmgard, allowing us to have validated random access reads of CAS objects (assuming we extend the protocol to allow extraction of (parts of) the Merkle tree). That said, the DigestFunction enum is likely big enough to fit both BLAKE2 and BLAKE3. ;-)

sstriker:
Seeing enough LGTMs to land.  Suggest the BLAKE[23] discussions are moved to the list, or to an issue to add said algorithms to digest functions in a PR.

vors:
+1 on adding BLAKE[23]

PR #149, opened 2020-07-15
Add optional salt to Action

erikma:
To allow invalidating bad caches.

### DIFF
--- a/build/bazel/remote/execution/v2/remote_execution.proto
+++ b/build/bazel/remote/execution/v2/remote_execution.proto
@@ -425,6 +425,14 @@ message Action {
   bool do_not_cache = 7;
 
   reserved 8; // Used for field moved to [Command][build.bazel.remote.execution.v2.Command].
+
+  // An optional additional salt value used to place this `Action` into a
+  // separate cache namespace from other instances having the same field
+  // contents. This salt typically comes from operational configuration
+  // specific to sources such as repo and service configuration,
+  // and allows disowning an entire set of ActionResults that might have been
+  // poisoned by buggy software or tool failures.
+  bytes salt = 9;
 }
 
 // A `Command` is the actual command executed by a worker running an


build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1029,6 +1029,19 @@ message ActionResult {
...
+  // the same field contents. This salt typically comes from operational
+  // configuration specific to sources such as repo and service configuration,
+  // and allows disowning an entire set of ActionResults that might have been
+  // poisoned by buggy suftware or tool failures.
+  //
+  // As an example for repo configuration, a repo could contain a special file
+  // such as "CacheEpoch.txt" containing an incrementing number, whose hash is
+  // added to this salt value. Repo owners could increment the number when
+  // needed to invalidate its action cache.
+  Digest action_salt = 10;
```
EdSchouten, line 1044: Instead of going down this route, I actually thought that @EricBurnett's idea of moving this into Action made lots of sense. I suspect it's sufficient to just add a field like this?
```
bytes salt = ...;
```

erikma: Good point. Updating.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -425,6 +425,19 @@ message Action {
...
+  // An optional additional salt value used to place this `Action` into a
+  // separate cache universe or "epoch" from other instances having the
+  // same field contents. This salt typically comes from operational
+  // configuration specific to sources such as repo and service configuration,
+  // and allows disowning an entire set of ActionResults that might have been
+  // poisoned by buggy software or tool failures.
+  //
+  // As an example for repo configuration, a repo could contain a special file
+  // such as "CacheEpoch.txt" containing an incrementing number, whose hash is
+  // added to this salt value. Repo owners could increment the number when
```
mostynb, line 438: Just curious, what is the benefit of adding the hash of an incrementing number, compared to adding the number itself?

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -425,6 +425,19 @@ message Action {
   bool do_not_cache = 7;
 
   reserved 8; // Used for field moved to [Command][build.bazel.remote.execution.v2.Command].
+
+  // An optional additional salt value used to place this `Action` into a
+  // separate cache universe or "epoch" from other instances having the
```
bergsieker, line 430: 'cache universe or "epoch"' I'd probably go with the more generic "cache namespace" here.

build/bazel/remote/execution/v2/remote_execution.proto
bergsieker: I agree that in the case of a simple rotating integer salt using the digest provides no additional value. It's nice to have flexibility beyond just using an integer, though. A more advanced system might, for example, try to capture the version of the kernel and installed packages, add a numeric salt, and use the digest of all of that here. (This would obviously make the cache namespace specific to machines with identical configuration, which may or may not be desirable.)

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -425,6 +425,19 @@ message Action {
...
+  // same field contents. This salt typically comes from operational
+  // configuration specific to sources such as repo and service configuration,
+  // and allows disowning an entire set of ActionResults that might have been
+  // poisoned by buggy software or tool failures.
+  //
+  // As an example for repo configuration, a repo could contain a special file
```
bergsieker, line 436: I would generally shy away from such a specific example in the proto comments. In this particular I think describing the simple case of incrementing an integer makes sense, but I probably wouldn't connect it specifically with how to add that integer to the repository.

build/bazel/remote/execution/v2/remote_execution.proto
erikma: We actually hash the full file contents as some teams leave comments why they changed things. Some teams also like a major.minor string instead of a counting integer. We leave it up to them, as long as a change in the file is unique across the lifetime of the cache entries (weeks, typically).

build/bazel/remote/execution/v2/remote_execution.proto
erikma: Updated.

erikma: updated.

EdSchouten:
Thanks for working on this, Erik!

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -425,6 +425,14 @@ message Action {
   bool do_not_cache = 7;
 
   reserved 8; // Used for field moved to [Command][build.bazel.remote.execution.v2.Command].
+
+  // An optional additional salt value used to place this `Action` into a
+  // separate cache namespace from other instances having the same field
+  // contents. This salt typically comes from operational configuration
+  // specific to sources such as repo and service configuration,
+  // and allows disowning an entire set of ActionResults that might have been
+  // poisoned by buggy software or tool failures.
```
sstriker, line 434: Why wasn't the tool part of the Action input to begin with?  Wouldn't revving the tool then render a new Action?  Is this pouring salt into the wound of underspecified environments?

erikma: Action contains a ref to Command so the specific tool paths and environment cause different universes. However internally in our systems if you run something critical like a compiler a million times you sometimes see a zero process exit code but a tool generates no outputs or truncated/torn outputs, causing cache poisoning. Same tool that worked 999,999 times elsewhere but a build failure on missing outputs nonetheless.

sstriker: @erikma isn't ExecuteRequest.skip_cache_lookup meant to address that use case?  Or is this just a case of using the blunt instrument of declaring bankruptcy on one cache namespace as a whole and moving to a clean slate, because the surgical approach may be too expensive?

erikma: skip_cache_lookup from my POV is a feature flag allowing testing pure remoting without caching overhead. Or possibly for use when the tools involved are generating deliberately nondeterministic outputs, think of an FPGA layout tool or TensorFlow training job doing randomized stochastic gradient descent where you deliberately want to rerun the job without caching.

For the salt it's very much the latter. Imagine if you will 40,000 machines needing the surgical option on one ActionResult entry, or in some cases an entire tree of outputs derived from one bad entry. Better in almost all cases to use the Big Hammer and get to bed earlier.

PR #152, opened 2020-07-17
`GetActionResultRequest.inline_output_files` description: consider `output_paths`

santigl:
This adds a mention in the description for `GetActionResultRequest.inline_output_files` that entries in that field can also be files specified in  `Command.output_paths`.

### DIFF
--- a/build/bazel/remote/execution/v2/remote_execution.proto
+++ b/build/bazel/remote/execution/v2/remote_execution.proto
@@ -1311,7 +1311,8 @@ message GetActionResultRequest {
   bool inline_stderr = 4;
 
   // A hint to the server to inline the contents of the listed output files.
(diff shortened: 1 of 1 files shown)


build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1311,7 +1311,8 @@ message GetActionResultRequest {
   bool inline_stderr = 4;
 
   // A hint to the server to inline the contents of the listed output files.
-  // Each path needs to exactly match one path in `output_files` in the
+  // Each path needs to exactly match one path in `output_paths` or
```
mostynb, line 1314: nitpick: I guess it would be more precise to say that it needs to match one _file path_ in either of these fields?

PR #154, opened 2020-07-21
ExecutedActionMetadata: Allow adding auxiliary metadata

EdSchouten:
ExecutedActionMetadata provides a number of fields that help us to get insight in the worker that executed a build action, such as the name of the worker and some basic timing statistics.

Buildbarn ships with a small local modification to ExecutedActionMetadata, adding a 'repeated google.protobuf.Any' field. This field is used by some of Buildbarn's components to embed even more metrics into ActionResult, such as POSIX getrusage(2) information and temporary space allocation stats:

https://github.com/buildbarn/bb-remote-execution/blob/master/pkg/proto/resourceusage/resourceusage.proto

By having this data embedded in the ActionResult, people can view this information by simply inspecting a build action through bb-browser. This makes it easier for people to diagnose build failures accurately.

By using google.protobuf.Any, build clusters can expose their own metrics without requiring any further protocol extensions. It doesn't make sense to standardise those metrics, as they tend to be strongly implementation specific.

### DIFF
--- a/build/bazel/remote/execution/v2/remote_execution.proto
+++ b/build/bazel/remote/execution/v2/remote_execution.proto
@@ -19,6 +19,7 @@ package build.bazel.remote.execution.v2;
 import "build/bazel/semver/semver.proto";
 import "google/api/annotations.proto";
 import "google/longrunning/operations.proto";
+import "google/protobuf/any.proto";
 import "google/protobuf/duration.proto";
 import "google/protobuf/timestamp.proto";
 import "google/protobuf/wrappers.proto";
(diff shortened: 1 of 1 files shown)


bergsieker (commented):
General questions for followup from RE API monthly conversations. I'll loop back to comment more on these later.

* Extensibility is good and desirable in general. It's possible this should be paired with some level of standardization, e.g., for POSIX-related data.

* This message is fairly bloated and may not be very useful in the context of cache hits. We may want to consider moving it elsewhere or not returning it with cache hits.

* Can probably approve this as-is and loop back on these questions later.

edbaunton (approved):

sstriker (approved):
While I prefer standardized fields for common statistics, I tend to agree with @bergsieker to not hold this up at this stage.

EdSchouten:
Thanks for merging this @sstriker!

@bergsieker If you want I can send a followup PR to add POSIXResourceUsage to remote_execution.proto. Or do we want to add this to some kind of separate proto file as part of this repository?

bergsieker:
I'm inclined to leave this as a generic field rather than breaking out POSIX data explicitly. (My comment above was a summary of discussion from the RE monthly, not my personal opinion.) So far we've managed to keep the API itself mostly free of platform-specific structures. I would anticipate that this data is generally used for logging purposes, not for active evaluation by the client, in which case I think there's not much point in overly standardizing it.

mostynb:
This change is missing a dependency which broke the bindings generation, fix here: https://github.com/bazelbuild/remote-apis/pull/182

PR #155, opened 2020-07-25
Dynamic remote execution and two-phase caching

erikma:
For two-phase caching and dynamic action execution.


build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1185,6 +1417,156 @@ message ExecuteRequest {
...
+    // `DirectoryNode` are not required to be in the CAS. If the
+    // `DirectoryNode` digest is sent to the worker, the referenced `Directory`
+    // information MUST have been previously uploaded to the CAS.
+    Directory directory_metadata = 2;
+  }
+
+  // Sent in response to a dynamic file hash request.
+  message DynamicFileHashResponse {
+    // The file path, copied from the file hash request.
+    string path = 1;
```
EdSchouten, line 1519: Considering that resumption is clearly not a goal of this protocol, is there a need to pass pathname strings back and forth? Couldn't this just be an integer that corresponds with the count of the DynamicFileContentRequests received?

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1185,6 +1417,156 @@ message ExecuteRequest {
...
+  message DynamicFileContentResponse {
+    // The file path, copied from the file content request.
+    string path = 1;
+
+    // When true, the file is missing or inaccessible in the client filesystem
+    // and could not be uploaded to the CAS. The worker MAY fail the `Action`
+    // execution with a PRECONDITION_FAILED error, or it MAY return a
+    // file-not-found return code to the executing `Action` file read request
+    // which may cause the action to fail.
+    bool could_not_upload = 2;
```
EdSchouten, line 1542: Is there really a need to communicate this back? Couldn't the client just cancel the execution request?

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1029,6 +1217,50 @@ message ActionResult {
...
+  // [ContentAddressableStorage][build.bazel.remote.execution.v2.ContentAddressableStorage].
+  Digest console_outputs_digest = 14;
+}
+
+// A sequence of console outputs in the order received for a remoted process.
+// This may contain the interleaved outputs from a process in the order
+// received during execution. If the service does not implement
+// interleaving, this may contain a summary of the stdout and stderr output
+// streams as single large writes in a server-defined order.
+message ConsoleOutputs
```
EdSchouten, line 1243: Would it make sense to decompose this PR into two parts, and first focus on getting this added? It looks like this feature is somewhat independent of the dynamic remote execution.

(Nit: Place the { on the same line as 'message ConsoleOutputs'.)

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1029,6 +1217,50 @@ message ActionResult {
...
+  // An output record for one or more outputs to stdout or stderr.
+  //
+  // Design note: For text coloring and other effects, ANSI codes may be
+  // embedded within these strings. Windows console API calls are not
+  // supported in this protocol, but a server implementation MAY choose to
+  // translate console API calls into ANSI codes, as more recent Windows 10
+  // releases included ANSI code support.
+  message ConsoleOutput {
+    // When true, the `Writes` were directed at the stderr stream.
+    bool is_stderr = 1;
```
EdSchouten, line 1254:
```proto
enum Stream {
  STDOUT = 0;
  STDERR = 1;
}
```
?

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1288,6 +1670,228 @@ message WaitExecutionRequest {
...
+  // than this many items. Regardless of whether a page size is specified, the
+  // server may place its own limit on the number of items to be returned and
+  // require the client to retrieve more items using a subsequent request.
+  int32 page_size = 4;
+
+  // A page token, which must be a value received in a previous
+  // [GetPotentialActionResultsResponse][build.bazel.remote.execution.v2.GetPotentialActionResultsResponse].
+  // If present, the server will use that token as an offset, returning only
+  // that page and the ones that succeed it.
+  string page_token = 5;
```
EdSchouten, line 1724: What if an action has, say, thousands of potential action results? Is it still worth doing action cache lookups at that point?

More general question: is it still worth in practice to even cache actions that are known not to be hermetic? Doesn't the bandwidth/computation requirements at some point exceed the amount of work that's done by the action?

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1185,6 +1417,156 @@ message ExecuteRequest {
...
+
+// A streaming request message for
+// [Execution.ExecuteWithDynamicRetrieval][build.bazel.remote.execution.v2.Execution.ExecuteWithDynamicRetrieval].
+message StreamedDynamicExecutionData {
+  // Sent in response to a dynamic directory metadata request.
+  message DynamicDirectoryMetadataResponse {
+    // The directory path, copied from the directory metadata request.
+    string path = 1;
+
+    // Metadata for the directory. [DirectoryNode][build.bazel.remote.execution.v2.Execution.DirectoryNode]
```
alercah, line 1506: I think there's enough use of metadata in this API that we should avoid introducing another one.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1185,6 +1417,156 @@ message ExecuteRequest {
...
+    Directory directory_metadata = 2;
+  }
+
+  // Sent in response to a dynamic file hash request.
+  message DynamicFileHashResponse {
+    // The file path, copied from the file hash request.
+    string path = 1;
+    
+    // Set to true if the requested file path does not exist.
+    bool file_does_not_exist = 2;
```
alercah, line 1522: How can this happen? The worker shouldn't be able to request information about a file unless the `FileNode` is present, right?

build/bazel/remote/execution/v2/remote_execution.proto
alercah: I agree.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1185,6 +1417,156 @@ message ExecuteRequest {
...
+    // The file path for which the worker does not have the contents.
+    // Depending on the server and client implementation this may be a
+    // relative or fully qualified path.
+    string path = 1;
+
+    // The file hash of the content expected by the worker.
+    // If this is not the current hash of the file content at `path`
+    // then the corrected hash is returned in
+    // `DynamicExecuteOperationMetadata.dynamic_directory_metadata_response`.
+    string file_hash = 2;
```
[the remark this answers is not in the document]
alercah: I'm a bit confused about why there's a hash here at all. Is there any reason that a worker would know what hash it expects a file to have, but that file is missing from the CAS, other than that it just made a request for the hash?

It seems to me that when a request is made for a file's hash, the client should also upload the file's contents so that they are available for the worker to use.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -108,12 +109,126 @@ service Execution {
...
+  // needed to enumerate the complete contents of the directory.
+  //
+  // A typical implementation of a server using dynamic execution will use a
+  // virtual filesystem to allow receiving callbacks about files and
+  // directories requested via operating system filesytem APIs. The server
+  // MAY allow rendering of a fully qualified, rooted filesystem by the
+  // client, by accepting in `Action.input_root_digest` a `Directory` tree
+  // that contains a filesystem root appropriate to the operating system.
+  // By convention, representation of such a filesystem utilizes a Directory
+  // with an empty string for a name. Under Linux this translates into the
```
alercah, line 145: This has significant security implications as it could allow a malicious worker to exfiltrate arbitrary information, including private keys and the like, from the client's filesystem. There should be some discussion of this.

build/bazel/remote/execution/v2/remote_execution.proto
alercah: There probably should be a way for the client to signal that the client encountered an error reading the file or uploading it to the CAS. I would prefer to only have one option for behaviour here, rather than a choice. I also think I'd prefer it to be an execution failure, since that's what would happen if the client knew that the file was a needed input: it would have to abort upload.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -108,12 +109,126 @@ service Execution {
...
+  // error.
+  //
+  // This API is bidirectional both from the client to server and from server
+  // to client to allow the client to answer questions from the server about
+  // the contents of unspecified directories, file hashes, or to upload file
+  // content on demand, based on the observed filesystem calls of the action.
+  // This allows the server to execute actions whose inputs are incompletely
+  // specified. It also prevents execution of the action in a fire-and-forget
+  // fashion; the client must remain connected to the server for the entire
+  // execution of the action, or else the server MUST cancel the in-progress
```
alercah, line 161: Is there a strong reason to make this a MUST rather than a SHOULD? The server might see if the action completes without further need for inputs, perhaps.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1185,6 +1417,156 @@ message ExecuteRequest {
...
+    // `DynamicExecuteOperationMetadata.dynamic_directory_metadata_response`.
+    string file_hash = 2;
+  }
+
+  // The current stage of execution.
+  ExecutionStage.Value stage = 1;
+
+  // The digest of the [Action][build.bazel.remote.execution.v2.Action]
+  // being executed.
+  Digest action_digest = 2;
```
alercah, line 1456: If this remains a separate rpc from `Execute`, I think this should be called `partial_action_digest` to indicate it may be underspecified.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1288,6 +1670,228 @@ message WaitExecutionRequest {
...
+  // If empty, signifies that this is the last page of results.
+  string next_page_token = 2;
+}
+
+// The ordered file reads and probes and directory enumerations used in
+// two-phase caching to compare to the client filesystem when seeking a match
+// from an action fingerprint formed from potentially incomplete
+// (underspecified) inputs to a set of outputs.
+//
+// When comparing to the filesystem, a corresponding observed-input digest is
```
alercah, line 1750: I found these two paragraphs somewhat difficult to follow.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1288,6 +1670,228 @@ message WaitExecutionRequest {
...
+// `ObservedAccessFileNames` filters. Lexicographically sort the names by code
+// point or, equivalently, by the UTF-8 bytes. Join the sorted strings together
+// with the ',' character to form a comma-delimited list; note for a
+// multi-member list this means that there is no tail ',', e.g. for
+// a.h and b.h the list string is "a.h,b.h". Convert this string to uppercase.
+// Hash the UTF-8 string bytes.
+//
+// Discussion of how to generate pathsets from action execution filesystem
+// observations can be found at
+// https://github.com/microsoft/BuildXL/blob/master/Documentation/Wiki/Advanced-Features/Search-Path-Enumeration.md .
```
alercah, line 1772: I did not find this document to be much help at all; it assumes a lot of context that I and likely other users of this API do not have.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1288,6 +1670,228 @@ message WaitExecutionRequest {
...
+  repeated ObservedPathSetAndInputDigest potential_action_matches = 1;
+
+  // If present, signifies that there are more results which the client can
+  // retrieve by passing this as the page_token in a subsequent
+  // [request][build.bazel.remote.execution.v2.GetPotentialActionResults].
+  // If empty, signifies that this is the last page of results.
+  string next_page_token = 2;
+}
+
+// The ordered file reads and probes and directory enumerations used in
```
alercah, line 1745: More discussion of what is actually being encoded is needed. This ordered list of operations is representing what, exactly?

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1288,6 +1670,228 @@ message WaitExecutionRequest {
...
+      // `Path` refers to a file that was read by the action. If the file is
+      // not present, this pathset does not match the filesystem. Otherwise
+      // the hash of the string constant "R" followed by the hash of the file
+      // must be added to the observed-input digest.
+      FILE_READ = 0;
+
+      // `Path` is a directory that was enumerated by the action using
+      // `DirectorySearchPattern`. If the directory is missing, this pathset
+      // does not match the filesystem. Otherwise the hash of the string
+      // constant "E" followed by the hash of the directory contents with the
```
alercah, line 1791: How does one hash the directory contents? Is it a `Directory` message?

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1288,6 +1670,228 @@ message WaitExecutionRequest {
...
+  // An individual file or directory access observed during execution of
+  // an [Action][build.bazel.remote.execution.v2.Action].
+  message Entry
+  {
+    // Operation kind.
+    enum OpKind
+    {
+      // `Path` refers to a file that was read by the action. If the file is
+      // not present, this pathset does not match the filesystem. Otherwise
+      // the hash of the string constant "R" followed by the hash of the file
```
alercah, line 1784: What is the hash of the file? Content hash or `FileNode` hash?

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1288,6 +1670,228 @@ message WaitExecutionRequest {
...
+}
+
+// The ordered file reads and probes and directory enumerations used in
+// two-phase caching to compare to the client filesystem when seeking a match
+// from an action fingerprint formed from potentially incomplete
+// (underspecified) inputs to a set of outputs.
+//
+// When comparing to the filesystem, a corresponding observed-input digest is
+// initialized. Each pathset entry is examined against the filesystem, and
+// depending on the outcome of the comparison, new hash information is added
```
alercah, line 1752: How is information "added" to a digest?

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1288,6 +1670,228 @@ message WaitExecutionRequest {
...
+    enum OpKind
+    {
+      // `Path` refers to a file that was read by the action. If the file is
+      // not present, this pathset does not match the filesystem. Otherwise
+      // the hash of the string constant "R" followed by the hash of the file
+      // must be added to the observed-input digest.
+      FILE_READ = 0;
+
+      // `Path` is a directory that was enumerated by the action using
+      // `DirectorySearchPattern`. If the directory is missing, this pathset
```
alercah, line 1789: Why does DIRECTORY_ENUMERATION use DirectorySearchPattern, but DIRECTORY_SEARCH not?

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1288,6 +1670,228 @@ message WaitExecutionRequest {
...
+      // added to the observed-input digest.
+      EXISTING_DIRECTORY_PROBE = 4;
+
+      // `Path` is a file or directory probed for existence by the action
+      // that was not present during execution. If path is present as a file
+      // or directory, this pathset does not match the filesystem. Otherwise
+      // the hash of the string constant "P" must be added to the
+      // observed-input digest.
+      ABSENT_FILE_OR_DIRECTORY_PROBE = 5;
+    }
```
alercah, line 1829: This approach to specifying file operations seems underspecified. There are a number of properties, like the contents of `FileProperties` and `is_executable`, which could be distinguished under this API but which aren't encoded here. On unix, probably the correct thing to do is encode that a `stat()` call was made and list all of the queried properties.

I'm unsure about what should be done about properties that could be in `FileProperty` and would fall outside that, which I imagine could include Linux extended ACLs. I guess this should also accept extension operations, then.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1288,6 +1670,228 @@ message WaitExecutionRequest {
...
+  // extensions (e.g. .h, .hpp, no extension in C++; .dll, .exe in .NET
+  // assembly search).
+  repeated string ObservedAccessedFileNames = 2;
+}
+
+// Represents a two-phase cache pathset, referenced or inlined, with an
+// observed-input digest and a reference to build outputs to retrieve from the
+// cache if the observed-input digest calculated from the client's filesystem
+// matches.
+message ObservedPathSetAndInputDigest {
```
alercah, line 1865: A shorter name, more descriptive of its use, would most likely be better.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1288,6 +1670,228 @@ message WaitExecutionRequest {
...
+  string next_page_token = 2;
+}
+
+// The ordered file reads and probes and directory enumerations used in
+// two-phase caching to compare to the client filesystem when seeking a match
+// from an action fingerprint formed from potentially incomplete
+// (underspecified) inputs to a set of outputs.
+//
+// When comparing to the filesystem, a corresponding observed-input digest is
+// initialized. Each pathset entry is examined against the filesystem, and
```
alercah, line 1751: initialized how?

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1598,6 +2268,10 @@ message CacheCapabilities {
 
   // Whether absolute symlink targets are supported.
   SymlinkAbsolutePathStrategy.Value symlink_absolute_path_strategy = 5;
+
+  // Support for two-phase caching for actions with possibly under-specified
+  // inputs.
+  TwoPhaseCacheSupport.Value two_phase_cache_support = 6;
```
alercah, line 2274: I am not sure a boolean would not be a better choice, ditto with the others.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1288,6 +1670,228 @@ message WaitExecutionRequest {
...
+// Returns a set of zero or more
+// (pathset, observed-input digest, `ActionResult` digest) results
+// from a two-phase cache lookup call to
+// [ActionCache.GetPotentialActionResults][build.bazel.remote.execution.v2.ActionCache.GetPotentialActionResults].
+// These results are compared to the local filesystem and, on a match to the
+// observed-input digest, the associated `ActionResult` is used to retrieve
+// cached outputs.
+message GetPotentialActionResultsResponse {
+  // A set of possible matches to the client filesystem along with a reference to build outputs.
+  repeated ObservedPathSetAndInputDigest potential_action_matches = 1;
```
EricBurnett, line 1736: Require sorting by pathset (under some scheme), so a pathset with multiple candidates can be checked at once? Or use a scheme like (repeated pathset -> (repeated digest->digest pairs)), since it's likely one pathset will have multiple candidates?

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1288,6 +1670,228 @@ message WaitExecutionRequest {
...
+// depending on the outcome of the comparison, new hash information is added
+// to the observed-input digest. At the end of pathset enumeration the
+// observed-input digest is compared to the observed-input digest provided in
+// the two-phase cache lookup (`observed_input_digest` in
+// [ObservedPathSetAndInputDigest][build.bazel.remote.execution.v2.ObservedPathSetAndInputDigest])
+// and if they match, this is a cache hit, and the resulting
+// [ActionResult][build.bazel.remote.execution.v2.ActionResult] should be
+// retrieved and used.
+//
+// Generating a directory hash: Create a list of the names of directory members
```
EricBurnett, line 1761: It seems a little odd to me to have a new scheme for generating hashes here. We do so elsewhere in the API by building a merkle tree, and you've already detailed above the generalization required to make a partial merkle tree with unspecified nodes and whatnot. Couldn't we instead say something like "update the partial merkle tree with the hashes/contents of the given files/directories", and then re-take the merkle tree hash as the new digest to match against? That way there's no new scheme to learn/implement, only effectively updating the client-predicted paths with an oracle's suggested paths.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1288,6 +1670,228 @@ message WaitExecutionRequest {
...
+// (files and directories) that pass the directory enumeration or
+// `ObservedAccessFileNames` filters. Lexicographically sort the names by code
+// point or, equivalently, by the UTF-8 bytes. Join the sorted strings together
+// with the ',' character to form a comma-delimited list; note for a
+// multi-member list this means that there is no tail ',', e.g. for
+// a.h and b.h the list string is "a.h,b.h". Convert this string to uppercase.
+// Hash the UTF-8 string bytes.
+//
+// Discussion of how to generate pathsets from action execution filesystem
```
EricBurnett, line 1770: It can be reasonable to link out to extra content, but this API must stand alone without it (e.g. when that link eventually goes stale). Please make sure a reader without access to this wiki page will have all necessary information to figure out the API here.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1029,6 +1217,50 @@ message ActionResult {
...
+  // translate console API calls into ANSI codes, as more recent Windows 10
+  // releases included ANSI code support.
+  message ConsoleOutput {
+    // When true, the `Writes` were directed at the stderr stream.
+    bool is_stderr = 1;
+    
+    // One or more written strings. There is an implicit operating system
+    // dependent newline after each string. Each string may contain embedded
+    // OS-dependent newline character sequences.
+    repeated string writes = 2;
```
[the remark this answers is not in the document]
erikma: The intent was for batching: One stdout|err boolean that leads to one or more lines of output. The most common case is for there to be a lot of stdout output and no stderr, which would result in a ConsoleOut->isStdErr=false + array-of-lines.

I was avoiding requiring concatenation in the implementation. AnyBuild's implementation does not concatenate (but could). If you feel strongly about requiring batching it does simplify the proto at the cost of the implementation having to concatenate.

build/bazel/remote/execution/v2/remote_execution.proto
erikma: Broke out into a new PR: https://github.com/bazelbuild/remote-apis/pull/173

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1029,6 +1217,50 @@ message ActionResult {
...
+
+  // The inlined contents of a
+  // [ConsoleOutputs][build.bazel.remote.execution.v2.ConsoleOutputs] message
+  // containing a series of stdout and stderr outputs in order of write from
+  // the action. The server SHOULD NOT inline `console_outputs` unless
+  // requested by the client in the
+  // [GetActionResultRequest][build.bazel.remote.execution.v2.GetActionResultRequest]
+  // message. The server MAY omit inlining, even if requested, and MUST do so if inlining
+  // would cause the response to exceed message size limits.
+  bytes console_outputs = 13;
```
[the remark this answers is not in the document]
erikma: Changed in #173 and pushed similar changes here.

build/bazel/remote/execution/v2/remote_execution.proto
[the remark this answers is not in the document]
erikma: (note any further discussion here needs to be reflected or copied into #173)

build/bazel/remote/execution/v2/remote_execution.proto
erikma: Changed to bool for ConsoleOut as part of #173 (and copied back here). Changed these to bool as well.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1185,6 +1417,156 @@ message ExecuteRequest {
...
+  // Sent in response to a dynamic file hash request.
+  message DynamicFileHashResponse {
+    // The file path, copied from the file hash request.
+    string path = 1;
+    
+    // Set to true if the requested file path does not exist.
+    bool file_does_not_exist = 2;
+
+    // The file digest and associated metadata.
+    FileNode file = 3;
```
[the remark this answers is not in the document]
erikma: We also need is_executable (if it survives beyond 2.x) and node_properties to allow setting file mode/permissions/readonly/etc.

build/bazel/remote/execution/v2/remote_execution.proto
erikma: Updated, and reflected into #173.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1185,6 +1417,156 @@ message ExecuteRequest {
...
+// which will be contained in the [metadata field][google.longrunning.Operation.response] of the
+// [Operation][google.longrunning.Operation].
+//
+// Design note: Dynamic directory and file requests are made using individual
+// messages for each directory or file path. The typical implementation
+// utilizes a virtual filesystem and a directory or file read request is
+// held pending in the VFS awaiting the information from the client.
+// VFS rendering callbacks may occur in a multithreaded fashion, but this is
+// unusual, and a chatty back-and-forth request series to the client resolves
+// pending I/O faster for each thread rather than adding in a Nagle
```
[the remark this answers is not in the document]
erikma: Agreed. Updated.

build/bazel/remote/execution/v2/remote_execution.proto
erikma: In actual practice the client's filesystem is sometimes changed underneath execution, particularly in a highly parallel build using build engines (like MSBuild) whose build DAG is not strongly specified.

build/bazel/remote/execution/v2/remote_execution.proto
erikma: Translated from a different field from the internal AnyBuild protocol. Agreed it does not belong, removed.

build/bazel/remote/execution/v2/remote_execution.proto
erikma: DIR_SEARCH is not a regular enumeration, it's a combination of one or more specific filenames (i.e. no wildcards involved). This is typical for C++ compilers, which (in most implementations) call the filesystem enumeration APIs with exactly the filename of the header/include desired. For example, if a .cpp contains:

#include "foo.h"
#include "bar.h"

And the search paths include "dir1;dir2" the filesystem will typically see the following sequence of directory enumerations:

dir1: enum "foo.h" (it's not here)
dir2: enum "foo.h" (it's here)
dir1: enum "bar.h" (it's here so file probing stops)

In the ObservedPathSet this becomes:
Entries:
    { path="dir1", kind=DIR_SEARCH }
    { path="dir2", kind=DIR_SEARCH }
observed_accessed_file_names = [ "foo", "bar" ]

It's important in these cases to ensure that the presence or absence of foo(.h) and bar(.h) in both dir1 and dir2 are added to the observed input digest as that affects correctness of the cache hits in the face of such directory searching/probing.

alercah: Is this a Windows-specific API? Why wouldn't the accessed file names be "foo" and "bar"? Is some equivalent of `stat` not used to query if a directory contains a file with a specific name?

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -108,12 +109,126 @@ service Execution {
...
+  // guess for the `Action`'s needed directory and file layout into the
+  // [ContentAddressableStorage][build.bazel.remote.execution.v2.ContentAddressableStorage].
+  // It then calls `ExecuteWithDynamicRetrieval` with an `action_digest`
+  // referring to them. The worker will run the action and eventually return
+  // the result.
+  //
+  // The `Action.input_root_digest` directory tree provides a filesystem
+  // layout that may be underspecified in terms of missing directories or
+  // unspecified file hashes. Any `Directory` entry provided in this tree
+  // must contain at least the subdirectory names and file names and sizes
```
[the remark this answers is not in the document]
erikma: Sizes are required to feed the VFS view of a directory from the POV of dir enumeration APIs. At the moment of an enumeration the { filename, size, create/update datetimes } for each file are passed to the virtual filesystem, and that view cannot then be changed. -1 is not acceptable, and tools care about file sizes for decision-making, so the correct view must be presented from the beginning.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -108,12 +109,126 @@ service Execution {
...
+  // client, by accepting in `Action.input_root_digest` a `Directory` tree
+  // that contains a filesystem root appropriate to the operating system.
+  // By convention, representation of such a filesystem utilizes a Directory
+  // with an empty string for a name. Under Linux this translates into the
+  // '/' root. Under Windows such a root contains `DirectoryNode`s that are
+  // the single-character drive letters available on the client, allowing
+  // the server to present a virtual filesystem spanning all drive roots
+  // for dynamic request by the server. There must be exactly one such
+  // empty-string root at the root of the directory tree; any use of empty
+  // strings other than the root SHOULD result in an `INVALID_ARGUMENT`
```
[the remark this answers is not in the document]
erikma: MUST is fine as a required semantic. Updated.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -108,12 +109,126 @@ service Execution {
...
+  //
+  // Errors discovered during creation of the `Operation` will be reported
+  // as gRPC Status errors, while errors that occurred while running the
+  // action will be reported in the `status` field of the `ExecuteResponse`. The
+  // server MUST NOT set the `error` field of the `Operation` proto.
+  // The possible errors include:
+  //
+  // * `INVALID_ARGUMENT`: One or more arguments are invalid.
+  // * `FAILED_PRECONDITION`: One or more errors occurred in setting up the
+  //   action requested, such as a missing input or command or no worker being
```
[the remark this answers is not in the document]
erikma: Good catch. Updated.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1613,6 +2287,12 @@ message ExecutionCapabilities {
 
   // Supported node properties.
   repeated string supported_node_properties = 4;
+
+  // Support for use of interleaved console output formats.
```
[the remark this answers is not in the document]
erikma: A client might decide not to support a server implementation that does not have this capability. Other than that, you're right, there's not much else to do besides asking for consoleouts and using the stdout/err streams as a fallback.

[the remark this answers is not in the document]
erikma: LMK if you think this is worth simply cutting out (or comment in #173 since I moved the consoleout stuff there).

build/bazel/remote/execution/v2/remote_execution.proto
erikma: stat() equivalence on Windows is inexact and apps do it differently. If you have an open file handle there's GetFileInformationByHandle() but the most common case when querying directory metadata is to start a dir enumeration with FindFirstFile() and following FindNextFile(), but using an exact name string like "foo.h" instead of using wildcard characters.

On Linux the sandbox would see a stat() call for a specific name but the resulting ObservedPathSet result is the same: A file search for a specific name, and tracking whether the file was present or not in the directory, to feed the summary hash to use for comparison with what the worker saw from its sandboxed API calls.

build/bazel/remote/execution/v2/remote_execution.proto
erikma: Speaking from experience direct and indirect - when you end up with thousands someone is doing something wrong or interesting. We have had such live-site issues and the fixes were varied, either by focusing the action_fingerprint_digest key by adding more specific input hashes to get fewer results, or letting the client decide when it is done reviewing results e.g. as a time-boxed effort. I was not aiming with the doc comments here to force any specific implementation since the user intent can vary (maybe like one case they actually wanted to spend the time on the hundreds of potential results).

build/bazel/remote/execution/v2/remote_execution.proto
erikma: I prefer paths for the same reasons as Eric - in practice the number of dynamic requests is low so wasting some wire space and comparison time is not worth fixing. If you disagree please reopen this comment.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -855,6 +1039,10 @@ message ExecutedActionMetadata {
 
   // When the worker finished uploading action outputs.
   google.protobuf.Timestamp output_upload_completed_timestamp = 10;
+
+  // Additional statistics for dynamic execution via
+  // [ExecuteWithDynamicRetrieval][build.bazel.remote.execution.v2.Execution.ExecuteWithDynamicRetrieval].
+  DynamicExecutionMetadata dynamic_execution = 11;
```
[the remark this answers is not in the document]
erikma: What's here were the most-likely-interesting stats cut down from ~40 we send back right now. In our iteration toward this API I was going to use the DynamicExecutionMetadata.extended_metadata Any field for the remainder. I can move all of this over to the #154 Any field, but then there's no common schema, just a bag of key-value pairs with a convention for the key names that might differ between vendors.

I'll keep this open and merge with 154 if/when it goes in.

build/bazel/remote/execution/v2/remote_execution.proto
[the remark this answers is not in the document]
erikma: Changed this to a Digest, and updated doc comments for the response with Eric's commentary.

build/bazel/remote/execution/v2/remote_execution.proto
erikma: @alercah I was avoiding adding such an example into the doc comment text but I don't see a great way to add in more information. I did add a link to the BuildXL discussion markdown for reference; do you think that URL would be usefully repeated on the per-field comments as well to aid understanding?

build/bazel/remote/execution/v2/remote_execution.proto
erikma: Quite. Changed to directory content from directory metadata in message and field names.

build/bazel/remote/execution/v2/remote_execution.proto
erikma: No strong reason. Loosened the wording here to discuss the possibility of letting the server decide based on its need for dynamic information.

erikma: Kept the indication that WaitExecution is not allowed; LMK if you disagree with it in the face of the new wording.

build/bazel/remote/execution/v2/remote_execution.proto
erikma: Ah I see from the comment above that you found that doc impenetrable. Let me create a better one or modify the existing and see if that works better.

build/bazel/remote/execution/v2/remote_execution.proto
erikma: Discussing with the sandboxing devs. My first response would be that file presence and hash are sufficient, but maybe we've missed a nuance here.

build/bazel/remote/execution/v2/remote_execution.proto
erikma: We had trouble finding a good name... but PotentialActionMatch seems a lot better.

build/bazel/remote/execution/v2/remote_execution.proto
erikma: Alexis had a comment about how the algo works (above), and also indicated this content is not understandable without a lot more context. Do you think adding in examples to the doc comments works?

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1029,6 +1217,50 @@ message ActionResult {
...
+  // containing a series of stdout and stderr outputs in order of write from
+  // the action. The server SHOULD NOT inline `console_outputs` unless
+  // requested by the client in the
+  // [GetActionResultRequest][build.bazel.remote.execution.v2.GetActionResultRequest]
+  // message. The server MAY omit inlining, even if requested, and MUST do so if inlining
+  // would cause the response to exceed message size limits.
+  bytes console_outputs = 13;
+
+  // The digest for a blob containing the
+  // [ConsoleOutputs][build.bazel.remote.execution.v2.ConsoleOutputs] for the action,
```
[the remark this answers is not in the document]
erikma: Validation: In each such usage I pair an inline with a digest as is done elsewhere, for the case of the server deciding not to inline. The next field after this is the ConsoleOut digest. Or am I missing something?

build/bazel/remote/execution/v2/remote_execution.proto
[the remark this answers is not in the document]
erikma: Removed.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1185,6 +1417,156 @@ message ExecuteRequest {
...
+  // A series of incremental (streamed) stdout and stderr outputs in order of
+  // write from the action during execution. The full set of console outputs
+  // from the action will also be written in the final
+  // [ExecuteResponse][build.bazel.remote.execution.v2.ExecuteResponse].
+  // The server MAY buffer these outputs in order to return larger batches
+  // of messages less frequently. The server MAY choose to omit incremental
+  // messages and instead wait to the final `ExecuteResponse` to return all
+  // outputs from the action, however in this case the server MUST NOT return
+  // any incremental messages in this field.
+  ConsoleOutputs streamed_console_outputs = 3;
```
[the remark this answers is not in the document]
erikma: My reasons for ConsoleOut inlining in the bidi protocol are different: Often we get some mega C++ compile steps (one remoted action running 500 compiler sub-processes each compiling 1 cpp file) where the incremental stdout (and occasional interleaved stderr) helps with interactive console feedback at the client from the build engine's parsing of incremental outputs. Same goes for remoting long-running things like TensorFlow training runs where incremental feedback instead of batchy final results helps with knowing the system is working.

The only alternative I could think of to get a similar user experience if mapping to streams would be to place time codes on segments of the stream so that interleaving could be restored on the client side for fidelity. That seemed nastier for no particular gain.

Any other ideas? The UX aspect of it is actually kind of important in this case, but willing to discuss alternatives. It's one of the reasons I state in the oversized doc comments here that the server might not return these, and that the final outputs (including from cached results) are replicated in the final messages. (We replay the ordered stderr+out strings to the waiting app on a cache hit too, it jjust looks like the 500 cpp files compiled very very rapidly.)

erikma:
Closing this PR and related issue: Though interesting for some community members, this proposal is unlikely to be accepted into the PE protocol (even for v3) as it needs a second implementation.

Issue #135, opened 2020-04-20
V3 idea: Let Digest.hash use 'bytes' instead of 'string'


EdSchouten:
Right now the Digest message uses a string to store the checksum of the object. This is wasteful, because for the hashing algorithms that we use, the output is always a sequence of bytes. Using a bytes field cuts down the storage space by a half (for that field; not necessarily for the full object).

EricBurnett:
Hi Ed! Thanks for filing this.

We've discussed this a couple times within Google in 2017/2018, so copying out the conclusions we came to from one of those evaluations as a starting point. I think we may also have discussed it in the community in the old BuildFarm group as well, but I couldn't find a handy reference for that discussion. Regardless, what we came up with then was:

**Arguments in favor of representing hashes as bytes when sent over the network:**

   * Reduces the bytes sent over the network: Even when using gzip, a hexadecimal representation increases network bandwidth requirements by around 30% (after eliminating outputs) for both cache hits and misses.
   * For a 4000 action build, all cache hits: 4000 * 3 (digests) * 32B * 30% ~= 115kB
   * For a 4000 action build, no cache hits: 4000 * ~100 digests * 32B * 30% ~= 3.8MB
   * Average is >90% cache hits -> 487 kB
   * Ballpark $60/year in bandwidth for TensorFlow (assuming GCP prices, 10k builds/week at ~10000 actions) [Date of numbers unknown; probably early 2018]
   * There’s no need to convert between bytes <-> hex, except during debugging.

**Arguments against:**

   * Hex must still be specified as a format for blobs in the API.
      * (Unless ByteStream is eliminated) for specifying digests in URLs
      * All clients and servers should log values consistently, to allow easy out-of-band communication during debugging.
      * Spreads logic for dealing with digest format through code, vs encapsulated 
   * Current debugging tools do not print bytes in an easily readable format.
   * It's an API change vs today, and so default is "no change" unless proven to be sufficiently worthwhile.

At the time, the deciding factor seemed to come down to whether the difference in bandwidth was worth the effort of changing and the increase in complexity in dealing with two formats. (Noting that for backend storage you can use whatever you want, so it's only the API that strictly must be defined one way or another). This could be argued on the grounds that either that the cost of the bandwidth is material, or that builds are measurably slowed (wallclock) because of the increase in time required to move the bytes around.

[...]

Issue #134, opened 2020-04-20
V3 idea: No longer allow Digest.size_bytes <= 0


EdSchouten:
Right now it is allowed to create Digest messages that have `size_bytes == 0`, referring to the empty blob. In #131 we're extending the protocol to require that the empty blob is always present, because it can be derived trivially. I personally find this a bit problematic:
- It makes the protocol less regular and consistent.
- Naïvely implemented client/servers will get this wrong. For example, what is FindMissingBlobs() on `{hash: "e984d2bdd07318c4e29f7a2ceea4a9e4569e2d8e695a953a4e2df6f69fbdec95", size_bytes: 0}` supposed to do? Report existence, because it has size zero? Or should it report absence, because the empty blob actually has SHA-256 sum `e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855`?
- When digests for empty blobs are embedded in other messages, they still waste space. We still end up storing a SHA-256 sum.

I would like to suggest that we simply deny the existence of Digest messages with `size_bytes <= 0`. Any field where the empty blob needs to be referenced, we should use null. This means that the optimization that Bazel performs of not loading the empty blob becomes the norm, as there is no longer any way to even address the empty blob.

juergbi:
>     * For example, what is FindMissingBlobs() on `{hash: "e984d2bdd07318c4e29f7a2ceea4a9e4569e2d8e695a953a4e2df6f69fbdec95", size_bytes: 0}` supposed to do? Report existence, because it has size zero? Or should it report absence, because the empty blob actually has SHA-256 sum `e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855`?

As it's not the correct empty blob, it should report absence.

>     * It makes the protocol less regular and consistent.

I'm not sure either optimization variant is really better with regards to regularity/consistency. Also with your suggested approach, special cases have to be defined. E.g., what happens if a client attempts to write a blob with a Digest with `size_bytes == 0`? I'd guess something like `INVALID_ARGUMENT`.

However, as this approach is overall more efficient and the implementation complexity doesn't seem higher to an optimized v2 implementation, I think it's a sensible plan.

There may be code in some clients and/or servers that currently associates a different meaning to null Digests in some contexts (e.g., Digest not yet calculated). However, I don't think this should be a reason against this idea. Assuming there is no ambiguity anywhere in the protocol itself.

EdSchouten:
[...]

PR #159, opened 2020-07-29
Switching Tree definition to use proto map

werkt:


werkt:
Discussion per https://groups.google.com/g/remote-execution-apis/c/FnkmSFOH42k

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1065,14 +1065,15 @@ message OutputFile {
...
 // single directory Merkle tree, compressed into one message.
 message Tree {
-  // The root directory in the tree.
-  Directory root = 1;
+  reserved 1, 2; // Used for removed fields in an earlier version of the API.
+
+  // The digest of the root directory in the tree.
+  Digest rootDigest = 3;
 
   // All the child directories: the directories referred to by the root and,
```
juergbi, line 1073: Would the new map only contain the child directories as per the comment or also the root `Directory`? The former seems odd to me as it would require an extra round trip for the root directory but not for its child directories.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1065,14 +1065,15 @@ message OutputFile {
...
-  // recursively, all its children. In order to reconstruct the directory tree,
-  // the client must take the digests of each of the child directories and then
-  // build up a tree starting from the `root`.
-  repeated Directory children = 2;
+  // recursively, all its children. The keys of the map are the hash of the digest
+  // of each directory.
+  map<string, Directory> directories = 4;
```
juergbi, line 1076: Using the hash string without a complete `Digest` object would be unusual for the REAPI. The map approach may also be incompatible with #134 and #135.

Have you considered a repeated embedded Digest+Directory message? I wouldn't expect the wire overhead to be significant. However, maybe the generated protobuf code would be less efficient and/or convenient. Or do you see any other downsides?

build/bazel/remote/execution/v2/remote_execution.proto
werkt: It contains the root directory, which is alluded to (and was confusing) in the children comment previously, that it should contain everything under a root.

build/bazel/remote/execution/v2/remote_execution.proto
werkt: Don't think its incompatible with either of those - we're constrained here by the limitations imposed by proto on maps: they would be Digests if I had my druthers. The keys are the keys, as rendered using hex, whether bytes or strings are used to retain them. Using the digest hash only provides a lookup that seems less awkward than the arbitrary hash/size rendering that we use everywhere else, but I'd be on board for that if necessary. The size of the referring digest can be checked easily with the size of the referent directory as a blob independently.

One other downside is repetition - while trees can't contain loops (and it would be difficult to manufacture one anyway), they can have repeated subtrees, which occurs more often than I'd like to think about. It's quite a departure to use a hierarchical message store in any case - I prefer the digest indexing.

peterebden, line 1076: Have we thought about what this means for calculating digests? The ordering of a map in the wire format is undefined so I think the standard serialise + hash approach will be nondeterministic here.

werkt: I'm pretty sure that doesn't matter - there is nothing using the digest of a Tree as a key explicitly (a Tree is identified solely by the root Directory digest), so the byte representation of the Tree blob should not affect anything from anyone's perspective except internal mechanisms for generating/storing them. Users of the Tree message won't be able to make any assumption as far as byte representations, comparisons between message types should prove equivalent, and blob comparisons are not useful or defined for non-CAS content. Larger transports of directory collections already happen via streaming.

peterebden: The ActionResult does use the digest of the whole Tree message:
```
  // corresponding directory existed after the action completed, a single entry
  // will be present in the output list, which will contain the digest of a
  // [Tree][build.bazel.remote.execution.v2.Tree] message containing the
  // directory tree, and the path equal exactly to the corresponding Action
  // output_directories member.
```
https://github.com/bazelbuild/remote-apis/blob/006a399149dd2dad4eedff247578e94bd957da9c/build/bazel/remote/execution/v2/remote_execution.proto#L922

werkt: Truth, well played.

The provenance of a Tree in that case might be worth discussing, given that there are only two entities that will likely construct them, and they will not be subject to re-serialization after creation: every one of their uses will be complete.

And one added bonus (for me at least) is that the Java serialization of protobuf maps, while undefined by language, respects a `deterministicSerialization` property of the target output stream. If two maps `.equals()`, then they serialize to the same bytes with this flag enabled. I'll be throwing it on in buildfarm, and recommending so for bazel as well.

peterebden: Agreed. I wonder if there is a possibility of referring to the digest of the directory as you suggested, and this message retaining the map as effectively an optimisation to save future GetTree calls. I'd consider it nicer to have more symmetry between the output and input facets.

Interesting on the Java serialisation; AFAICT there is no similar option for Go (which is most relevant to me, unsure on other languages right now).

sstriker: Looking at maps in https://developers.google.com/protocol-buffers/docs/proto3#maps it's not ideal.  Quoting:

> * Wire format ordering and map iteration ordering of map values is undefined, so you cannot rely on your map items being in a particular order.
> * When generating text format for a .proto, maps are sorted by key. Numeric keys are sorted numerically.
> * When parsing from the wire or when merging, if there are duplicate map keys the last key seen is used. When parsing a map from text format, parsing may fail if there are duplicate keys.
> * If you provide a key but no value for a map field, the behavior when the field is serialized is language-dependent. In C++, Java, and Python the default value for the type is serialized, while in other languages nothing is serialized.

Then there is the backward compatibility section (https://developers.google.com/protocol-buffers/docs/proto3#backwards_compatibility), which leads me to believe that @juergbi 's suggestion may be the viable path forward.  That is:

```
message TreeEntry {
  Digest digest = 1;
  Directory directory = 2;
}

repeated TreeEntry directories = 4;
````

I would suggest to preceed this with the fact that the tree entries should be ordered by digest.  And when parsing that in case of duplicate values, the last encountered wins (rationale: stay aligned with map)

File: build/bazel/remote/execution/v2/remote_execution.proto [hunk 3]
```diff
@@ -1065,14 +1065,15 @@ message OutputFile {
 // [Directory][build.bazel.remote.execution.v2.Directory] protos in a
 // single directory Merkle tree, compressed into one message.
 message Tree {
-  // The root directory in the tree.
-  Directory root = 1;
+  reserved 1, 2; // Used for removed fields in an earlier version of the API.
```
sstriker, line 1068: I assume we need to keep this field in for backward compatibility with existing implementations.

sstriker:
@werkt, how would you like to proceed?

build/bazel/remote/execution/v2/remote_execution.proto
werkt: I'm not 100% sure on how these sorts of migrations are supposed to happen, but at some point you have to drop the previous definitions - if we come to a semver/requested version interaction conclusion here, then I suppose so, but otherwise   I believe the fields should just be dropped.

build/bazel/remote/execution/v2/remote_execution.proto
werkt: The only way to have duplicate entries is to either a) be non-CAS or have different digest functions, or b) literally have replicated directory copies with the same byte representation, so it should be immaterial whether one 'wins.'

werkt:
The repeated digest works for me, I'm a little sad that I will need to create a tree utils mechanism to validate and construct the map in-language, but if that gets an accept, it's worth it.

build/bazel/remote/execution/v2/remote_execution.proto
sstriker: You're right, stating that the directories MUST be sorted by digest should be sufficient.

EdSchouten:
As I mentioned to @werkt in private, the downside of this approach is that it introduces redundancy to Tree. The digest of a Directory is stored twice:
- Once as the map key,
- Once in a way that can be derived by hashing the individual Directory object.

The downside of that approach is that it may easily cause security issues if insufficient checking is done. For example, what if a piece of code takes all of the entries in a Tree, stores them in the CAS and trusts the key to be valid? That could cause you to create CAS entries that don't match up with the digest.

An alternative way of solving this that doesn't introduce this ambiguity is to remove digests from DirectoryNode entirely. Instead of referencing child directories by digest, we could replace it by a simple integer. In the case of Tree objects, the indices refer to sibling Directories. By requiring that directory entries are stored in topological order, it's easy to guarantee that the Tree remains acyclic.

Here is a prototype change that I wrote that at least makes this concept clear. It is by no means intended to be merged directly.

https://github.com/EdSchouten/remote-apis/commit/dd6864f5af573c7e4b1f1e96185273db99142877

bergsieker:
George, can you convert this in to an issue for v3?

PR #163, opened 2020-08-13
Add a clarification comment about the empty std*_digest

vors:
Bazel source code suggests that there should not be any difference in the client behavior between omitted `stdout_digest` and digest that points to an empty blob.

https://github.com/bazelbuild/bazel/blob/39bc97eab295bddb35b38bfc4a2ff3d2b15d034e/src/main/java/com/google/devtools/build/lib/remote/RemoteCache.java#L490

Same for the `stderr_digest`.
This change tries to codify this into the API.

### DIFF
--- a/build/bazel/remote/execution/v2/remote_execution.proto
+++ b/build/bazel/remote/execution/v2/remote_execution.proto
@@ -1021,6 +1021,7 @@ message ActionResult {
   // The digest for a blob containing the standard output of the action, which
   // can be retrieved from the
   // [ContentAddressableStorage][build.bazel.remote.execution.v2.ContentAddressableStorage].
+  // When ommited, MUST be treated as the digest of the empty blob.
   Digest stdout_digest = 6;
 
   // The standard error buffer of the action. The server SHOULD NOT inline
@@ -1033,6 +1034,7 @@ message ActionResult {
   // The digest for a blob containing the standard error of the action, which
   // can be retrieved from the
   // [ContentAddressableStorage][build.bazel.remote.execution.v2.ContentAddressableStorage].
+  // When ommited, MUST be treated as the digest of the empty blob. 
   Digest stderr_digest = 8;
 
   // The details of the execution that originally produced this result.


sstriker:
Thanks for your suggestion!

I'm not sure this need clarification.  When the stdout_digest or stderr_digest are omitted, there is obviously no content by virtue of there being no file blob that contains this content.  If it is the empty blob digest, then there is a file blob that happens to have no content.

Personally I don't think there is a need to codify that no file and the empty file should be treated the same in implementations.

vors:
Yeah, I totally understand that it may seem overly pedantic / self-evident. The angel under which I find it useful is the following. Let’s say you are developing a suite of tests for REAPI to test the implementation correctness. Assuming the tests are hermetic you can actually match almost all fields on the ActionResult proto. In this circumstances, it’s desirable to have a clear cut for how the fields should be matched. A practical aspect of it is that it’s easier to communicate about the discrepancies between 2 implementations when spec is explicit about it.

vors:
Maybe a better comment extension could be about the method which backend SHOULD prefer to represent empty blobs for stderr and stdout (since it’s so common). I’d imagine that omitting the digest is more preferable because it reduces the size of ActionResult proto.

bergsieker (commented):
We would like to take a more holistic approach to this and clarify how to handle missing digests in a number of different contexts. I'll open an issue to track this.

EdSchouten:
Random braindump: I see that the number of places where we want to distinguish between inlining a blob and returning a reference to it is increasing. Maybe REv3 should address this by simply adding a oneof type?
```proto
message CASReference {
  oneof payload {
    bytes data = 1;
    Digest digest = 2;
  }
}
```
That way the logic to distinguish between both cases only needs to be implemented once.

vors:
That would be usefully imo, not sure it's worth breaking change though.

EdSchouten:
Exactly. It would make sense for a REv3.

sstriker:
I think there is consensus not to land this change.

Issue #162, opened 2020-08-11
CI hooks: generate go bindings


mostynb:
It would be good to add a CI hook, which either checks the generated go bindings are up to date, or generates them itself and updates the PR somehow.

sstriker:
+1, great idea.  Would this be achievable with GitHub Actions?

mostynb:
I believe so. Here's an example action which commits and then pushes to origin: https://lannonbr.com/blog/2019-12-09-git-commit-in-actions

Our action could run `hooks/pre-commit` when master is updated, and if there are any changes commit and push to master. I can investigate this next week, if you like.

mostynb:
Closing this, in favour of #177.

PR #124, opened 2020-03-07
Convert *:go_default_library should be alias of *:xyz_go_proto

Yannic:
Fixes #123

PR #164, opened 2020-08-17
[ci] autogenerate go bindings on master

mostynb:
This is a github action which attempts to keep the generated *.pb.go code in sync with the corresponding .proto files on master. This action runs when master is updated- it takes a few minutes to run and will push a new commit to master if there were any changes required.

I assume if there are more than one of these actions running at the same time and both try to push, only the first will succeed. We should try to avoid this situation by not landing changes until the previous action has completed.

Possible (temporary?) solution to #162.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,38 @@
+name: CI
+
+on:
+  push:
+    branches: [ master ]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+    # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
+    - uses: actions/checkout@v2
+
+    - name: Configure git
+      run: |
+        git config user.name "GitHub Actions Bot"
+        git config user.email "<>"
+
+    - name: Install bazelisk
+      run: |
+        curl -LO "https://github.com/bazelbuild/bazelisk/releases/download/v1.6.0/bazelisk-linux-amd64"
+        mkdir -p "${HOME}/bin"
+        mv bazelisk-linux-amd64 "${HOME}/bin/bazel"
+        chmod +x "${HOME}/bin/bazel"
+
+    - name: Run hooks
+      run: |
+        env
+        PATH=${HOME}/bin:$PATH ./hooks/pre-commit
+
+    - name: Check for changes
+      run: |
+        if ! git commit -m "regenerate pb.go files"
+        then
+          exit
+        fi
+        git push origin master


mostynb:
@sstriker: FYI

ulfjack:
Have we considered moving the Go bindings to a separate repo? This makes it so Go bindings lag behind the proto change, which also seems undesirable (it won't affect me personally).

mostynb:
> Have we considered moving the Go bindings to a separate repo? This makes it so Go bindings lag behind the proto change, which also seems undesirable (it won't affect me personally).

That would be a breaking change, unless we also keep the .pb.go files in the current location. It would also need to have the same go module version numbers to avoid confusion. The only benefit I see would be that bazel-based consumers woulnd't needlessly build protobuf and regenerate .pb.go files themselves.

Maybe we could achieve the same result by duplicating the pb.go files in a different directory of this repository, and let consumers opt in to using the newer import path?

sstriker (approved):
I am not intimately familiar with GitHub Actions, but this reads as not overstepping the scope of the PR description.

ulfjack:
Is anyone else concerned about the lag? Go developers may not realize that the bindings are in a separate commit from the one that changes the proto files, and may spend a lot of time debugging why their bindings are off. Personally, I also think Go bindings should not be checked in at all, but I suppose that ship has sailed. :-P

sstriker:
@ulfjack currently we run the risk of the go bindings being completely out of date, not just a commit apart.  There is probably more conversation needed around Go bindings, see also #111 and #124.  But I do think that in the mean time this is an improvement on the status quo.

ulfjack:
Sure.

bergsieker:
I agree that this is both better than the current situation, and not a perfect solution. We should investigate better solutions but I see no reason to block this in the meantime.

bergsieker (approved):

EdSchouten:
I agree with @ulfjack. What he proposes is also easy to achieve. Just add a check that copies the .pb.go files out of bazel-bin/, followed by running `git diff --exit-code HEAD --`. I see no reason that couldn’t be changed prior to merging.

bergsieker (commented):
Discussion from RE API monthly: this would be better as a presubmit. Mostyn, can you configure a GH action as a presubmit? @cphang99 might also be able to help.

mostynb:
If we do this when PR branches are updated then there's a race condition between overlapping PRs. @bergsieker: is that what you mean by "presubmit"?

Ideally we want this to happen serialized at "merge time", but github doesn't have a "pre-merge" event as far as I can tell:
https://docs.github.com/en/actions/reference/events-that-trigger-workflows

We could use a different event to trigger the merge instead of clicking the merge button, like adding a particular label. The github action would merge, run the hooks, push to master and close the PR. There would be no lag, but there's still the issue of concurrent merges (which is hard to solve without some external state/bot/queue/similar).

mostynb:
I dug into the "trigger a merge when the PR is labelled" idea a bit- unfortunately this doesn't work if the PR is created from a forked repository because the action runner only has write access to the source repository. To work around this, we could use a manually triggered or cron-style scheduled action that uses the github REST API to process PRs with a particular label. It's a bit more code, since it needs to query the REST endpoints, but has the benefit of being able to process PRs sequentially in a relatively safe manner (still theoretically racy unless we can find a way to prevent concurrent actions).

I setup a prototype over here: https://github.com/mostynb/github-action-test

First, there's a manual action with some basic setup:
https://github.com/mostynb/github-action-test/blob/master/.github/workflows/manual.yml

And here's the code that interacts with the REST API, pushes to git etc (working with json is too annoying in bash, but python would also work):
https://github.com/mostynb/github-action-test/blob/master/gha.go

Then you can label a few PRs as "merge-me", and trigger the action. For each open PR with the "merge-me" label, the PR is merged, pretend hooks are run, modified results are amended into the tip of the PR (this could alternatively be a separate commit), the master branch is pushed to origin, the label is removed from the PR, a comment is left and the PR is closed.

mostynb:
Closing this, since we plan to deprecate the generated files in this repository.

PR #167, opened 2020-09-01
RemoteEx 2.2: Promote platform properties from command to action

### LINKED ISSUE #166: Promoting or inlining platform properties
We discussed this during a monthly meeting (I seem to recall that @ulfjack initially raised it), adding here for tracking.

The current design of platform properties is that they are indirectly embedded in the [Command property of the Action](https://github.com/bazelbuild/remote-apis/blob/master/build/bazel/remote/execution/v2/remote_execution.proto#L392). The Action does not send the Command directly but rather sends a Command digest.

The upshot of this is that additional blob uploads are required from the client before submitting and action as well as additional CAS interactions on the server side if any data is required from the Command. For example, platform properties.

I think the initial discussion specifically mentioned that if the server wanted to making routing decisions for an action based on the platform properties it would require and additional hit to the CAS to determine those for the action.

It seems to me that some of this extra CAS interaction overhead could be avoided if we inlined the command or platform properties into the action.

I can see that we could probably:

1. Extend the action message to have both `command_digest` or the actual `command` 
2. Inline the platform properties
3. Something else?

edbaunton:
Platform properties are currently a member of the `command` message which is
referred to in the action by digest. This requires the Execution Service to
make a call to the CAS to retrieve the contents of the command if it wishes to
inspect it. The platform properties are commonly used for making routing
decision about the action. Therefore, in order to route an action common
execution service implementations must introduce an additional call to the CAS
to fully hydrate the action and determine where it should be routed.

This commit promotes the platform properties from the command to the action. We
deprecate the platform properties contained within the action and bump the
minor version to version 2.2, following the model for `output_paths`.

Fixes #166

Signed-off-by: Ed Baunton <ebaunton1@bloomberg.net>


ulfjack:
This seems reasonable.

EdSchouten (approved):
Looks good!

Question: should we make it a bit more explicit what a transition path looks like? Should clients provide both for the time being? Should the one in Action be preferred over the one in Command?

At the same time, some time ago I got a request on the Buildbarn side whether it's required to specify platform properties at all. Is it valid to send requests where both Action.platform and Command.platform are null/nil/.../None, or is it required that at least one of them contains a message?

edbaunton:
I am not 100% familiar with the right way to perform a proto field migration but I tried to update the docs as I would myself have expected to do it. Basically if you use 2.2 you must use the new field, if you are not using 2.2 you must populate the old and can populate the new if you want. From 2.2 you shouldn't populate the old.

From your comment @EdSchouten I also added in an 'optional' to the new field to denote that they can be empty. I didn't change that for the deprecated field.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -433,6 +433,26 @@ message Action {
...
+  // [ServerCapabilities][build.bazel.remote.execution.v2.ServerCapabilities]'s
+  // high_api_version as at least 2.2 MUST populate this field instead of that
+  // which was deprecated inside the
+  // [Command][build.bazel.remote.execution.v2.Command].  If the server does
+  // not advertise itself as supporting 2.2 then the client MUST populate the
+  // platform properties inside the
+  // [Command][build.bazel.remote.execution.v2.Command] as before and SHOULD
+  // populate this new member too with the same information to facilitate
+  // upgrading.
+  // Execution services conforming to 2.2 MUST ignore platform properties
```
EdSchouten, line 453: With this requirement in place, how will a REv2.1 client be able to run actions on a REv2.2 cluster? The problem is that there's no way to attach desired protocol versions to individual requests. There is no two-way version negotiation between client and server.

I think all of this is far too exact. Relying on API versions, etc. Isn't it sufficient to just say:
- Clients SHOULD set the platform properties both in Action and Command.
- Servers SHOULD prefer the platform properties in the Action over the ones specified in the Command.
- Command.platform is deprecated and will be removed in REv3.

EdSchouten (commented):
Giving this some more thought, one 'downside' of this approach is that clusters can no longer use the command digest as a key for correlating equivalent actions. For example, inspecting historical build actions to see whether their resource usage increased over time, etc..

That said, clusters are still capable of achieving this by introducing their own composite type:
```proto
message Key {
  Digest command_digest = 1;
  Platform platform = 2;
}
```
And hashing that to obtain a proper key for stuff like that. There's no need to account for that in the client <-> cluster protocol.

build/bazel/remote/execution/v2/remote_execution.proto
edbaunton: My aim was to reduce the amount of redundant data being sent when the server clearly advertises that it's at 2.2, which it can through capabilities. Totally fine to relax this however.

edbaunton:
How do others feel about how strict we should be with version migrations?

edbaunton:
I have simplified the documentation to keep this simple change moving.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -559,6 +569,12 @@ message Command {
   // the client SHOULD ensure that running the action on any such worker will
   // have the same result.
   // A detailed lexicon for this can be found in the accompanying platform.md.
+  // DEPRECATED as of v2.2: platform properties are now specified directly in
+  // the action to allow execution services to make routing decisions based on
+  // these without introducing an additional step to reach the CAS to determine
+  // the properties. See documentation note in the
+  // [Action][build.bazel.remote.execution.v2.Action] for migration.
+  // This will be deprecated in v3.
```
bergsieker, line 577: Presumably this will actually be removed in v3, not deprecated. Regardless, I don't think it's necessary to say that explicitly here.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -559,6 +569,12 @@ message Command {
   // the client SHOULD ensure that running the action on any such worker will
   // have the same result.
   // A detailed lexicon for this can be found in the accompanying platform.md.
+  // DEPRECATED as of v2.2: platform properties are now specified directly in
+  // the action to allow execution services to make routing decisions based on
+  // these without introducing an additional step to reach the CAS to determine
+  // the properties. See documentation note in the
```
bergsieker, line 575: Remove this from the comment. The rationale is already in the PR description (where it should be).

bergsieker (approved):

edbaunton:
Keen for one more approver @EdSchouten and @ulfjack if you have a chance

EdSchouten (approved):

sstriker (approved):

PR #173, opened 2020-09-08
Add optional support for interleaved stdout/err outputs

erikma:
Broken out and somewhat modified from https://github.com/bazelbuild/remote-apis/pull/155 based on comments in that PR. Not all comments there were resolved, discussion can continue here.

### DIFF
--- a/build/bazel/remote/execution/v2/remote_execution.proto
+++ b/build/bazel/remote/execution/v2/remote_execution.proto
@@ -1029,6 +1029,49 @@ message ActionResult {
 
   // The details of the execution that originally produced this result.
   ExecutedActionMetadata execution_metadata = 9;
+
+  // Inlined [ConsoleOutputs][build.bazel.remote.execution.v2.ConsoleOutputs]
+  // containing a series of stdout and stderr outputs in order of write from
+  // the action. The server SHOULD NOT inline `console_outputs` unless
+  // requested by the client in the
+  // [GetActionResultRequest][build.bazel.remote.execution.v2.GetActionResultRequest]
+  // message. The server MAY omit inlining, even if requested, and MUST do so if inlining
+  // would cause the response to exceed message size limits.
+  ConsoleOutputs console_outputs = 10;
+
+  // The digest for a blob containing the
+  // [ConsoleOutputs][build.bazel.remote.execution.v2.ConsoleOutputs] for the action,
+  // which can be retrieved from the
+  // [ContentAddressableStorage][build.bazel.remote.execution.v2.ContentAddressableStorage].
+  Digest console_outputs_digest = 11;
+}
+
+// A sequence of console outputs in the order received for a remoted process.
+// This may contain the interleaved outputs from a process in the order
+// received during execution. If the service does not implement
+// interleaving, this may contain a summary of the stdout and stderr output
+// streams as single large writes in a server-defined order.
+message ConsoleOutputs
+{
+  // An output record for one or more outputs to stdout or stderr.
+  //
+  // Design note: For text coloring and other effects, ANSI codes may be
+  // embedded within these strings. Windows console API calls are not
+  // supported in this protocol, but a server implementation MAY choose to
+  // translate console API calls into ANSI codes, as more recent Windows 10
+  // releases included ANSI code support.
+  message ConsoleOutput {
+    // When true, the `Writes` were directed at the stderr stream.
+    bool is_stderr = 1;
+    
+    // One or more written strings. There is an implicit operating system
+    // dependent newline after each string. Each string may contain embedded
+    // OS-dependent newline character sequences.
+    repeated string writes = 2;
+  }
+
+  // A series of stdout and stderr outputs in order of write from the remoted process.
+  repeated ConsoleOutput outputs = 2;
 }
 
 // An `OutputFile` is similar to a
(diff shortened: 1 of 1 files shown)


build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1029,6 +1029,54 @@ message ActionResult {
...
+  // [ContentAddressableStorage][build.bazel.remote.execution.v2.ContentAddressableStorage].
+  Digest console_outputs_digest = 11;
+}
+
+// A sequence of console outputs in the order received for a remoted process.
+// This may contain the interleaved outputs from a process in the order
+// received during execution. If the service does not implement
+// interleaving, this may contain a summary of the stdout and stderr output
+// streams as single large writes in a server-defined order.
+message ConsoleOutputs
```
EdSchouten, line 1054: Style nit: place { on the same line.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1029,6 +1029,54 @@ message ActionResult {
...
+      STDERR = 1;
+    }
+    
+    // The output stream to which the `Writes` were directed.
+    Stream stream = 1;
+    
+    // One or more written strings. There is an implicit operating system
+    // dependent newline after each string. Each string may contain embedded
+    // OS-dependent newline character sequences.
+    repeated string writes = 2;
```
EdSchouten, line 1075: Would this be a good occasion to add per-line timestamp support?

Also a followup question from the previous PR: why are you avoiding doing string concatenation? String concatenation likely gives you a more compact representation than using a repeated string field.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1029,6 +1029,54 @@ message ActionResult {
...
+  message ConsoleOutput {
+    enum Stream {
+      STDOUT = 0;
+      STDERR = 1;
+    }
+    
+    // The output stream to which the `Writes` were directed.
+    Stream stream = 1;
+    
+    // One or more written strings. There is an implicit operating system
```
EdSchouten, line 1072: I don’t think we should add implicit newlines. The issue is that some POSIX applications mix stdout and stderr output on single lines. By using implicit newlines, there would be no way to express the output of those.

Though slightly contrived, be sure to read the POSIX spec of the `fuser` utility.

moroten: I'm also in favour of explicit newlines. If an application that chooses to use LF instead of CRLF under Windows, the output would have to be concatenated into a single entry. Would one have to look into the platform fields of an action to find out if LF or CRLF should be used? How would outputs without trailing newline be handled with implicit newline?

EdSchouten: Alternatively, we could simply standardize on one convention. Even for UNIX it would make sense to use \r\n, as that is actually what’s emitted on the console. Userspace applications write \n, but termios converts it to \r\n in case ONLCR is enabled.

That said, I’m not 100% sure it’s safe to use pseudo-terminals instead of pipes on UNIX. That may cause applications to assume they need to be run interactively. Maybe the best strategy is to document that applications can both emit \n and \r\n, and that it’s up to the tool that displays the logs to seamlessly support both formats.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1029,6 +1029,54 @@ message ActionResult {
...
+  // containing a series of stdout and stderr outputs in order of write from
+  // the action. The server SHOULD NOT inline `console_outputs` unless
+  // requested by the client in the
+  // [GetActionResultRequest][build.bazel.remote.execution.v2.GetActionResultRequest]
+  // message. The server MAY omit inlining, even if requested, and MUST do so if inlining
+  // would cause the response to exceed message size limits.
+  ConsoleOutputs console_outputs = 10;
+
+  // The digest for a blob containing the
+  // [ConsoleOutputs][build.bazel.remote.execution.v2.ConsoleOutputs] for the action,
```
EdSchouten, line 1043: The downside of this approach is that it’s memory inefficient to load logs in case an application writes a lot of them. Could we consider letting this use that Protobuf RecordIO file format?

EdSchouten: Hmmmm... giving it some more thought.

The schema that we use is awfully similar to what Docker already uses to store stdout and stderr output of containers in logs: https://docs.docker.com/config/containers/logging/json-file/ This is I think also the format that Kubernetes uses internally, though I’m not 100% sure.

Would there be a way (and/or desire) to converge on format here? What about using the same field names in the Protobufs, so that jsonpb is capable of exchanging these formats transparently?

For the externally (non-inlined) logs, maybe it makes sense to use exactly the same format as Docker?

build/bazel/remote/execution/v2/remote_execution.proto
bergsieker: Given the use case--sending back the "raw" console output as it would have appeared if the action had been run locally--I'm inclined toward's Ed's suggestion to make the API as "raw" as possible and let it be the responsibility of the client to interpret the output properly. This is similar reasoning to how we're handling absolute paths (which are OS-dependent) and relative paths (which are not, and therefore use a "generic" separator regardless of OS).

build/bazel/remote/execution/v2/remote_execution.proto
bergsieker: This is certainly the place to add it if it's required...

build/bazel/remote/execution/v2/remote_execution.proto
ulfjack: For posterity: Tools often generate binary output and any post-processing of that output would break that. For example, gzip can write output to stdout, as can tar, as well as a variety of other CLI tools. Bazel does as well in some cases - this already broke Google-internal API definition that assumed that Bazel always outputs valid UTF-8. Post-processing stdout/stderr would be a very, very bad idea.

erikma:
Closing in favor of #188.

PR #176, opened 2020-10-12
Add riscv ISA values to the Platform Lexicon document.

robjh:
This adds the following values to the ISA list in platform.md
 - riscv32
 - riscv64

GCC, Linux and LLVM seem to agree on these names so this ought to be all that's needed.

### DIFF
--- a/build/bazel/remote/execution/v2/platform.md
+++ b/build/bazel/remote/execution/v2/platform.md
@@ -35,6 +35,8 @@ The following standard property `name`s are defined:
     - `arm-t32-be` (big endian)
     - `power-isa-be` (big endian)
     - `power-isa-le` (little endian)
+    - `riscv32`
+    - `riscv64`
     - `sparc-v9` (big endian)
     - `x86-32`
     - `x86-64`


EdSchouten:
Question: what about RISC-V optional features, such as division, floating point arithmetic, etc.? How do those get encoded in the platform properties?

bergsieker:
At the monthly sync, the general feeling was that using the official RISC-V naming is probably the right way to go here. Rob, if you can update (and complete the CLA) we'll be happy to merge this.

robjh:
hi @bergsieker 
I assume you mean ```RV32I``` and ```RV64I```.
I'm unsure how to handle the extentions as @EdSchouten points out, as it seems there would be several varying combinations.
Maybe I could add something along the lines of ```RV32[I,E][M,A,F,D,Q,C]``` and ```RV64I[M,A,F,D,Q,C]```?

sstriker:
@robjh it's been a while since this pull request has been updated.  Are you still interested in this change?

sstriker:
Closing as there doesn't seem to be an active interest at this time.

nanonyme:
@sstriker there is definitely interest in getting this merged but we haven't found anyone who has signed CLA to re-submit this. We were earlier discussing with @juergbi that just having RV32I and RV64I basic ISA's would solve the BuildStream requirements for this spec. There is no intent to disambiguate based on extensions for that use case.

Issue #177, opened 2020-10-13
Remove generated go bindings


bergsieker:
Many of the open PRs and issues on this repository have to do with the go bindings. This problem seems generally intractable in an open source environment: we could probably patch up the current issues and add CI to help us keep on top of things, but that only really allows us to support a single version of Go. What about C# or other languages? What about different versions of Go?

We should deprecate the existing bindings and remove them once clients have had a chance to migrate. If we don't want to go all the way to requiring clients to generate their own bindings, one possibility would be to move the bindings to the remote-apis-sdks repo. Not sure if that's viable or not, but it seems like a better fit given that that repo is already an SDK...

bergsieker:
@ola-rozenfeld What are your thoughts on moving these bindings to remote-apis-sdks?

sstriker:
@ola-rozenfeld I prefer to think about this repo as where we document the Remote APIs.  We happen to express these in the form of a proto definition, with supporting documentation in markdown files.  For implementations of the APIs I don't think it is unreasonable to have to translate from the specification to insert-favorite-programming-language.  And that translation may happen via tools, or manually, as long as the end result follows the specification.  As such, I think the SDK repo is an excellent place to put language specific artifacts.  Make sense?

EdSchouten:
The googleapis repo also doesn’t provide generated source files, and we rely on it for bytestream and longrunning. With that in mind, I don’t think it makes sense for us to provide generated sources on our end.

ola-rozenfeld:
Okay, then I don't mind if we follow suit and move all language-specific stuff to the clients (maybe only Go to start?).
This will require to migrate the existing Go clients first, and then remove the bindings -- right?

mostynb:
I wonder if we could solve #105 at the same time?

bergsieker:
Mostyn: yes, I believe this will fix that issue as well (by pushing it back to the consuming repo). At the monthly sync we determined that a large chunk of the open issues and PRs on this repo have to do with go bindings, which it doesn't really make sense to include in this repo.

hunshcn:
What is the current state? pb.go has not been updated for a long time.

We need to make some changes, delete or update it.

EdSchouten:
My recommendation would actually to keep it. Or alternatively, provide another repo where we serve pre-generated copies of this file. Otherwise there is no uniform URL under which this protocol can be imported.

bergsieker:
[...]

PR #182, opened 2020-11-28
Fix go bindings generation

mostynb:
Followup to #154, which forgot to update the bazel configuration.

This makes hooks/pre-commit work again, and adds a simple github action to confirm this still works. If we wanted to test the check, we could land the check first, a test PR to verify that the check fails, then update this PR to trigger a check and confirm that it succeeds.


EdSchouten:
Why doesn’t CI also enforce this?

EdSchouten (approved):

mostynb:
> Why doesn’t CI also enforce this?

This repository doesn't actually have CI. I could add a simple github action with tests though.

.github/workflows/main.yml
```diff
@@ -0,0 +1,31 @@
+name: CI
+
+on:
+  push:
+    branches: [ master ]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
```
EdSchouten, line 9: You could add:
```yaml
container: docker://l.gcr.io/google/bazel:<some version number>
```
That way the action runs on a container that has Bazel immediately present, meaning that you don't need to install Bazelisk.

.github/workflows/main.yml
```diff
@@ -0,0 +1,31 @@
...
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Configure git
```
EdSchouten, line 14: Should we add some kind of comment explaining why this is necessary?

.github/workflows/main.yml
```diff
@@ -0,0 +1,31 @@
...
+        curl -LO https://github.com/bazelbuild/bazelisk/releases/download/v1.7.4/bazelisk-linux-amd64
+        mkdir -p "${HOME}/bin"
+        mv bazelisk-linux-amd64 "${HOME}/bin/bazel"
+        chmod +x "${HOME}/bin/bazel"
+
+    - name: Check bindings generation
+      run: |
+        set -e
+        PATH=${HOME}/bin:$PATH ./hooks/pre-commit
+        # This might stage some files, if the bindings need to be regenerated.
```
EdSchouten, line 30: You could add a step that runs `git diff --exit-code HEAD --`.

.github/workflows/main.yml
mostynb: I think this is only needed if we try to commit the changes that the hook stages- we can just remove it.

.github/workflows/main.yml
mostynb: If we do this, then the check will fail on all PRs that haven't committed the updated go bindings- I'm not sure that we want to do that (in which case, maybe it would be better to `git commit --amend --no-edit`).

.github/workflows/main.yml
mostynb: Those container images seem to lag behind the released bazel versions more than bazelisk, the latest version there is 3.5.0 while the latest bazel release is 3.7.1
https://console.cloud.google.com/gcr/images/cloud-marketplace-containers/GLOBAL/google/bazel?gcrImageListsize=30

Not that it matters too much either way.

sstriker:
Weren't we going to drop the bindings completely?  Not sure if we want to go the opposite direction with CI?

mostynb:
> Weren't we going to drop the bindings completely as per #177? Not sure if we want to go the opposite direction with CI?

This github action doesn't commit the changes- it just runs the hook to confirm that it works.

sstriker (approved):
As per the Dec 8th meeting.

mostynb:
It looks like the github action confirmed that the bindings generation works again:
https://github.com/bazelbuild/remote-apis/runs/1527006790

PR #183, opened 2020-11-30
fix missing deps

atetubou:
This fixes following build error

```
build/bazel/remote/execution/v2/remote_execution.proto: google/protobuf/any.proto is imported, but //build/bazel/remote/execution/v2:remote_execution_proto doesn't directly depend on a proto_library that 'srcs' it.
```


mostynb:
See also #182 - which has this fix + adds a CI check.

atetubou:
Oh, thank you.

PR #184, opened 2021-01-06
Add the Remote Output API

EdSchouten:
As being discussed in this thread, I am currently working on letting
Bazel store its output directory on a FUSE file system:

https://groups.google.com/g/remote-execution-apis/c/qOSWWwBLPzo/m/TbcNYzwMBwAJ

The plan is to let a separate daemon manage the FUSE daemon. It's not
going to be part of Bazel itself. This is why I've worked on a gRPC
protocol that Bazel can use to send instructions to the FUSE daemon.
This commit brings in a first version of this protocol.

### DIFF
--- a/build/bazel/remote/output/v1/remote_output.proto
+++ b/build/bazel/remote/output/v1/remote_output.proto
@@ -1,0 +1,309 @@
+// Copyright 2021 The Bazel Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+
+package build.bazel.remote.output.v1;
+
+import "build/bazel/remote/execution/v2/remote_execution.proto";
+import "google/protobuf/empty.proto";
+
+option go_package = "remoteoutput";
+option java_multiple_files = true;
+option java_outer_classname = "RemoteOutputProto";
+option java_package = "build.bazel.remote.output.v1";
+
+// The Remote Output Service may be used by users of the Remote
+// Execution API to construct a directory on the local system that
+// contains all output files of a build.
+//
+// Primitive implementations of this API may simply download files from
+// the Content Addressable Storage (CAS) and store them at their
+// designated location. Complex implementations may use a pseudo file
+// system (e.g., FUSE) to support deduplication, lazy loading and
+// snapshotting.
+service Output {
+  // Methods that can be invoked at any point in time.
+
+  // Clean all data associated with a single output path, so that the
+  // next invocation of StartBuild() yields an empty output path. This
+  // may be implemented in a way that's faster than removing all of the
+  // files from the file system manually.
+  rpc Clean(CleanRequest) returns (google.protobuf.Empty);
+
+  // Signal that a new build is about to start.
+  //
+  // The client uses this call to obtain a directory where outputs of
+  // the build may be stored, called the output path. Based on the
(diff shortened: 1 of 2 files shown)


build/bazel/remote/output/v1/remote_output.proto
```diff
@@ -0,0 +1,295 @@
...
+  // In case the path corresponds to a symbolic link, include the target
+  // of the symbolic link in the response.
+  bool include_symlink_target = 3;
+
+  // If the last component of the path corresponds to a symbolic link,
+  // return the status of the file at the target location.
+  //
+  // Symbolic links encountered before the last component of the path
+  // are always expanded, regardless of the value of this option.
+  bool follow_symlinks = 4;
```
EdSchouten, line 222: I am somewhat unsure about this part of the API. For Bazel's getFastDigest(), we want to follow symlinks. For [BatchStat.batchStat()](https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/vfs/BatchStat.java), it looks like we don't want to follow symlinks. This is why the `follow_symlinks` option is there.

But what about the `include_file_digest` option? Bazel never calls BatchStat.batchStat() with this flag set to false. Maybe we should remove that option and always set it to true?

Something similar holds for `include_symlink_target`. I suppose it's a nice feature to have in case you want to bypass the VFS, but there is nothing in Bazel that uses it. Maybe I should just remove it, and change FIleStatus.symlink to a google.protobuf.Empty.

build/bazel/remote/output/v1/remote_output.proto
```diff
@@ -0,0 +1,295 @@
...
+  // The identifier of a finalized build whose results are stored in the
+  // output path.
+  string build_id = 1;
+
+  // Paths that have been modified since the build finalized.
+  //
+  // If the remote output service freezes the contents of the output
+  // path between builds, this field can be left empty.
+  repeated string modified_paths = 2;
+}
```
EdSchouten, line 132: The protocol currently doesn't describe how we need to deal with remote files that disappear. A couple of options:

- We don't tell it to Bazel at the start of a build. We let the build run and somehow communicate NOT_FOUND errors. Bazel could then convert that to a LostInputsActionExecutionException. That said, I don't know whether that part of Bazel is fully functional, or whether it's something that only works well in Blaze. It looks like this requires a per-action file system, which isn't in scope for this protocol (yet). It also looks path based, which rules out the possibility of hardlinking/deduplicating identical files to a single inode at the FUSE level.
- We require that StartBuild() does a full scan of the file system and removes any files that are no longer present remotely. Any files that were absent are reported through InitialOutputPathContents.modified_paths. This means we can't do any backtracking in case files disappear during a build. This would be acceptable for Buildbarn at least, as Buildbarn's storage setup is strong enough to guarantee this in practice.

Do we think the second approach is acceptable?

build/bazel/remote/output/v1/remote_output.proto
```diff
@@ -0,0 +1,295 @@
...
+  rpc FinalizeBuild(FinalizeBuildRequest) returns (google.protobuf.Empty);
+}
+
+message CleanRequest {
+  // The output base identifier that was provided to
+  // StartBuildRequest.output_base_id whose data needs to be removed.
+  string output_base_id = 1;
+}
+
+message StartBuildRequest {
```
EdSchouten, line 68: There is currently no AP that allows Bazel to check whether a build action failed due to a retriable/non-retriable I/O error. One complication is that Bazel likely wants to know this information at the action level, even though the remote output service is oblivious of actions.

Maybe we should add a (set of) RPC(s) to read the last I/O error associated with the current build, and provide a way to clear the error in case Bazel wants to do a retry? Maybe this is also sufficient to deal with NOT_FOUND errors, or do we consider that to be orthogonal?

build/bazel/remote/output/v1/remote_output.proto
EdSchouten: As suggested by @EricBurnett on the mailing list, the second option would be preferred. I've extended the description of StartBuild() to require that FindMissingBlobs() is called.

janakdr:
Adding @alexjski because he's interested in how Bazel can manage its output file tree.

sstriker:
@EdSchouten do you wish to change this to a Draft PR for the time being?

PR #185, opened 2021-01-18
improve BUILD files

tschuett:
I can now build my project without the 
build_file = "BUILD.googleapis"
hack.

### DIFF
--- a/build/bazel/remote/asset/v1/BUILD
+++ b/build/bazel/remote/asset/v1/BUILD
@@ -1,7 +1,7 @@
-package(default_visibility = ["//visibility:public"])
+package(default_visibility=["//visibility:public"])
 
 load("@rules_proto//proto:defs.bzl", "proto_library")
-load("@bazel_remote_apis_imports//:imports.bzl", "java_proto_library", "go_library", "go_proto_library", "cc_grpc_library")
+load("@bazel_remote_apis_imports//:imports.bzl", "cc_grpc_library", "go_library", "go_proto_library", "java_proto_library")
 
 licenses(["notice"])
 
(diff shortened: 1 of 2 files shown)


build/bazel/remote/asset/v1/BUILD
```diff
@@ -1,7 +1,7 @@
-package(default_visibility = ["//visibility:public"])
+package(default_visibility=["//visibility:public"])
```
EdSchouten, line 1: Out of curiosity, did buildifier remove this whitespace? Odd...

build/bazel/remote/asset/v1/BUILD
```diff
@@ -12,9 +12,9 @@ proto_library(
         "//build/bazel/remote/execution/v2:remote_execution_proto",
         "@com_google_protobuf//:duration_proto",
         "@com_google_protobuf//:timestamp_proto",
-        "@googleapis//:google_api_annotations_proto",
-        "@googleapis//:google_api_http_proto",
-        "@googleapis//:google_rpc_status_proto",
+        "@googleapis//google/api:annotations_proto",
```
EdSchouten, line 15: Nice! I hope that this also makes my life easier on the Buildbarn side. I currently apply this local change to be able to build these Protobufs.

https://github.com/buildbarn/bb-storage/blob/master/patches/com_github_bazelbuild_remote_apis/golang.diff

EdSchouten (approved):

build/bazel/remote/asset/v1/BUILD
EdSchouten: Hmm... And the build still fails. Maybe you need to upgrade `http_archive(name = "googleapis")` in `WORKSPACE` as well?

build/bazel/remote/asset/v1/BUILD
tschuett: I use https://github.com/codesuki/bazel-mode. It calls bazel-format on save.

tschuett:
I run buildifier manually over the BUILD files. The changes are gone. Sorry for the noise.

build/bazel/remote/asset/v1/BUILD
tschuett: I don't understand the error :-(

tschuett: I works for me locally.

mostynb: Could it be a bazel version difference?

tschuett: I use 3.7.2.

tschuett: The CI uses basilisk and Bazel version 3.7.2

tschuett:
Sorry for the delay, but I never used Go before.

PR #186, opened 2021-02-08
Add additional optional metadata fields

illicitonion:


build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1690,4 +1690,15 @@ message RequestMetadata {
...
+  // A one-word description of the kind of action, for example, CppCompile or GoLink.
+  // There is no standard agreed set of values for this, and they are expected to vary between different client tools.
+  string action_mnemonic = 5;
+
+  // A namespace within which the build was performed, for instance a repository/workspace name.
+  string invocation_namespace_id = 6;
+
+  // An identifier for the target, unique within the invocation_namespace_id, which produced this action.
+  // No guarantees are made around how many actions may relate to a single target.
+  string target = 7;
```
EdSchouten, line 1703: The Build Event Stream identifies individual nodes in the build graph by a Bazel target and a Configuration ID:
https://github.com/bazelbuild/bazel/blob/dd8afa0fc79a7ff0fc0633eed6d592e0efd3bbfc/src/main/java/com/google/devtools/build/lib/buildeventstream/proto/build_event_stream.proto#L135-L138

The reason for this is that a single target can be compiled multiple times (e.g., through aspects, transitions, etc.). 

Should we also add this information to the metadata, so that we can correlate REv2 RPCs with build events?

illicitonion: Sounds good - added

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1690,4 +1690,21 @@ message RequestMetadata {
   // An identifier to tie multiple tool invocations together. For example,
   // runs of foo_test, bar_test and baz_test on a post-submit of a given patch.
   string correlated_invocations_id = 4;
+
+  // A one-word description of the kind of action, for example, CppCompile or GoLink.
+  // There is no standard agreed set of values for this, and they are expected to vary between different client tools.
+  string action_mnemonic = 5;
+
+  // A namespace within which the build was performed, for instance a repository/workspace name.
```
EricBurnett, line 1698: Consider also adding something about the expected (or not) global uniqueness of this. E.g. "May not be globally unique, as with multiple users having workspaces of the same name".

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1690,4 +1690,21 @@ message RequestMetadata {
...
+  string invocation_namespace_id = 6;
+
+  // An identifier for the target, unique within the invocation_namespace_id, which produced this action.
```
EricBurnett, line 1701: I'm not sure what it means to be "unique within the invocation_namespace_id". I could see unique within the build (tool_invocation_id) though even that gets confusing with multiple configurations, but I'm not sure how to extend that to multiple invocations on a namespace. Maybe call out what can be inferred from this uniqueness?

(Also, this may require invocation_namespace_id be globally unique).

illicitonion: What this is trying to get at is basically: Within one namespace, two actions with the same target were generated by the same target. (Which, if you can rely on your users' namespace_ids being unique, is a strong statement, and otherwise probably just means that within a tool_invocation_id two actions with the same target were generated by the same target). I can just remove the comment about uniqueness, if you think that would be more clear?

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1690,4 +1690,22 @@ message RequestMetadata {
...
   // runs of foo_test, bar_test and baz_test on a post-submit of a given patch.
   string correlated_invocations_id = 4;
+
+  // A one-word description of the kind of action, for example, CppCompile or GoLink.
+  // There is no standard agreed set of values for this, and they are expected to vary between different client tools.
+  string action_mnemonic = 5;
+
+  // A namespace within which the build was performed, for instance a repository/workspace name.
+  // There are no guarantees that these are globally unique, e.g. two users may create namespaces with the same name.
+  string invocation_namespace_id = 6;
```
ulfjack, line 1700: If it's per-invocation, putting it in the per-action metadata is *not* ideal. I've previously seen cases where metadata ended up being more than 50% of the network traffic, with most of it being redundant. Please don't.

illicitonion: That's a reasonable concern - the same is presumably true for `correlated_invocations_id` - I expect that's tied to a single `tool_invocation_id`?

Perhaps both of these fields could be attached to `GetCapabilities` (either as in-line metadata in the request, or as header metadata) so that a remote execution system can record an association between them, and then they could be omitted from the per-request headers?

EdSchouten: An alternative would be to only attach the bare minimum here, and rely on the Build Event Stream to provide all other info. More concretely, this info should be part of the metadata:
- invocation ID
- target name
- configuration ID
But this info could be part of the BES:
- correlated invocations ID
- tool name and version

illicitonion: If you happen to be running a BES, I think that makes complete sense, but I'm not sure closely coupling those two systems is broadly applicable enough to have one depend on the other.

ulfjack: Can we leave this field out for now and check the rest in? :-D The other new items seem obviously a good idea to add to the proto.

illicitonion: Totally - updated to remove that field :)

ulfjack (approved):

EdSchouten (approved):
Ship it!

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1690,4 +1690,18 @@ message RequestMetadata {
   // An identifier to tie multiple tool invocations together. For example,
   // runs of foo_test, bar_test and baz_test on a post-submit of a given patch.
   string correlated_invocations_id = 4;
+
+  // A one-word description of the kind of action, for example, CppCompile or GoLink.
```
bergsieker, line 1694: Do we intend to enforce this as being one word? Or can we just say "brief" (which is unenforceable, but at least it's vague enough to make it clear that it won't be formally enforced).

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1690,4 +1690,18 @@ message RequestMetadata {
...
+  // There is no standard agreed set of values for this, and they are expected to vary between different client tools.
+  string action_mnemonic = 5;
+
+  // An identifier for the target which produced this action.
+  // No guarantees are made around how many actions may relate to a single target.
+  string target = 6;
```
bergsieker, line 1700: The other similar fields all use the suffix 'id' so I think this should be target_id.

illicitonion: Done

build/bazel/remote/execution/v2/remote_execution.proto
illicitonion: Changed to "brief" :)

bergsieker (approved):

bergsieker:
I'll merge this tomorrow morning unless someone else leaves comments.

illicitonion:
@bergsieker friendly ping? :)

coeuvre:
Please import this change as well.

PR #188, opened 2021-03-09
Add support for interleaved stdout/stderr outputs

EdSchouten:
REv2 currently stores stdout/stderr as two separate objects. A downside
of this approach is that information gets lost on how log messages are
ordered between these two streams. Another issue is that no timestamp
information is stored. When tests fail due to timeouts, it may be useful
to figure out which parts of the test take long to run.

This proposal is an alternative to PR #173. Whereas PR #173 defined
Protobuf messages containing individual console messages, this PR
proposes that they are stored using JSON, using the same schema as
what's generated by the Docker daemon:

https://docs.docker.com/config/containers/logging/json-file/

It doesn't make a lot of sense to invent our own standard for storing
console messages if there is already a commonly used format out there.


EdSchouten:
Cc @erikma

EdSchouten:
Hi @erikma, @EricBurnett,

I've just updated the protocol to reflect what we agreed on during the monthly Remote Execution working group meeting. Please take a look and let me know whether a change like this is acceptable from your side.

tdyas:
What is the motivation for deprecating `stdout`/`stderr` versus just supplementing them with `console`?

`stdout`/`stderr` as separate streams still seems like something that a build tool might want to retrieve separately especially if the invoked build action's output does not fit nearly into this console approach. At the very least, it would require extra work for a client to parse out just one stream from console.

Does deprecation now imply that `stdout`/`stderr` as separate streams will be dropped from REAPI v3?

EdSchouten:
> What is the motivation for deprecating `stdout`/`stderr` versus just supplementing them with `console`?

Storing both formats adds unnecessary redundancy. During the Remote API working group meeting we had earlier this week, several people (Sander Striker, I, others) expressed that they preferred not having a proliferation of different mechanisms for doing the same thing.

> `stdout`/`stderr` as separate streams still seems like something that a build tool might want to retrieve separately especially if the invoked build action's output does not fit nearly into this console approach. At the very least, it would require extra work for a client to parse out just one stream from console.

Is it common for clients to only download stdout or stderr (so not both)? I mean, it's perfectly valid for them to display it separately, but is it common to only load exactly one of them while not caring about the other one? Assuming clients generally want both of them, then this new approach is even simpler, in that it reduces the number of operations against the CAS.

With regards to extra work to parse out just one stream from the console: it's not so bad. It can be done in a (quasi-)streaming fashion. In the case of Java, you can call parseDelimited() to read a chunk, followed by filtering entries by the right stream.

> Does deprecation now imply that `stdout`/`stderr` as separate streams will be dropped from REAPI v3?

I hope so.

peterebden:
I'm not sure about this as the only future direction of output in REAPI. I take the point that it's one operation against the CAS to get both streams, but from our perspective we only ever fetch both on failure (on success we fetch stdout in a limited set of cases), in which case the download overhead probably isn't a big concern. It will also increase the number of bytes moved, possibly by a fair factor.
Conversely, I do worry that it introduces another format to be understood and parsed in an already complex API; the existing digests are easy to reason about and don't need any real client-side processing to be used.

I'd be tempted to go for a third optional field that is the combination of both streams (with the caveat that servers have to populate the individual stdout/stderr digests as well for tools that care). I suppose I am deliberately biasing to simplicity rather than generality, so maybe it doesn't answer your use case?

EdSchouten:
Adding a third field would work for my use case. It adds the functionality that I want. I don't think it's preferable, because it adds multiple ways of achieving the same thing. I could live with it.

That said, I don't agree with your reasoning. Let me explain why:

> It will also increase the number of bytes moved, possibly by a fair factor. [...] I'd be tempted to go for a third optional field that is the combination of both streams (with the caveat that servers have to populate the individual stdout/stderr digests as well for tools that care).

Doesn't keeping three streams around in the long run make this even worse?

Consider a client that also does plain remote caching (potentially in addition to remote execution). Such a client now needs to upload three output streams into a CAS when storing an AC entry if it wants to make use of this feature. The same applies to workers, of course. It is not always the case that workers run close to storage. People may run their build infra in the cloud, while keeping specialized workers on-prem (e.g., development boards). Those may also have bandwidth limitations.

> Conversely, I do worry that it introduces another format to be understood and parsed in an already complex API

REv2 is a complex API, because it tries to solve a complex problem. Referring to Fred Brooks's paper titled [No Silver Bullet](https://web.archive.org/web/20160910002130/http://worrydream.com/refs/Brooks-NoSilverBullet.pdf), it's mainly essential complexity -- not accidental complexity.

sstriker:
@EdSchouten worth bringing up-to-date and putting on next months agenda to hopefully close out?

PR #189, opened 2021-04-06
Add support for the Deflate compression algorithm

EdSchouten:
I don't think that it's a good idea to use Deflate in general. That
said, the reason I'm proposing that we add it is because it allows a
remote execution service to do light-weight "extraction" of ZIP files.
An existing ZIP file can be carved up into individual file payloads,
without decompressing/recompressing them.

Furthermore, adding this algorithm will allow Buildbarn to get rid of a
similar enum, which already includes Deflate.


EdSchouten:
Cc @mostynb

mostynb:
Could you describe the zip extraction use case a bit more? Does that use "compressed-blobs" bytestream transfers, or some other mechanism?

EdSchouten:
Buildbarn has a feature called [Indirect CAS, or ICAS for short](https://github.com/buildbarn/bb-adrs/blob/master/0004-icas.md) that allows you to merge external data sets into the CAS. These data sets are not ingested into the CAS; it's just that the CAS will contain internal references to the actual location of these resources. From the client's point of view, these objects are indistinguishable from regular CAS objects, as Buildbarn will simply follow these references when objects are referenced. This is ideal for adding large data sets to the build in such a way that the CAS size doesn't blow up.

Buildbarn's ICAS code already has support for decompressing objects when fetched, using [its own Protobuf enumeration](https://github.com/buildbarn/bb-storage/blob/master/pkg/proto/icas/icas.proto#L66-L73). Combined with being able to do range requests, this allows you to create ICAS entries that reference parts of ZIP files. You could, for example, store a ZIP file in S3 and let Buildbarn access individual files inside, assuming you seed the ICAS properly.

My reasoning is that if Buildbarn were to implement support for compressed blobs, the ICAS feature may be extended to simply pass blobs through if the ICAS entry specifies a compression algorithm that the client supports.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1679,6 +1679,9 @@ message Compressor {
 
     // Zstandard compression.
     ZSTD = 1;
+
+    // RFC 1951 Deflate, commonly used by ZIP files.
```
mostynb, line 1683: Since there are a few interrelated forms of deflate, I think this should be more explicit- IIUC you mean "raw deflate", correct? Is it worth saying "without zlib or gzip structure" and dropping the reference to zip?

mostynb (commented):
> Buildbarn's ICAS code already has support for decompressing objects when fetched, using its own Protobuf enumeration. Combined with being able to do range requests, this allows you to create ICAS entries that reference parts of ZIP files. You could, for example, store a ZIP file in S3 and let Buildbarn access individual files inside, assuming you seed the ICAS properly.

Just to confirm, in this scenario buildbarn's ICAS layer would act as the client for compresed-blobs bytestream reads/writes, and would store individual zip file entries in the CAS (not the entire zip file)?

So to add a blob to the cache, the ICAS layer would extract a file from the zip file, hash it to get the uncompressed file's Digest, then to a compressed-blob write with the compressed entry from the zip file?

Then to read one of these blobs corresponding to a zip file entry later, the ICAS layer would make a compressed-blobs read request with the uncompressed digest of the particular file within the zip file? (Just curious, how would the ICAS layer get the uncompressed digest? Does the ICAS layer store that info somewhere?)

build/bazel/remote/execution/v2/remote_execution.proto
EdSchouten: Isn't "RFC 1951 Deflate" already unambiguously describing what the format is?

EdSchouten:
> Just to confirm, in this scenario buildbarn's ICAS layer would act as the client for compresed-blobs bytestream reads/writes

Buildbarn's ICAS layer would act as a HTTP, S3 or [...] client that loads data from some kind of external service. Whatever your organisation uses to store data sets.

> and would store individual zip file entries in the CAS (not the entire zip file)?

The ICAS itself would contain tiny Reference messages that contain organisation specific URLs, potentially referencing ranges of ZIP files stored in the external service mentioned previously.

> So to add a blob to the cache, the ICAS layer would extract and uncompress a file from the zip file, hash it to get the uncompressed file's Digest, then do a compressed-blob write with the compressed entry from the zip file?

If you add a blob to the cache, it will always end up in the regular CAS. The ICAS is merely an overlay on top of the CAS that merges other data sets into it, so that they don't need to become part of the CAS itself.

The question then becomes: how do entries get added to the ICAS? One could write a custom implementation of the Remote Asset API (or a special REv2 worker) that is able to resolve company specific URLs (acmecorp://cat_images/fluffy.jpg), generating ICAS entries for the objects that are being requested. Because all of that is organisation specific, I'm not providing any off-the-shelf tooling for that.

[Here's an ADR that describes ICAS in more detail.](https://github.com/buildbarn/bb-adrs/blob/master/0004-icas.md)

build/bazel/remote/execution/v2/remote_execution.proto
mostynb: Technically yes, but we've had a couple of decades of `Content-Encoding: deflate` meaning something other than "plain deflate" to confuse things. Maybe it's just me though.

mostynb:
If this enum value would not be used in any "standard" REAPI implementation then I would be against adding it, for fear of new server implementations feeling the need to implement it without good cause. (In this case I might be OK with reserving a value so you could mirror the enum without fear of conflicts, but I'd have to think it over some more.)

I would be in favour of adding DEFLATE to the enum, if that would help a specific use case for another REAPI bytestream compressed-blobs CAS implementation- and I would love to see this added to bb-storage. But it's not obvious to me if that's what you're proposing.

EdSchouten:
I would be very, very interested in adding support for compressed blobs to Buildbarn. Unfortunately, there are things that first need to be resolved before we can get there. As I commented on the draft proposals for compressed blobs, I find the existing definition impractical to work with. Below are issues that I see from my side, some of which I reported back then but were not addressed, some of which I realised later on.

- The sending side does not communicate the compressed size of blobs to the receiver. This makes it hard for the receiving side to allocate space for these objects. Not an issue if you're just dumping stuff on a UNIX file system, but very impractical if you want to store objects densely, like Buildbarn's LocalBlobAccess does.
- Buildbarn has invested a lot of effort in getting a proper FUSE file system added to its workers. Even though the existing FUSE code and large parts of the storage layer have very good support for random access (partial) I/O, we can't do this at the network level, because there would be no way for us to validate the contents. SHA-256 is Merkle-Damgård based; not a Merkle tree. Whereas it was previously just the checksumming that prevented us from doing this, now we have the same issue with compression.
- For read requests, clients are requesting content using an explicit compression algorithm embedded in the bytestream path. That's impractical if you want to migrate from one compression algorithm to another, or want to use features like ICAS to merge external data sets into a CAS. You can't always assume that everything is compressed using the same algorithm. A better choice would have been to extend calls like Read() to take a list of supported compressors, allowing the server to pick one of those on a per-blob basis.

My long-term goal is to add support for compression to Buildbarn, but it will first require more changes to REv2 before we can do this. It's just a matter of time (and effort). Until we get there, support for compression will most likely be limited to Buildbarn's ICAS support. An important first step there is that we can at least agree on a common set of identifiers for compression algorithms that we use. Without it, it will be even harder for us to get there. I am going to need DEFLATE support, which is why I submitted this PR. I don't see how this differs from other parties making similar requests in the past, such as Microsoft requesting that we add VSO hashing.

mostynb:
> I would be very, very interested in adding support for compressed blobs to Buildbarn. Unfortunately, there are things that first need to be resolved before we can get there. As I commented on the draft proposals for compressed blobs, I find the existing definition impractical to work with. Below are issues that I see from my side, some of which I reported back then but were not addressed, some of which I realised later on.
> 
> * The sending side does not communicate the compressed size of blobs to the receiver. This makes it hard for the receiving side to allocate space for these objects. Not an issue if you're just dumping stuff on a UNIX file system, but very impractical if you want to store objects densely, like Buildbarn's LocalBlobAccess does.

With the current design I guess the only (potentially) easy and practical solution here is to buffer items before adding them to "permanent" storage. If the server wants to verify the blob, it has to at least decompress first anyway.

> * Buildbarn has invested a lot of effort in getting a proper FUSE file system added to its workers. Even though the existing FUSE code and large parts of the storage layer have very good support for random access (partial) I/O, we can't do this at the network level, because there would be no way for us to validate the contents. SHA-256 is Merkle-Damgård based; not a Merkle tree. Whereas it was previously just the checksumming that prevented us from doing this, now we have the same issue with compression.

If you're willing to buffer and recompress on the server, perhaps you could recompress into independent blocks to make random access IO less expensive (or use a seekable compression format)? You would still have the checksumming issue, of course.

> * For read requests, clients are requesting content using an explicit compression algorithm embedded in the bytestream path. That's impractical if you want to migrate from one compression algorithm to another, or want to use features like ICAS to merge external data sets into a CAS. You can't always assume that everything is compressed using the same algorithm. A better choice would have been to extend calls like Read() to take a list of supported compressors, allowing the server to pick one of those on a per-blob basis.

I agree with you on this. Trying to shoehorn this onto bytestream is not great though IMO. For REAPIv3 I think we should either fork bytestream or replace it with something better for our needs.

> My long-term goal is to add support for compression to Buildbarn, but it will first require more changes to REv2 before we can do this. It's just a matter of time (and effort). Until we get there, support for compression will most likely be limited to Buildbarn's ICAS support. An important first step there is that we can at least agree on a common set of identifiers for compression algorithms that we use. Without it, it will be even harder for us to get there. I am going to need DEFLATE support, which is why I submitted this PR. I don't see how this differs from other parties making similar requests in the past, such as Microsoft requesting that we add VSO hashing.

I'm not sure if there's a preexisting policy on this sort of thing, but if there is then we should probably follow it. Though we do have a couple of options for unblocking you while still discouraging deflate usage (either reserving a value, or adding a big scary warning in the comments).

EdSchouten:
> > * For read requests, clients are requesting content using an explicit compression algorithm embedded in the bytestream path. That's impractical if you want to migrate from one compression algorithm to another, or want to use features like ICAS to merge external data sets into a CAS. You can't always assume that everything is compressed using the same algorithm. A better choice would have been to extend calls like Read() to take a list of supported compressors, allowing the server to pick one of those on a per-blob basis.
> 
> I agree with you on this. Trying to shoehorn this onto bytestream is not great though IMO. For REAPIv3 I think we should either fork bytestream or replace it with something better for our needs.

To me it's not evident why this wasn't done in the first place. Because Bytestream reads and writes take different pathname schemes, it's not as if you can just implement a protocol agnostic Bytestream-based storage service that merely happens to facilitate the requirements of REv2. It needs to be REv2 aware to begin with.

Instead of `CacheCapabilities.supported_compressor`, we could have equally added a boolean option named `CacheCapabilities.supports_improved_transfer_protocol`. When set, clients could have opted to use a completely new set of RPCs to do chunked I/O, compression with bidirectional negotiation, etc.. A change like that would have been equally intrusive. If the existing compressed blobs support was eligible to be included in REv2, so would this change have been.

Anyway, all of that is water under the bridge at this point. REv2 compressed blobs support is there, and we'll just have to accept it as is. Better luck next time!

EricBurnett:
> > > * For read requests, clients are requesting content using an explicit compression algorithm embedded in the bytestream path. That's impractical if you want to migrate from one compression algorithm to another, or want to use features like ICAS to merge external data sets into a CAS. You can't always assume that everything is compressed using the same algorithm. A better choice would have been to extend calls like Read() to take a list of supported compressors, allowing the server to pick one of those on a per-blob basis.
> > 
> > 
> > I agree with you on this. Trying to shoehorn this onto bytestream is not great though IMO. For REAPIv3 I think we should either fork bytestream or replace it with something better for our needs.
> 
> To me it's not evident why this wasn't done in the first place. Because Bytestream reads and writes take different pathname schemes, it's not as if you can just implement a protocol agnostic Bytestream-based storage service that merely happens to facilitate the requirements of REv2. It needs to be REv2 aware to begin with.

W.r.t. Bytestream, yeah it turned out to not be the right choice in hindsight. We anticipated interop with more generic bytestream tooling that never panned out (and doesn't exist, as far as I can tell); we've also started adding use-cases that Bytestream gets in the way of (like compression, hints, metadata). I think everyone is in agreement that it'll be dropped for V3 :).

W.r.t. supporting one compression algorithm rather than a list, that was intentional and independent of bytestream. We're strongly segregating the concepts of "storage compression" and "transport compression". In storage, store whatever you want, including a mix of algorithms during transitions (or for your ICAS use-case). In transport, use one for the whole client session, including recompressing bytes in the server if needed to present this consistent view to the clients. This was for a few reasons:

- Recompressing is cheap enough. zstd-1 adds a small enough CPU overhead that the difference between "transport bytes to the client; decompress there" and "decompress bytes in the server; recompress with zstd for transport; decompress zstd bytes in client" is a tolerable CPU delta in total.
- Allowing a client to provide a list of supported algorithms doesn't eliminate the need for server-side recompression (the intersection might be just one algorithm anyways), so it's a worst-of-all-worlds scenario w.r.t. implementation complexity - all the same complexity on the server, plus the client now has to support multiple algorithms simultaneously.
- Forcing all clients to support all formats and provide a maximally-complete list would avoid server-side recompression, but then be a blocker for future migrations (what do you do about old clients?).

W.r.t. a new bytestream replacement that supports "chunked I/O, compression with bidirectional negotiation, etc" - please feel free to develop a proposal, though I'd suggest reviewing all the discussions around the current design before you do. Very little ended up being dictated due to bytestream itself; design tradeoffs and features that sound good on the surface but didn't survive a detailed review led to most of the features and non-features you see today, including the choice to support a single client-selected algorithm and the focus on an API that talks about the logical bytes instead of the resulting bytes. Just to give one example from up-thread, pre-informing the client of the compressed size to expect requires _knowing_ the compressed size to expect, which is incompatible with both transport/storage recompression and offset reads. My experience has been that developing the ideal generic compression API is a lot harder than it sounds, though it's certainly possible there's an even more elegant design out there that we missed.

One direction I imagine us going is some sort of chunked / stitchable / recursive format, and carrying that through all the APIs that can change to take advantage of it. But even there, not all parties' goals will be simultaneously satisfiable - e.g. variable-length chunks (for efficient chunk-level deduping and storage reduction) and fixed-length chunks (for easy verifiable offset reads) are incompatible. We'll have to see how the discussions go - I imagine Microsoft has strong opinions around things compatible with the mental model of VSO; Google has an internal recursive hash format of our own used in forge et al; goals in this thread about supporting arbitrary subset reads that are still hash-verifiable probably require something else again; etc. I do look forward to the eventual discussion, but I don't know going in that we'll be able to come up with _anything_ that all parties are happy with.

Back on topic, considering DEFLATE: I'm personally fine in adding an enum for it, though would suggest we then go and mark ZSTD as the recommended choice for all implementers and the algorithm with broadest client/server support. Anyone with a use-case for DEFLATE should not be restricted from being able to express it, but it should clearly be something you opt in to with eyes open rather than looking like an as-good option amongst equal peers.

tdyas:
> My experience has been that developing the ideal generic compression API is a lot harder than it sounds, though it's certainly possible there's an even more elegant design out there that we missed. . . . But even there, not all parties' goals will be simultaneously satisfiable - e.g. variable-length chunks (for efficient chunk-level deduping and storage reduction) and fixed-length chunks (for easy verifiable offset reads) are incompatible. 

Tangential to this PR, but related to v3: Has anyone been collecting all known REAPI use cases in one document? If not, REAPI Working Group should do that. It would help with v3 discussions for all parties involved to be aware of each other's use cases. (I can help coordinate this as well.)

EricBurnett:
> Tangential to this PR, but related to v3: Has anyone been collecting all known REAPI use cases in one document? If not, REAPI Working Group should do that. It would help with v3 discussions for all parties involved to be aware of each other's use cases. (I can help coordinate this as well.)

I'm not aware of anything. Known implementors are listed [here](https://github.com/bazelbuild/remote-apis), and specific use-cases are enumerated/considered when evaluating any given change proposal, but I don't think anyone has tried to do a holistic evaluation of all use-cases for all features.

build/bazel/remote/execution/v2/remote_execution.proto
EdSchouten: I've extended the description. At the same time, I've made it more explicit that this algorithm is not the recommended one.

bergsieker (approved):

mostynb (approved):

EdSchouten:
Hey! Is this good to land?

EdSchouten:
Rebased! PTAL!

EricBurnett:
I don't normally use my push permissions so apologies to everyone if I've missed some step, but you've been waiting on this too long already. Merging.

PR #195, opened 2021-05-10
update rules_go

atetubou:
This is to use protobuf v2 library for Go.


atetubou:
cc @EricBurnett

EdSchouten:
Could we also use this occasion to migrate to Gazelle’s new way of naming targets? We still declare some go_default_library targets.

atetubou:
Which way are you supposed to migrate? I currently feel that there is no established way to handle generated *.pb.go for bazel.

Also I'd like to focus only to update bazel rules in this PR.

atetubou:
Thanks, but we use this via https://github.com/bazelbuild/remote-apis-sdks. So it is better to have *.pb.go eventually there?

PR #200, opened 2021-07-08
add BUILD rule for python proto

atetubou:
This is for http://shortn/_F3JVJ8vcuC


EdSchouten:
What does the shortened URL point to? I can't open it.

atetubou:
Sorry, that URL points to google internal issue of our customer which we need to support their adoption to RBE-CAS.

bergsieker:
In general we'd prefer not to have language-specific BUILD targets in this repository. (We have a pending workitem to move the Go targets elsewhere.) The API itself is both language- and version-agnostic, and it's not possible to meet all the constraints of users who want to use API version/commit X but language/toolchain version Y. The API maintainers also generally lack sufficient expertise and setups to test and validate PRs that touch languages they don't normally use.

Is it possible to simply have these targets in the client repo? If that's not possible, would remote-apis-sdks be a more suitable destination?

atetubou:
I asked to have grpc_proto rule in our customer's BUILD file like http://shortn/_cSxVu7yKXO, but that seems not possible due to http://shortn/_tdAMa97jjL.

This doesn't allow to have the rule in client repo too.

bergsieker:
Interesting. That restriction in py_grpc_library.bzl seems like a strange one to enforce. I wonder what happens if you try commenting that out and generating the library? Perhaps we can request that the restriction be removed.

If this is for internal-only usage, we could consider adding the targets to the internal copy of this proto file. I view that as roughly equivalent to putting a similar library rule in remote-apis-sdks--it's essentially separating the abstract API definition (in this repo) from the practical code that is being compiled and used.

atetubou:
Discussed with bergsieker, we decied to have this only in our internal copy.

Issue #201, opened 2021-07-12
Compression support for inlined data (v2-compatible)


mostynb:
We added support for compressed blobs via the bytestream API a while back (#147), and this has been working well so far. Given that we don't seem to be gearing up for a non-backwards compatible REAPIv3 release, I think we should start investigating adding backwards-compatible support for compression of inlined data, aiming to minimize roundtrips while still using compressed data.

One idea that I have been thinking about is to create a new `Blob` message which has the same fields as `Digest`, but with two additional optional fields: `bytes data` to hold the inlined data and `Compressor.Value compressor` to describe the encoding. Then, we would replace each usage of `Digest` in a message that currently also has a separate field for inlined (uncompressed) data with a `Blob` field, and deprecate the previous data field.

For example:
````
message Blob {
  // The first two field types + numbers must match those in Digest
  string hash = 1;
  int64 size_bytes = 2; // This refers to the uncompressed size

  reserved 3, 4; // Leave some room in case we want to extend Digest later(?)

  bytes data = 5; // Possibly compressed data, if set
  Compressor.Value compressor = 6; // Encoding used, if the data field is set
}
````

For "request" messages that can be responded to with inlined data, we would add a repeated field that specifies the compression types that the client accepts. The server would pick one of those encodings (or identity/uncompressed) for each Blob that it chooses to inline data for. We would also add a new field in the Capabilities API for servers to advertise support for this.

I believe this would be backwards-compatible- old clients would not request compressed inlined data, and would receive `Blobs` that are decodable as `Digests` in response, and servers would have a clear signal for when this feature can be used in responses.

Is anyone else interested in this feature?

EricBurnett:
Heya Mostyn,

We'd be interested in compressed inlined blobs too, yep! We have
encountered some builds where it's relevant ourselves, with the batch APIs
moving a non-trivial percent of the total bytes. (Builds particularly heavy
in small files).

For implementation details, what's the benefit of introducing a Blob
message to wrap these fields vs just adding compressor where necessary?
Specifically, I could see three main variants:

   1. Just stick a `compressor` field beside `bytes` where relevant.
   (BatchUpdateBlobsRequest, BatchReadBlobsResponse, OutputFile, ActionResult
   x2 for std*_raw).
   2. Use `Blob` to wrap `bytes` and `compressor`, but *not* the Digest
   fields.
   3. Encapsulate Blob as a replacement for Digest, as you've described.

[...]

PR #202, opened 2021-07-13
Add support for inlined compressed data (minimal changes)

mostynb:
This is a small API change which allows for inlined data to be compressed.

Refers to #201.


build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1412,6 +1429,11 @@ message GetActionResultRequest {
   // `output_files` (DEPRECATED since v2.1) in the
   // [Command][build.bazel.remote.execution.v2.Command] message.
   repeated string inline_output_files = 5;
+
+  // A list of acceptable encodings to used for inlined data.  Must be
+  // `IDENTITY` (or unspecified), or one or more of the compressors
+  // supported by the server.
```
EdSchouten, line 1435: I think that this phrasing is far too strict. Why does it need to contain compressors supported by the server? Couldn't the server just ignore the ones it doesn't support?

Furthermore, what's the point in making IDENTITY explicit in this list? I would do it the other way around: forbid IDENTITY from being listed explicitly, allowing servers to always pick this compressor for responses.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1724,6 +1760,11 @@ message CacheCapabilities {
   // Note that this does not imply which if any compressors are supported by
   // the server at the gRPC level.
   repeated Compressor.Value supported_compressor = 6;
+
+  // Whether compression of inlined data is also supported (if so, the
+  // server must support each of the compressors specified in the
+  // `supported_compressor` field).
+  bool inlined_compressed_blobs = 7;
```
EdSchouten, line 1767: Relatedly, I also think that this phrasing far too strict. With the repeated field that you're adding, you've already introduced a two-way method for negotiating compression algorithms on a per-request basis.

Do we even need this flag? Why not always allow clients to just provide a list of compression functions, regardless of whether the server does anything with it?

build/bazel/remote/execution/v2/remote_execution.proto
mostynb: s/Must/Should/ sounds wise. Note this is just a quick first pass, for comparison with #203.

If a client specifies `IDENTITY` here, the server knows it's not an old client, and can inline uncompressed data in the Blob instead of in the legacy field.

EdSchouten: But this PR doesn't contain Blob.

build/bazel/remote/execution/v2/remote_execution.proto
mostynb: It is possible to upload blobs with bytestream in parallel, but it's often simpler to do it in a single RPC call, and batching small requests together can reduce certain overheads (it's a balancing act as always).

build/bazel/remote/execution/v2/remote_execution.proto
mostynb: You're right, this is probably something that can be dropped from this option.

build/bazel/remote/execution/v2/remote_execution.proto
EdSchouten: Have you got numbers on what the overhead is?

mostynb: I don't have any useful numbers from my own setup to share (I'm currently using gRPC level compression for these unary RPCS with inlined data, though I would prefer to migrate to REAPI-level compression).

EricBurnett: For most use-cases it doesn't matter, but when uploading lots of small blobs it can be very material. With e.g. 1kb blobs, you're likely to be parallelism bound (100 RPCs in flight / connection -> 1000 blobs/second/connection or so); batch uploads are then very necessary if you've got thousands of blobs to upload.

For the field comment: should this apply to _all_ inline data, or just be scoped to client-sent data? I.e. I'm wondering about changing it to "Whether compression of client-sent inlined data is also supported". That'd allow a server to support returning compressed data (via client-specified acceptable_compressors) without having to support accepting it.

Though...maybe that's a bad thing, and we should keep the existing semantics so as to introduce less variation in servers.

EricBurnett (commented):
LGTM in general - two outstanding comment threads to resolve, but I expect to be fine with whatever exact text you settle on, nothing controversial there.

build/bazel/remote/execution/v2/remote_execution.proto
mostynb: Updated the comment. This only applies to client-sent data since we have allowed the server to always return uncompressed data if it chooses to.  It would be toothless to state here that servers must also support this for downloads.

Jumping back to @EdSchouten's question...
> What's the use of uploading inlined blobs?

This (small) PR doesn't add support for inlining blobs in new places. It just allows compressed inlined data where we previously only allowed uncompressed inlined data. Which IMO is a net benefit for servers that choose to support inlined data for uploads.

EdSchouten: > This (small) PR doesn't add support for inlining blobs in new places. It just allows compressed inlined data where we previously only allowed uncompressed inlined data.

I get that. But there's also protocol evolution, right? Does it make sense to extend features that would in retrospect be considered redundant/unnecessary? If we're adding support for compressed uploads in BatchUploadBlobs(), does it even make sense to invest in extending UpdateActionResult() to support it?

mostynb: IMO this is a useful improvement for UpdateActionResult rather than something we would want to discourage. On the server side it is probably not much harder to implement this for both messages than for only BatchUpdateBlobs.

I wouldn't be opposed to advertising the capabilites of each rpc call separately, but I'm not sure if that would help any server imlementations in practice.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1280,6 +1292,11 @@ message ExecuteRequest {
   // The server will have a default policy if this is not provided.
   // This may be applied to both the ActionResult and the associated blobs.
   ResultsCachePolicy results_cache_policy = 8;
+
+  // A list of acceptable encodings to used for inlined data.  Must be
+  // `IDENTITY` (or unspecified), or one or more of the compressors
+  // supported by the server.
+  Compressor.Value compressor = 9;
```
mostynb, line 1299: This should be a repeated field, named "acceptable_compressors" for consistency.

build/bazel/remote/execution/v2/remote_execution.proto
EricBurnett: Upon re-reading the API, I'm actually leaning towards Ed's suggestion of prohibiting compressed blob inlining in UpdateActionResult. While it's hopefully not *that* hard, it is a new set of fields and whatnot to support compression on (clients don't *set* stdout/stderr fields anywhere else, output contents, etc; Servers don't usually read it), and allowing it would force servers to support rewriting compressed fields on the way out (e.g. if one client uploads an action result with inlined compressed data, and a second client calls GetActionResult and does not specify compression being allowed.)

Of course, if anyone actually cares to have support of compressed+inlined uploads, that's a different story entirely. But right now it sounds like it's unmotivated by concrete use-cases, and mostly desirable for symmetry? Which I care about as well, though perhaps the tax in this case outweighs it.

mostynb: Don't servers already need to support modifying update ActionResult messages though? There's nothing preventing clients and workers from uploading ActionResult messages with inlined uncompressed data, and servers must not inline this data unless requested by the client in response to GetActionResult messages.

The strategy I implemented in bazel-remote was to accept ActionResult messages with possibly inlined data, store any inlined blobs separately in the CAS since they may be requested separately, de-inline them from the ActionResult before storing that in the ActionCache, and then returning that possibly modified Actionresult.

This allows clients and workers to upload the results of executing an action in a single RPC call in the best case. If we limit compressed inlined blob support to BatchUpdateBlobs for uploads, then the client/worker needs to make at least two RPC calls if you want to use compression (but not gRPC level compression). Even if they're not sequential it's simpler to do this in a single RPC.

EdSchouten:
I can no longer respond to the thread above, so here it goes:

> Don't servers already need to support modifying update ActionResult messages though? There's nothing preventing clients and workers from uploading ActionResult messages with inlined uncompressed data, and servers must not inline this data unless requested by the client in response to GetActionResult messages.

The spec is a bit incomplete here. There is nothing in the spec that states that clients even MAY/SHOULD/${VERB} inline data as part of GetActionResult. For example, take the description of stdout_raw:

```proto
  // The standard output buffer of the action. The server SHOULD NOT inline
  // stdout unless requested by the client in the
  // [GetActionResultRequest][build.bazel.remote.execution.v2.GetActionResultRequest]
  bytes stdout_raw = 5;
```

Notice how the entire paragraph discusses the semantics of GetActionResult, but there is not a single mention of "upload" or "UpdateActionResult". There are also boolean flags as part of GetActionResultRequest to control what may be inlined, but there are no corresponding boolean flags in Capabilities that control this feature in the upload case.

This raises the question: was inlining stdout/stderr as part of UpdateActionResult even considered in the first place? Buildbarn, for example, tolerates it, but then goes ahead and stores the ActionResult in literal form. This means that if a client later on requests the ActionResult with inline_stdout/stderr set to false, Buildbarn goes out of spec: it will return stdout/stderr inline.

As far as I know, you are interested in the remote *caching* case; not remote *execution*. Let's take a step back and look at what you're trying to achieve. Consider this workflow:

- Client calls GetActionResult(). It returns a negative response.
- Client builds the action locally.
- Client calls FindMissingBlobs() on all output files.
- Outputs are absent, but small, so client calls BatchUpdateBlobs().
- Client calls UpdateActionResult() with stdout and stderr inlined.

So that's at least three RPCs: GetActionResult(), FindMissingBlobs() and UpdateActionResult(). At most, it's four RPCs: all of the previous ones and BatchUpdateBlobs(), in case one or more outputs were missing.

My proposal is to not add support for compressed uploads to UpdateActionResult(). Instead, you make the stdout and stderr part of the FindMissingBlobs() and BatchUpdateBlobs() calls and provide the digests to UpdateActionResult(). This keeps the protocol lean and immediately acts as a soft-deprecation of inline uploads.

The only downside of this approach is that you may need to perform four RPCs more frequently, namely in the case where stdout or stderr were reported missing, but all other outputs were present. My assumption is that this only happens rarely. Most actions don't even generate stdout/stderr. And even if this is not rare, there is so much more opportunity here. With very little effort, we can extend the Remote Execution protocol to *always* allow you to complete such actions with three RPCs. Here's the diff (lacking capabilities, etc.):

```
diff --git a/build/bazel/remote/execution/v2/remote_execution.proto b/build/bazel/remote/execution/v2/remote_execution.proto
index 9b58064..588d94a 100644
--- a/build/bazel/remote/execution/v2/remote_execution.proto
+++ b/build/bazel/remote/execution/v2/remote_execution.proto
@@ -1463,7 +1463,7 @@ message FindMissingBlobsResponse {
 // [ContentAddressableStorage.BatchUpdateBlobs][build.bazel.remote.execution.v2.ContentAddressableStorage.BatchUpdateBlobs].
 message BatchUpdateBlobsRequest {
   // A request corresponding to a single blob that the client wants to upload.
-  message Request {
+  message ContentAddressableStorageRequests {
     // The digest of the blob. This MUST be the digest of `data`.
     Digest digest = 1;
 
@@ -1479,7 +1479,15 @@ message BatchUpdateBlobsRequest {
   string instance_name = 1;
 
   // The individual upload requests.
-  repeated Request requests = 2;
+  repeated ContentAddressableStorageRequests content_addressable_storage_requests = 2;
+
+  message ActionCacheRequests {
+    Digest action_digest = 1;
+    ActionResult action_result = 2;
+  }
+
+  repeated ActionCacheRequests action_cache_requests = 3;
+  ResultsCachePolicy results_cache_policy = 4;
 }
 
 // A response message for
```

Another bonus of this change is that it adds batching support to the AC. Extend BatchReadBlobs() similarly and you will cut down the number of RPCs tremendously. Once done, we can soft-deprecate GetActionResult() and UpdateActionResult() as well.

So from a Pareto point of view, it's absolutely not evident to me why you want to add support for inlined compressed data to UpdateActionResult().

mostynb:
I'm working both on cache-only and remote execution scenarios (execution workers have a similar process to cache clients). I have a few cases where I use a specialised cache client for actions whose outputs are reusable, but the outputs are very unreproducible, so it's not worth calling FindMissingBlobs. In the best case (when all the data fits inline) it's 1 RPC call for a cache hit (GetActionResult), and 2 RPC calls for a cache miss (GetActionResult, UpdateActionResult). If the data doesn't all fit, then the client has to make more RPC calls to transfer the leftover blobs (hence why adding support for compression here would be useful to me, at least).

I guess I have been using an underspecified corner of the spec then, clients/workers inlining data where it's possible and not forbidden. Should we continue to allow that (and possibly add support for compressed data), or recommend against it? Or prohibit it?

@EdSchouten: do you want to open another PR for your idea? It's a larger change, but I think it would cover my usecases.

EdSchouten:
> I guess I have been using an underspecified corner of the spec then, clients/workers inlining data where it's possible and not forbidden. Should we continue to allow that (and possibly add support for compressed data), or recommend against it? Or prohibit it?

I'm personally fine with prohibiting it. My only concern is that someone else will show up within the next couple of days and call me crazy. In case that happens, I would at least suggest that we recommend against it. @EricBurnett, what are your thoughts?

> @EdSchouten: do you want to open another PR for your idea? It's a larger change, but I think it would cover my usecases.

I'd say, let's first work towards adding compression support to BatchUpdateBlobs() as part of this PR, while leaving UpdateActionResult() as is. Once that lands, merging the functionality of the latter into the former is a logical extension.

mostynb:
Filed #206 to consider client-inlining separately, so it isn't lost in this discussion.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1724,6 +1757,13 @@ message CacheCapabilities {
   // Note that this does not imply which if any compressors are supported by
   // the server at the gRPC level.
   repeated Compressor.Value supported_compressors = 6;
+
+  // Compressors supported for inlined data in messages uploaded to the
+  // server like
+  // [BatchUpdateBlobsRequest][build.bazel.remote.execution.v2.BatchUpdateBlobsRequest],
+  // [ActionResult][build.bazel.remote.execution.v2.ActionResult] and
+  // [OutputFile][build.bazel.remote.execution.v2.OutputFile].
+  repeated Compressor.Value supported_inline_compressors = 7;
```
EdSchouten, line 1766: supported_upload_compressors?
Considering that we're no longer doing inlining for uploads

mostynb: I think it might be better to be more specific than `supported_upload_compressors`. How about `supported_batch_compressors` ? Even though it's only strictly required to know in order to use compressed data in `BatchUpdateBlobsRequest`, the same values will probably be the ones selected from for `BatchReadBlobsResponse`.

And what do you think about renaming `supported_compressors` to `supported_bytestream_compressors` ?

mostynb:
Rebased this, and limited it to the batch APIs.

mostynb:
Here's a sample implementation for bazel-remote: https://github.com/buchgr/bazel-remote/pull/459

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1510,6 +1510,11 @@ message BatchUpdateBlobsRequest {
 
     // The raw binary data.
     bytes data = 2;
+
+    // The format of `data`. Must be `IDENTITY` (or unspecified), or one of the
+    // compressors advertised by the `CacheCapabilities.supported_compressor` if
```
EricBurnett, line 1515: supported_compressor -> supported_batch_compressors , and I think inlined_compressed_blobs is no longer a field?

EricBurnett (approved):
LGTM - one stale comment to correct, but otherwise looks good; the main outstanding point about inline uploaded data has been mooted by #206.

I'll wait till Ed signs off (and/or has additional comments) before merging this though.

build/bazel/remote/execution/v2/remote_execution.proto
mostynb: Fixed.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1765,6 +1779,13 @@ message CacheCapabilities {
   // Note that this does not imply which if any compressors are supported by
   // the server at the gRPC level.
   repeated Compressor.Value supported_compressors = 6;
+
+  // Compressors supported for inlined data in
+  // [BatchUpdateBlobs][build.bazel.remote.execution.v2.ContentAddressableStorage.BatchUpdateBlobs]
+  // and
+  // [BatchReadBlobs][build.bazel.remote.execution.v2.ContentAddressableStorage.BatchReadBlobs]
```
EdSchouten, line 1786: To me it's still not clear why we want to mention BatchReadBlobs here. In the case of batch reading, this feature is already self-negotiating.

mostynb: BatchReadBlobs has basic negotiation- the client will never get blobs in a compressed encoding that it didn't ask for. But by itself it doesn't allow the client to prioritise compressors.

My thought here was that if the server advertises the supported compressors ahead of time, then the client can pick its favourite from the advertised list and make a request for that. I suspect that most servers which implement this will support the same set of compressors for BatchReadBlobs and BatchUpdateBlobs, and we could therefore use the same field to cover both. Maybe it would be better to use separate fields though.

Alternatively, maybe we could just state that `BatchReadBlobsRequest.acceptable_compressors` are specified in the client's order of preference?

EdSchouten: I’m a fan of the latter.

mostynb: WFM- done. Renamed the capabilities field to `supported_batch_update_compressors`.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1765,6 +1779,11 @@ message CacheCapabilities {
   // Note that this does not imply which if any compressors are supported by
   // the server at the gRPC level.
   repeated Compressor.Value supported_compressors = 6;
+
+  // Compressors supported for inlined data in
+  // [BatchUpdateBlobs][build.bazel.remote.execution.v2.ContentAddressableStorage.BatchUpdateBlobs]
+  // requests.
+  repeated Compressor.Value supported_batch_update_compressors = 7;
```
EdSchouten, line 1786: Maybe mention whether this is in order of preference, or explicitly why not?

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1551,6 +1557,10 @@ message BatchReadBlobsRequest {
 
   // The individual blob digests.
   repeated Digest digests = 2;
+
+  // A list of acceptable encodings to used for inlined data, in order of
```
EdSchouten, line 1561: In decreasing order

mostynb: How does "in order of preference, from most to least preferred" sound?

EdSchouten: SGTM!

mostynb: Updated.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1510,6 +1510,12 @@ message BatchUpdateBlobsRequest {
 
     // The raw binary data.
     bytes data = 2;
+
+    // The format of `data`. Must be `IDENTITY` (or unspecified), or one of the
```
EdSchouten, line 1514: Nit: ProtoV3 doesn't differentiate between `IDENTITY` and unspecified. That '(or unspecified)' part could be dropped.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1551,6 +1557,11 @@ message BatchReadBlobsRequest {
 
   // The individual blob digests.
   repeated Digest digests = 2;
+
+  // A list of acceptable encodings to used for inlined data, in order of
+  // preference from most to least preferred. `IDENTITY` is always allowed
+  // even if not specified here.
```
EdSchouten, line 1563: So based on this phrasing, if I say this:
```json
["ZSTD", "DEFLATE", "IDENTITY"]
```
Then that would be the same thing as this:
```json
["ZSTD", "DEFLATE"]
```
That makes sense. But what about this?
```json
["ZSTD", "IDENTITY", "DEFLATE"]
```
Considering that this list is in order of preference, and that `IDENTITY` is always allowed, then would that be the same as the client providing this?
```json
["ZSTD"]
```
I think we should explicitly forbid `IDENTITY` from being part of this. It only leads to confusion.

build/bazel/remote/execution/v2/remote_execution.proto
mostynb: It could, but I don't think it hurts to mention this.

build/bazel/remote/execution/v2/remote_execution.proto
mostynb: The server is not required to select the client's highest supported priority, so `["ZSTD", "IDENTITY", "DEFLATE"]` is not the same as `["ZSTD"]`.

EdSchouten: Is that documented anywhere?

mostynb: Not currently. 

How about replacing the last sentence with: "Servers MAY select any of the specified compressors or IDENTITY (even if it is not specified by the client), regardless of the client's order of preference."

mostynb: I suspect there's no perfect algorithm here, and that we will need to live with some level of weakness. The client has the advantage of knowing the formats the server supports (either advertised by the capabilities API or by probing), and can try to force the server to pick the client's preference by making a request that only allows that format (or fall back to uncompressed data).

That's not great for server admins, because the server potentially has more information that it could exploit- eg it already has the blob in a particular encoding, or it knows that a particular encoding is more efficient, or it knows that CPU usage is high and doesn't want to reencode right now.

In practice I expect most implementations to only support a few compression formats, and that should limit the problem somewhat.

mostynb: Based on the discussion in the meeting today, updated to:
"A list of acceptable encodings for the returned inlined data, in no particular order. `IDENTITY` is always allowed even if not specified here."

mostynb:
@EdSchouten: how does this look now?

build/bazel/remote/execution/v2/remote_execution.proto
EdSchouten: Maybe remove the word 'or' then?

> Must be `IDENTITY` (unspecified), or [...]

EdSchouten (approved):

build/bazel/remote/execution/v2/remote_execution.proto
mostynb: How about `IDENTITY`/unspecified ?

EdSchouten: Sure! As long as we don't phrase it in a way that could be interpreted as being two different representations.

mostynb: Done.

PR #203, opened 2021-07-13
Add support for inlined compressed data (larger refactoring)

mostynb:
By defining a Blob message that overlaps with Digest, we can repurpose Digest fields to allow data inlining with optional compression in more places.  The Digest message is then only used in cases where the receiver wants a reference rather than data.

In this example I have changed most of the Digest fields to Blobs, which may lead to over-inlining of data.  Each case needs to be considered separately before this change is ready for consideration.

Refers to #201.

### DIFF
--- a/build/bazel/remote/execution/v2/remote_execution.proto
+++ b/build/bazel/remote/execution/v2/remote_execution.proto
@@ -452,10 +452,10 @@ service Capabilities {
 // reproducible so that serving a result from cache is always desirable and
 // correct.
 message Action {
-  // The digest of the [Command][build.bazel.remote.execution.v2.Command]
+  // The digest and optionally content of the [Command][build.bazel.remote.execution.v2.Command]
   // to run, which MUST be present in the
   // [ContentAddressableStorage][build.bazel.remote.execution.v2.ContentAddressableStorage].
-  Digest command_digest = 1;
+  Blob command = 1;
 
   // The digest of the root
   // [Directory][build.bazel.remote.execution.v2.Directory] for the input
@@ -464,7 +464,7 @@ message Action {
   // directory, as well as every subdirectory and content blob referred to, MUST
   // be in the
   // [ContentAddressableStorage][build.bazel.remote.execution.v2.ContentAddressableStorage].
-  Digest input_root_digest = 2;
+  Blob input_root = 2;
 
   reserved 3 to 5; // Used for fields moved to [Command][build.bazel.remote.execution.v2.Command].
 
@@ -828,8 +828,8 @@ message FileNode {
   // The name of the file.
   string name = 1;
 
-  // The digest of the file's content.
-  Digest digest = 2;
+  // The file's Digest and optionally content.
+  Blob blob = 2;
 
   reserved 3; // Reserved to ensure wire-compatibility with `OutputFile`.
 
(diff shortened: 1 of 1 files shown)


build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -452,10 +452,10 @@ service Capabilities {
 // reproducible so that serving a result from cache is always desirable and
 // correct.
 message Action {
-  // The digest of the [Command][build.bazel.remote.execution.v2.Command]
+  // The digest and optionally content of the [Command][build.bazel.remote.execution.v2.Command]
   // to run, which MUST be present in the
   // [ContentAddressableStorage][build.bazel.remote.execution.v2.ContentAddressableStorage].
-  Digest command_digest = 1;
+  Blob command = 1;
```
EdSchouten, line 458: I strongly object to this change.

With the platform properties being promoted from Command to Action, scheduler processes may nowadays be implemented in such a way that they don't need to load the Command message from the CAS. This effectively reverts that change.

The Command message can get big really quickly, especially in the case of projects where people use (strip_)include_prefix too much.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -848,11 +848,11 @@ message DirectoryNode {
   // The name of the directory.
   string name = 1;
 
-  // The digest of the
+  // The digest and optionally content of the
   // [Directory][build.bazel.remote.execution.v2.Directory] object
   // represented. See [Digest][build.bazel.remote.execution.v2.Digest]
   // for information about how to take the digest of a proto message.
-  Digest digest = 2;
+  Blob blob = 2;
```
EdSchouten, line 855: Is it permitted to also use this construct inside of Tree objects? If not, should it be stated? If so, any guidelines on canonicalisation?

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -916,6 +916,23 @@ message Digest {
...
+  // This is a hack to allow repurposing Digest fields, which we can
+  // replace with Blobs and old clients will ignore the newly added
+  // fields (which should be unused by old clients anyway).
+  //
+  // For V3, we might consider embedding a Digest inside Blob, but
+  // that would require deprecating more fields.
+  string hash = 1;
+  int64 size_bytes = 2; // This refers to the uncompressed size
+
+  reserved 3, 4; // Leave some room in case we want to extend Digest later(?)
```
EdSchouten, line 930: If we add more fields to Digest, then clients/servers will already update to a newer version of the .proto. In that case they will already switch from Digest to Blob.

In other words, I don't see how reserving numbers here helps us.

build/bazel/remote/execution/v2/remote_execution.proto
EdSchouten: To add to that: is it even likely that inlining Command into Action is advantageous? You only upload Action and Command into the CAS in case of remote execution; not remote caching. Most remote execution services hopefully have their workers located close to storage. Do we really care about reducing latency/roundtrips there?

mostynb: Sounds reasonable- this is the "over-inlining" problem I mentioned in the description (this is only intended as an early draft for discussion).

build/bazel/remote/execution/v2/remote_execution.proto
mostynb: You might be right. If we update Digest later, we could reserve fields in Digest to "jump over" the extra fields in Blob.

EdSchouten: If we update Digest later, it's completely safe to use any numbers that overlap with Blob.

mostynb:
I'm going to close this draft PR since there is more interest in #202.

PR #204, opened 2021-07-13
Make it explicit that the timeout may not be in real time

EdSchouten:
Buildbarn's FUSE file system is capable of lazily loading the input root
of an action as it's being executed. Because this adds a
non-deterministic amount of noise to execution times, there is an option
for automatically compensating the execution timeout of time spent
reading data from the CAS.

Let's extend the spec to explicitly allow this.


build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -484,6 +484,12 @@ message Action {
   // timeout is too short. By encoding it directly in the `Action`, a lower
   // timeout will result in a cache miss and the execution timeout will fail
   // immediately, rather than whenever the cache entry gets evicted.
+  //
+  // The timeout does not necessarily correspond to elapsed real time. A
```
EricBurnett, line 488: It seems we didn't really detail the purpose of the timeout at all. How about expanding the start of this paragraph to cover it at the same time, e.g.

```
The timeout is only intended to cover the "execution" of the specified action and not time in queue
nor any overheads before or after execution such as marshalling inputs/outputs. The server
SHOULD avoid including time spent the client doesn't have control over, and MAY extend the
timeout to account for delays that occur during execution itself (e.g., lazily loading data from the
Content Addressable Storage, live migration of virtual machines, emulation overhead).
```

?

EdSchouten: That sounds good!

That said, I just realised that there is also another small feature I'd like to add as part of this PR. Now that workers get a lot of freedom in applying the timeout, it would also be useful if the amount of time as measured by the worker is provided as part of the results. We currently only offer a start and end time, but those can no longer be used to compute an execution duration that is usable as a timeout value.

What are your thoughts on the latest commit I just pushed?

EricBurnett: Generally, makes sense to add a field for this as well, so SGTM. I'd personally prefer the semantics be listed on the `timeout` field and have the `execution_duration` reference it, rather than the other way around, since `timeout` is the input parameter clients should be sending whereas `execution_duration` is the optional output metadata. But that's mostly about the place the comment should live, semantics make sense either way :).

Also, I'd still like to see the timeout comment expanded to cover the purpose overall, as per my previous comment.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -473,7 +473,10 @@ message Action {
   // as long as the server will let it. The server SHOULD impose a timeout if
   // the client does not specify one, however, if the client does specify a
   // timeout that is longer than the server's maximum timeout, the server MUST
-  // reject the request.
+  // reject the request. The method of timekeeping used to apply the timeout
+  // MUST be consistent with what is used to compute
```
EricBurnett, line 477: MUST -> SHOULD? I'm not sure what material harm it would cause if a server is inconsistent here. E.g. if a server applied hidden padding to the timeout in some cases because it's expecting unknown delays (+100ms to all timeouts in case a live migration happens, say), but reports the simple delta to users, that seems fine enough to me?

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -942,6 +945,17 @@ message ExecutedActionMetadata {
...
 
+  // The amount of time the worker spent executing the action command,
+  // not including overheads before or after execution such as
+  // marshalling inputs/outputs.
+  //
+  // The server SHOULD avoid including time spent the client doesn't
+  // have control over, and MAY reduce the execution duration to account
+  // for delays that occur during execution itself (e.g., lazily loading
+  // data from the Content Addressable Storage, live migration of
+  // virtual machines, emulation overhead).
```
EricBurnett, line 956: Maybe also add something to the effect of `SHOULD be <= execution_start_timestamp - execution_completed_timestamp.` ?

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -942,6 +945,17 @@ message ExecutedActionMetadata {
...
+  google.protobuf.Duration execution_duration = 12;
```
EricBurnett, line 957: Do you think we need "new in v2.3" to indicate this is a new metadata field? Or do we think that all metadata being optional suffices to save us here - no caller should ever depend on this being populated, so when it was introduced isn't semantically meaningful?

(Having typed this, we're probably fine to leave it off. But I'll still send this comment to get a second opinion :))

build/bazel/remote/execution/v2/remote_execution.proto
EdSchouten: Buildbarn supports a feedback-driven analysis process to automatically determine the optimal worker size for an action:

https://www.youtube.com/watch?v=3eKVBwlAHsk

One of the things it wants to prevent is causing excessive stalls in case an action behaves exceptionally poorly on a worker that is too small. For example, an action that succeeds in 10 seconds on a 4-CPU system, but takes more than 5 minutes on a 1-CPU system. To solve that, it uses samples from previous executions to derive a safe execution timeout. For this to work correctly, the method of timekeeping must be consistent.

Right now this feature is integrated into the scheduler, so in a certain way an implementation detail of the system. That said, there is no reason why certain forms of analysis can't be done on the client side if you wanted to. That's why I picked the word MUST.

build/bazel/remote/execution/v2/remote_execution.proto
EdSchouten: We could, but is this something we want to limit explicitly? We're mentioning the term 'emulation' in the description, so actually executing something faster than wall time is not impossible, especially for low-end embedded targets.

build/bazel/remote/execution/v2/remote_execution.proto
EdSchouten: Seems like a good idea to do that. Added!

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -942,6 +950,25 @@ message ExecutedActionMetadata {
...
+  // including time spent the client doesn't have control over, and MAY extend
+  // or reduce the execution duration to account for delays or speedups that
+  // occur during execution itself (e.g., lazily loading data from the Content
+  // Addressable Storage, live migration of virtual machines, emulation
+  // overhead).
+  //
+  // The method of timekeeping used to compute the execution duration MUST be
+  // consistent with what is used to enforce the
+  // [Action][[build.bazel.remote.execution.v2.Action]'s `timeout`. There is no
+  // arithmetical relationship between the execution duration and the values of
```
mostynb, line 968: nitpick: the last sentence is a bit difficult to read. How about something like "The difference between execution_start_timestamp and execution_completed_timestamp may be greater than the specified timeout, to account for non-execution time." ?

EdSchouten: But it may also be less. There simply is no relationship.

mostynb: Right- you're pointing out that it can also be greater due to overheads/non-execution times (and that's the part to emphasise).

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -942,6 +950,25 @@ message ExecutedActionMetadata {
...
+  // `execution_start_timestamp` and `execution_completed_timestamp`.
+  google.protobuf.Duration execution_duration = 12;
```
EdSchouten, line 970: TODO: virtual_execution_duration

build/bazel/remote/execution/v2/remote_execution.proto
EricBurnett: Right, makes sense.

EricBurnett:
LGTM from me, thanks Ed. I'll leave it a little longer to allow for further comments from Mostyn or others, then merge it if there are no objections.

build/bazel/remote/execution/v2/remote_execution.proto
mostynb: Would s/arithmetical/specific/ do?

mostynb (approved):
Apart from a one-word readability nitpick, this is fine with me.

build/bazel/remote/execution/v2/remote_execution.proto
sstriker: Or even just s/arithmetical// ?

sstriker (approved):

build/bazel/remote/execution/v2/remote_execution.proto
EdSchouten: Removed arithmetical

PR #205, opened 2021-07-18
Pluralize repeated field names

EdSchouten:
The Protobuf style guide is pretty explicit about this: repeated fields
should use pluralized names:

https://developers.google.com/protocol-buffers/docs/style


mostynb (approved):

bergsieker (approved):

EricBurnett:
We don't have any experience I can recall with renaming fields after they're in. Are there any languages for which this will cause problems, does anyone know, beyond the general annoyance of needing to update code in lockstep with importing a new version of these protos?

There's also a possible issue if anyone is using pbtxt or json-serializing these messages, but I'd guess (hope?) that's likely not the case beyond unit tests for these specific messages.

Also, GitHub won't let me comment on the line but `ExecutionCapabilities.digest_function` should be kept consistent with `CacheCapabilities.digest_function[s]`.

Ed, what's the motivation for this change? I'm wondering how much value this will deliver, relative to the hassle for folk. Maybe this should be deferred to V3?

EdSchouten:
I'm fine with deferring it. That said, supported_compressors is relatively new. Maybe it's only worth leaving digest_function intact?

bergsieker:
I feel like there's some value in conforming to the style guide for things like this. It *shouldn't* cause problems (beyond a broken build) for real-world use cases, but as you point out Eric there are technically ways to depend on the field names. Overall this is much less disruptive than the usual add-deprecate-remove cycle if we wanted to replace these fields with properly named ones.

EdSchouten:
> Also, GitHub won't let me comment on the line but `ExecutionCapabilities.digest_function` should be kept consistent with `CacheCapabilities.digest_function[s]`.

ExecutionCapabilities' digest_function field is not repeated, which is why I left it singular. (I don't understand why that field can't be repeated, by the way.)

juergbi:
> It _shouldn't_ cause problems (beyond a broken build) for real-world use cases, but as you point out Eric there are technically ways to depend on the field names.

In Python this API break will only be caught at runtime. Tests would hopefully catch this, however, there is a potential risk of real-world issues for code in Python and similar languages.

EricBurnett:
> > Also, GitHub won't let me comment on the line but `ExecutionCapabilities.digest_function` should be kept consistent with `CacheCapabilities.digest_function[s]`.
> 
> ExecutionCapabilities' digest_function field is not repeated, which is why I left it singular. (I don't understand why that field can't be repeated, by the way.)

Oh whoops! Good catch. That actually does matter for execution, where Merkle trees and similar would be ambiguous if multiple were reported. (They don't specify the digest per-entry inline). So nevermind, nothing to change there. And a good example of why repeated fields being pluralized would make differences clearer :).

EdSchouten:
> > > Also, GitHub won't let me comment on the line but `ExecutionCapabilities.digest_function` should be kept consistent with `CacheCapabilities.digest_function[s]`.
> > 
> > 
> > ExecutionCapabilities' digest_function field is not repeated, which is why I left it singular. (I don't understand why that field can't be repeated, by the way.)
> 
> Oh whoops! Good catch. That actually does matter for execution, where Merkle trees and similar would be ambiguous if multiple were reported. (They don't specify the digest per-entry inline). So nevermind, nothing to change there. And a good example of why repeated fields being pluralized would make differences clearer :).

Could you maybe clarify this? As in, Buildbarn supports multiple digest functions transparently. We currently set this to SHA-256, simply because the protocol requires this, but to me it's not clear why we have to pick exactly one.

mostynb:
> In Python this API break will only be caught at runtime. Tests would hopefully catch this, however, there is a potential risk of real-world issues for code in Python and similar languages.

This would only break when the python implementation updates to the new version of the .proto file though IIUC. We could also bump the minor version of the API to give a stronger hint that testing should be performed.

EricBurnett:
> Could you maybe clarify this? As in, Buildbarn supports multiple digest functions transparently. We currently set this to SHA-256, simply because the protocol requires this, but to me it's not clear why we have to pick exactly one.

Well, the action cache is a mapping from digest->action proto, which assumes there's a unique digest per action. Action inputs are specified as a merkle tree of protos, referenced recursively by digests; output metadata in ActionResult is equally digest-referenced. Nowhere in any of this is it possible to specify what digest function is being used, nor defined semantics for dealing with the ambiguity.

That said, it's definitely possible to fake it - e.g. use the length of a digest to infer which kind of digest between functions of different output lengths; implicitly require trees are internally composed of only one digest type; output digests with a function that matches the input Action proto request; etc. (I'm assuming that's roughly what Buildbarn does?) But that's well outside the spec, which right now is written to assert that trees must be deterministic, etc.

This is less true for the CAS, but still a little true I'll note - if a server wants to validate that input bytes actually match the purported digest, it needs to know how to hash it, which is not explicitly present in e.g. the ByteStream protocol. (Though you can again infer it by digest length). So I'll note that it's a little inconsistent to have digest_function repeated in CasCapabilities also.

If you want to make digest function be repeated for ExecutionCapabilities (or if you're feeling helpful, to clarify the semantics of digest_functions itself), I'd suggest documenting the semantics explicitly throughout the API. Dictate that digests must be differentiable by length, that trees must be internally consistent, that input and output functions should match, etc. At that point, expanding ExecutionCapabilities to have a list would I think also be fine?

EdSchouten:
> That said, it's definitely possible to fake it - e.g. use the length of a digest to infer which kind of digest between functions of different output lengths; implicitly require trees are internally composed of only one digest type; output digests with a function that matches the input Action proto request; etc. (I'm assuming that's roughly what Buildbarn does?) But that's well outside the spec, which right now is written to assert that trees must be deterministic, etc.
>
> This is less true for the CAS, but still a little true I'll note - if a server wants to validate that input bytes actually match the purported digest, it needs to know how to hash it, which is not explicitly present in e.g. the ByteStream protocol. (Though you can again infer it by digest length). So I'll note that it's a little inconsistent to have digest_function repeated in CasCapabilities also.

Deriving the digest function from the length is exactly what Buildbarn does, simply because I want to be able to validate all uploads. Also for the download case it's useful: Buildbarn automatically detects corrupted CAS objects and discards them from its bookkeeping.

Even though it's by no means required by the spec, Buildbarn currently has no issues whatsoever dealing with input roots where files use different digest functions. It's perfectly valid to present it an input root where the root directory uses SHA-256, while subdirectories contains files that use MD5. However, it does make sure that all of the outputs of a build action use the same digest function as that derived from the Action digest.

> If you want to make digest function be repeated for ExecutionCapabilities (or if you're feeling helpful, to clarify the semantics of digest_functions itself), I'd suggest documenting the semantics explicitly throughout the API. Dictate that digests must be differentiable by length, that trees must be internally consistent, that input and output functions should match, etc. At that point, expanding ExecutionCapabilities to have a list would I think also be fine?

Sounds good to me! Will see if I can find some time to get that addressed.

EricBurnett:
Thanks Ed!

> Even though it's by no means required by the spec, Buildbarn currently has no issues whatsoever dealing with input roots where files use different digest functions.

While I doubt it would have any material effect beyond reduced cache effectiveness if hit, I'll note that it's implicitly contradictory with the intent of https://github.com/bazelbuild/remote-apis/blob/main/build/bazel/remote/execution/v2/remote_execution.proto#L713 . 

...though we've had discussions elsewhere about wanting to relax the canonicalization constraint anyways, so this is not an actual objection to documenting how support for multiple should work :).

sbunce:
My $0.02.

I don't think names should be changed on existing fields because it'll break people depending on the json format. This came up in an internal code review at my company. We were surprised that a field name would be changed and had to reason about whether it'd break us. It didn't.

Reminds me of this. Heh.
https://en.wikipedia.org/wiki/HTTP_referer

I think it would have been preferable to commit to the bad name. But oh well. 🤷

Issue #170, opened 2020-09-08
Tree Interface


werkt:
A number of material differences exist for Trees throughout the API:

Directories are uploaded individually as blobs.
Trees are fetched in a list streaming fashion from getTree
OutputDirectories are represented as digests to whole-message Trees
The digests for OutputDirectories cannot use getTree to retrieve large directories pre-parsed.
In all of these circumstances, resolving directory references throughout the tree requires computed keying of a digest, with a specific hash function, and integrity throughout (related #159)
With likely others.

For v3, we should provide a transport and storage definition for trees that resolves this bespoke treatment of trees for input/output, and meets all criteria for an efficient and tractable representations

Issue #140, opened 2020-06-01
REv3 idea: CAS.ExpandTree()?


EdSchouten:
Build systems like Bazel currently treat directory outputs as being opaque. There is no support for addressing individual files contained within.

If Bazel wants to pass on a directory output from one action to another, it currently needs to download a Tree object from the CAS and expand it into individual Directory objects. These then need to be reuploaded into the CAS.

This roundtrip could be avoided by adding a dedicated RPC of the shape:
```proto
rpc ExpandTree(Digest) returns (Digest);
```
In this case, the argument corresponds to the digest of the Tree object. The return value corresponds to the digest of the root directory contained within the Tree object. The call should also bump the TTL on all resulting CAS objects.

Do we consider this to be enough of a problem that it makes sense to add such an RPC?

ulfjack:
That seems to be predicated on the assumption that Bazel *is not* going to allow addressing individual files in the future. That assumption does not seem correct to me. I'm not sure if there are features in Bazel around this right now, but it seems like a perfectly safe bet that there will be, even if Bazel team isn't *currently* planning to add them.

The reason for this is simple: adding unnecessary input files to an action is bad for build performance (reduces caching & increases action costs). This is particularly true the *larger* the output directory tree is, i.e., the more you're trying to save network bandwidth with this feature, the worse performance gets. Therefore, I would bet on Bazel adding features to allow addressing individual files or filtering files or something else along these lines that requires Bazel to have a local list of files.

EdSchouten:
> That seems to be predicated on the assumption that Bazel _is not_ going to allow addressing individual files in the future.

No, that's not correct. It is predicated around the fact that in many cases directories declared using `declare_directory()` are simply passed on in literal form. If Bazel actually needs a subset of a Tree (which would likely only be used in certain cases), it is still capable of downloading the Tree object and carving it up as needed.

Even in the case where a Tree needs to be carved up, it might still be preferable to call `ExpandTree()` to at least store all subtrees that end up being used in literal form.

ulfjack:
My position is that rules will (should?) ~always use a subset of larger trees in order to avoid the performance costs. Maybe that's overly optimistic, but even so, if that's the recommended best practice, why spend significant effort on optimizing the other case?

If the tree needs to be carved up, why would you have `ExpandTree` as a separate API compared to requiring action execution to generate / store all the subtrees?

EdSchouten:
[...]

PR #207, opened 2021-08-09
Relax input tree requirements

### LINKED ISSUE #141: V3 idea: Loosen the restriction of action input as Merkle tree
I have observed that Bazel can spend a lot of CPU resources calculating merkel tree digests. This has been discussed in https://github.com/bazelbuild/bazel/issues/10875 and [Extend the Action Cache with alias digests](https://groups.google.com/forum/#!topic/remote-execution-apis/F0Qb4m0J4Vg).

The key point is that the single input Merkle tree only needs to be resolved on cache miss, which should be rare, so the client should be allowed to check for cache hit using something else.

One idea was to create an alias cache entry where the client would be able to calculate the digest in any suitable way. The problem is that the alias has to be uploaded by the clients, a trusted CI machine or an untrusted developer machine, but not by the remote execution server side. Therefore, using action cache alias makes the system vulnerable for cache poisoning.

Instead, @EricBurnett suggests to loosen the restriction on the input to describe partial trees:
> https://github.com/bazelbuild/remote-apis/issues/140#issuecomment-636983411
> For merkle trees as inputs, the general properties we care about are:
> 
>    - Recursively defined, so that sharing trees in inputs doesn't require
>    operating on a whole tree each time
>    - Parallelly uploadable, so that it doesn't add unnecessary round-trips
>    on the order of the depth of the tree.

> https://groups.google.com/forum/#!msg/remote-execution-apis/F0Qb4m0J4Vg/QANi1BMdAgAJ
> I will note that Merkle Trees, when used as inputs, are defined as they are to achieve:
> 1. Reusability (sub-trees shared by two actions will share Merkle Tree nodes),
> 2. Determinism (the same set of inputs will always get the same tree, regardless of client)

What would be a good design?
1. Extend `message Directory` to include more extra roots, not just subdirectories?
2. Let `Action.input_root_digest` be repeated?

Any other design ideas or any ideas to solve the problem in a totally different way?

moroten:
To reduce the workload of the remote execution clients, the requirement of constructing Merkle trees is removed. Instead, the client can choose any structure of it input file tree it would like. A client might choose to replicate the internal dependency graph.

The down side is that different clients, even different versions of the clients, might not get cache hit on effectively the same action. The use case of sharing the cache for different clients is not a realistic use case. Also, clients are most likely stable in their serialization between versions.

Fixes #141.

This first draft basically duplicates the file tree messages, i.e. `DirectoryNode` gets a copy called `RelaxedDirectoryNode`. Another option is to reuse the existing file tree messages (`Directory`, `FileNode` etc.) but let `name` represent a "relaxed path" instead.

### DIFF
--- a/build/bazel/remote/execution/v2/remote_execution.proto
+++ b/build/bazel/remote/execution/v2/remote_execution.proto
@@ -457,14 +457,28 @@ message Action {
   // [ContentAddressableStorage][build.bazel.remote.execution.v2.ContentAddressableStorage].
   Digest command_digest = 1;
 
-  // The digest of the root
-  // [Directory][build.bazel.remote.execution.v2.Directory] for the input
-  // files. The files in the directory tree are available in the correct
-  // location on the build machine before the command is executed. The root
-  // directory, as well as every subdirectory and content blob referred to, MUST
-  // be in the
-  // [ContentAddressableStorage][build.bazel.remote.execution.v2.ContentAddressableStorage].
-  Digest input_root_digest = 2;
+  one_of input_root {
+    // The digest of the root
+    // [Directory][build.bazel.remote.execution.v2.Directory] for the input
+    // files. The files in the directory tree are available in the correct
+    // location on the build machine before the command is executed. The root
+    // directory, as well as every subdirectory and content blob referred to, MUST
+    // be in the
+    // [ContentAddressableStorage][build.bazel.remote.execution.v2.ContentAddressableStorage].
+    Digest input_root_digest = 2;
+
+    // The digest of the root
+    // [RelaxedFileTree][build.bazel.remote.execution.v2.RelaxedFileTree] for the input
+    // files. The files in the directory tree are available in the correct
+    // location on the build machine before the command is executed. The root
+    // directory, as well as every subdirectory and content blob referred to, MUST
+    // be in the
+    // [ContentAddressableStorage][build.bazel.remote.execution.v2.ContentAddressableStorage].
+    //
+    // Support for this field can be checked using the
+    // [Capabilities][build.bazel.remote.execution.v2.Capabilities] API.
+    Digest relaxed_input_root_digest = 11;
+  }
 
   reserved 3 to 5; // Used for fields moved to [Command][build.bazel.remote.execution.v2.Command].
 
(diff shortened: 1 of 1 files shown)


build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -876,6 +890,166 @@ message SymlinkNode {
...
+// Each `Node` contains either its absolute or relative path, either the digest
+// of its content (either a file blob or a `RelaxedFileTree` proto) or a
+// symlink target, as well as possibly some metadata about the file or
+// directory.
+//
+// The paths MUST use `/` as path component separator, even if the environment
+// where the Action is executed use other separators. If the path starts with
+// `/` it is considered to be refering to the root of the file tree, otherwise
+// the path is relative to the parent directory. The path MUST NOT contain any
+// `.` or `..` components. An empty string refers to the same directory.
```
moroten, line 909: Given that the `RelaxedFileTree` contains a base `path`, one can require all children paths to be relative to that base path.

juergbi:
I'm in favor of relaxing the requirements. However, my preferred approach would be to modify `DirectoryNode`, replacing the `digest` by a `one_of` with the choice of `Digest` or an embedded `Directory` (as suggested in https://github.com/bazelbuild/remote-apis/issues/140#issuecomment-725922504). That seems less invasive in the proto definition and probably also in client and server code. Or am I missing an issue or disadvantage with my suggestion?

juergbi:
Any thoughts on the output side? I think we should also replace `tree_digest` in `OutputDirectory` with a `one_of`. Either allow a `RelaxedFileTree` digest or, following the approach of my previous comment, a `Directory` digest. This would allow moving away from `Tree` messages, addressing #170.

This may need a flag in the `Command` to let the client opt-in.

EdSchouten:
Call me crazy, but I've never understood this discussion. Nobody seems interested in presenting in actual detail what the problem is. And also very importantly: nobody has even described why this cannot be fixed by optimizing the client.

For example, I can imagine that Bazel has a very hard time with setups like the following. Assume there are a million files named `"%d/%d" % (i, j)`, where `i` and `j` are both between 0 and 1000. Now let's place them in a thousand filegroups, where each filegroup contains exactly one file from every directory:

```
N = 1000

[
    filegroup(
        name = "fg%d" % j,
        srcs = [
            "%d/%d" % (i, j)
            for i in range(N)
        ],
    )
    for j in range(N)
]
```

If we then merge these file groups all over the place, stuff gets complicated. Computing Merkle trees for that is simply hard. But that's not how any normal source tree works, right? In reality, dependency graphs *roughly* mirror the file system layout. Because they do, you rarely need to construct full Merkle trees. You can often get away with merging existing ones together. Ones that don't tend to have too much overlap, meaning that merging is cheap.

My observation has always been that the slowness is *not* caused by using a Merkle tree per se. It's because the Merkle tree generation is currently just bolted as part of Bazel's REv2 client code. For every action that gets run, Bazel starts from a virtually clean slate when generating the Merkle tree. This is just silly, as actions often share large parts of the file system. If Bazel's core data structures could somehow memoize (abridged versions of) REv2 Merkle trees, then the process would become a lot more light-weight than what it currently is.

I think we should first go back and write a report on what the problem actually is. What is the (amortized) algorithmic complexity of Bazel's Merkle tree computation right now? What is optimal, given the existing protocol? What is the expected improvement, if any, of having this change applied?

moroten:
> Call me crazy, but I've never understood this discussion.

I appreciate that you are honest. Thank you!

> My observation has always been that the slowness is _not_ caused by using a Merkle tree per se. It's because the Merkle tree generation is currently just bolted on as part of Bazel's REv2 client code. For every action that gets run, Bazel starts from a virtually clean slate when generating the Merkle tree. This is just silly, as actions often share large parts of the file system. If Bazel's core data structures could somehow memoize (abridged versions of) REv2 Merkle trees, then the process would become a lot more light-weight than what it currently is.
> 
> I think we should first go back and write a report on what the problem actually is. What is the (amortized) algorithmic complexity of Bazel's Merkle tree computation right now? What is optimal, given the existing protocol? What is the expected improvement, if any, of having this change applied?

The profiling I've done in https://github.com/bazelbuild/bazel/issues/10875 points to two specific functions in Bazel that lists all files and creates the Merkle tree. The most expensive operation there is `O(n*log(n))` for sorting the list of all input files. My focus has been to remove that altogether.

My idea has been to only create the `Directory` message once per path when looking at a single action. Memoizing that is impossible as the dependency graph might look like `/foo/impl.cpp -> /foo/private/inc.h -> /base.h`, in which case you need to know all files before constructing the Merkle tree. Your comment that merging is cheap makes sense. Then each part of the dependency chain will get a new root of the Merkle tree, but merging three times should still be cheaper than existing `O(n*log(n))`. I will give that a try in Bazel.

Assuming 20 files per folder, 200 folders => 4000 files where up to 40 merges are needed, then `n*log(n) = 4000*12 = 48000` is significantly larger than `40*20 = 800` so such an implementation should be enough if the constant factor is small enough.

Measurements show that we save 30% of Bazel's execution time when removing the current Merkle tree code (measured on our production CI over two days). Let's see what I end up with when merging Merkle trees.

moroten:
I've created https://github.com/bazelbuild/bazel/pull/13879 which implements Merkle tree memoization and merging in Bazel. One example of 3000 inputs cut the calculation time from 78 ms to 0.7 ms. With the fast alias hash I've tried before, I got down to 0.1 ms. Bazel has more overhead, in the range of 10 ms for my example action, so 1 ms or 0.1 ms doesn't really make any difference.

Now, let me do some measurements on this patch for a few days before I post real world results.

EdSchouten:
Wow! That's great news, @moroten. Thanks a lot for working on this.

moroten:
Closing this issue as bazelbuild/bazel#13879 solves the problem. Further discussions about closing #141 can be held there.

PR #208, opened 2021-08-11
Advise cache clients not to inline data when uploading ActionResult messages

### LINKED ISSUE #206: clarification: are clients allowed to inline data in ActionResult?
`ActionResult`'s `stdout_raw` and `stderr_raw` fields have comments which state that servers must not inline these unless the client requests it, but do not specify if clients are allowed to inline data here. Similarly for the `output_files` field, which has a `contents` field.

The implementation in bazel-remote (cache server only) allows this, it deinlines these fields before storage, and re-inlines them later before responding to GetActionResult. I have a few custom (proprietary) clients which take advantage of this to minimise roundtrips.

But other implementations like buildbarn quietly accept `ActionResult` messages with inlined data, and then incorrectly return these unchanged for client requests even if they aren't requesting inlined data.

Should clients be allowed to inline data like this? And if so, should we allow servers to advertise support for this via the Capabilities service? 

And if clients are not be allowed to inline data like this, should we clarify this as "SHOULD NOT" or "MUST NOT" ?

Relates to #202.

mostynb:
This possibility was not originally considered, and we suspect that most server implementations do not handle it properly. Let's forbid this for now, to promote compatibility between implementations.

Resolves #206.

### DIFF
--- a/build/bazel/remote/execution/v2/remote_execution.proto
+++ b/build/bazel/remote/execution/v2/remote_execution.proto
@@ -1132,6 +1132,7 @@ message ActionResult {
   // [GetActionResultRequest][build.bazel.remote.execution.v2.GetActionResultRequest]
   // message. The server MAY omit inlining, even if requested, and MUST do so if inlining
   // would cause the response to exceed message size limits.
+  // Clients SHOULD NOT populate this field when uploading to the cache.
   bytes stdout_raw = 5;
 
   // The digest for a blob containing the standard output of the action, which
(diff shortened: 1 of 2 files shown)


EdSchouten (approved):
LGTM!

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1132,6 +1132,7 @@ message ActionResult {
   // [GetActionResultRequest][build.bazel.remote.execution.v2.GetActionResultRequest]
   // message. The server MAY omit inlining, even if requested, and MUST do so if inlining
   // would cause the response to exceed message size limits.
+  // Clients SHOULD NOT populate this field when uploading to the cache.
```
sstriker, line 1135: SHOULD NOT or MUST NOT?

mostynb: I'm leaning towards SHOULD NOT because it's unlikely that things will break for inline-happy clients and  naïve servers.

sstriker (approved):

EricBurnett:
LGTM as well, and no objections were raised in the monthly meeting from any party.

PR #210, opened 2021-11-12
Tidy up the semantics of Command.arguments[0]

EdSchouten:
Bazel Buildfarm, Buildgrid and Buildbarn all perform resolution relative
to the working directory of an action; not the input root directory.
Instead of requiring that all implementations are updated, we should
consider just altering the spec.

Performing resolution relative to the input root directory can also be
very tricky, as it means that argv[0] as visible to the calling process
must also be rewritten. Applications may get confused otherwise. For
example, consider the case where the working directory is "foo" and
argv[0] is "bar/baz". In that case argv[0] as visible to the calling
process must become "../bar/baz" or be made absolute. Making it absolute
is inconsistent with what Bazel does right now. Attempting to keep it
relative can be complex when symbolic links are involved.

Furthermore, the specification doesn't mention what kind of path
separators are used for argv[0]. The only reasonable solution here is to
use path separators that are native to the host, as successive arguments
also need to be provided in that form.

**EDIT:** As discussed on the mailing list, let's re-add PATH resolution. RBE, Buildbarn and Bazel Buildfarm all do that right now.

### DIFF
--- a/build/bazel/remote/execution/v2/remote_execution.proto
+++ b/build/bazel/remote/execution/v2/remote_execution.proto
@@ -537,9 +537,21 @@ message Command {
     string value = 2;
   }
 
-  // The arguments to the command. The first argument must be the path to the
-  // executable, which must be either a relative path, in which case it is
-  // evaluated with respect to the input root, or an absolute path.
+  // The arguments to the command.
+  //
+  // The first argument specifies the command to run, which may be either an
+  // absolute path, a path relative to the working directory, or an unqualified
+  // path (without path separators) which will be resolved using the operating
+  // system's equivalent of the PATH environment variable. Path separators
+  // native to the operating system running on the worker SHOULD be used. If the
+  // `environment_variables` list contains an entry for the PATH environment
+  // variable, it SHOULD be respected. If not, the resolution process is
+  // implementation-defined.
+  //
+  // Changed in v2.3. v2.2 and older require that no PATH lookups are performed,
+  // and that relative paths are resolved relative to the input root. This
+  // behavior can, however, not be relied upon, as most implementations already
+  // followed the rules described above.
   repeated string arguments = 1;
 
   // The environment variables to set when running the program. The worker may
--- a/build/bazel/remote/execution/v2/remote_execution.pb.go
+++ b/build/bazel/remote/execution/v2/remote_execution.pb.go
@@ -470,9 +470,21 @@ type Command struct {
 	sizeCache     protoimpl.SizeCache
 	unknownFields protoimpl.UnknownFields
 
-	// The arguments to the command. The first argument must be the path to the
-	// executable, which must be either a relative path, in which case it is
-	// evaluated with respect to the input root, or an absolute path.
+	// The arguments to the command.
+	//
+	// The first argument specifies the command to run, which may be either an
+	// absolute path, a path relative to the working directory, or an unqualified
+	// path (without path separators) which will be resolved using the operating
+	// system's equivalent of the PATH environment variable. Path separators
+	// native to the operating system running on the worker SHOULD be used. If the
+	// `environment_variables` list contains an entry for the PATH environment
+	// variable, it SHOULD be respected. If not, the resolution process is
+	// implementation-defined.
+	//
+	// Changed in v2.3. v2.2 and older require that no PATH lookups are performed,
+	// and that relative paths are resolved relative to the input root. This
+	// behavior can, however, not be relied upon, as most implementations already
+	// followed the rules described above.
 	Arguments []string `protobuf:"bytes,1,rep,name=arguments,proto3" json:"arguments,omitempty"`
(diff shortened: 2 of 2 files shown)


build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -539,7 +539,10 @@ message Command {
 
   // The arguments to the command. The first argument must be the path to the
   // executable, which must be either a relative path, in which case it is
-  // evaluated with respect to the input root, or an absolute path.
+  // evaluated with respect to the working directory, or an absolute path. Path
```
mostynb, line 542: The first sentence says that the first argument must be either relative or absolute, but doesn't mention the third possibility- a command name without any path separators.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -539,7 +539,10 @@ message Command {
...
+  // separators native to the operating system running on the worker are used.
+  // Paths that do not contain contain any path separators must be resolved
+  // using the operating system's equivalent of the PATH environment variable.
```
mostynb, line 545: Are shell builtins or functions allowed here? Should we say anything about those possibilities?

(Useful posix reference: https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_09_01_01)

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -538,8 +538,23 @@ message Command {
...
-  // evaluated with respect to the input root, or an absolute path.
+  // executable, which is interpreted as follows:
+  //
+  // - If the path contains no path separators, the path of the executable is
+  //   resolved using the operating system's equivalent of the PATH environment
+  //   variable. Workers are not required to accept shell builtins or functions.
+  // - If the path is relative, it is resolved relative to the working directory
+  //   of the action.
+  // - If the path is absolute, it is resolved relative to the operating
+  //   system's equivalent of the root directory.
```
mostynb, line 549: I would drop this sentence- absolute paths are already resolved, and windows doesn't have a root directory.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -538,8 +538,23 @@ message Command {
...
+  //
+  // Path separators native to the operating system running on the worker are
```
mostynb, line 551: s/are/SHOULD/ ?

build/bazel/remote/execution/v2/remote_execution.proto
EdSchouten: I added this sentence, because I wanted to make it explicit that paths are allowed to be absolute. Maybe the second part of the sentence can be improved? What are your thoughts?

mostynb: How about something like this:

"The first argument specifies the command to run, which may be either an absolute path, a path relative to the working directory, or an unqualified path (without path separators) which will be resolved using the operating system's equivalent of the PATH environment variable."

EdSchouten: I processed this change two weeks ago, but forgot to mark it as resolved. Doing this retroactively. :-)

mostynb (approved):
lgtm

bergsieker (approved):

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -537,9 +537,18 @@ message Command {
...
 
-  // The arguments to the command. The first argument must be the path to the
-  // executable, which must be either a relative path, in which case it is
-  // evaluated with respect to the input root, or an absolute path.
+  // The arguments to the command.
+  //
+  // The first argument specifies the command to run, which may be either an
+  // absolute path, a path relative to the working directory, or an unqualified
+  // path (without path separators) which will be resolved using the operating
+  // system's equivalent of the PATH environment variable. Path separators
```
ulfjack, line 545: Maybe specify if this takes into account `PATH` env variables specified in the request and what happens if no such variable is specified.

bergsieker: +1

EdSchouten: I have added the following:

> If the `environment_variables` list contains an entry for the PATH environment variable, it SHOULD be respected. If not, the resolution process is implementation-defined.

I'm not too happy with introducing implementation-defined behaviour, but at least it matches POSIX:

https://pubs.opengroup.org/onlinepubs/9699919799.2018edition/basedefs/V1_chap08.html#tag_08_03

> If PATH is unset or is set to null, the path search is implementation-defined.

EdSchouten:
Hey @ulfjack, could you approve this PR if you're fine with the latest version?

ulfjack (approved):

PR #214, opened 2022-02-04
Add execution_uuid for executed actions

moroten:
A unique identifier for the execution of an action makes it possible to distinguish different action executions from each other, e.g. to identify if the action cache has been updated with new results.

This was initially discussed at the _Remote Execution API and Implementations_ meeting 2022-01-11, see [notes](https://docs.google.com/document/d/1EtQMTn-7sKFMTxIMlb0oDGpvGCMAuzphVcfx58GWuEM/edit#heading=h.l0ia2k6779xu).


moroten:
@werkt , @EdSchouten: Here is the PR with the `execution_uuid` field that was discussed on the RE API meeting a few weeks ago.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1407,6 +1412,11 @@ message ExecuteOperationMetadata {
   // being executed.
   Digest action_digest = 2;
 
+  // A unique identifier for an ongoing execution of the action, which will
+  // also be returned as `ActionResult.execution_metadata.execution_uuid`.
+  // If the execution is restarted, the server SHOULD provide a new identifier.
+  string execution_uuid = 5;
```
EdSchouten, line 1418: Maybe it's not important, but still worth asking: at what point in time do we generate an `execution_uuid`? Does a `CACHE_CHECK` or `QUEUED` operation already have a non-empty `execution_uuid`, or does it only apply to operations in stages `EXECUTING` or `COMPLETED`?

EdSchouten (approved):
Thanks for working on this, Fredrik!

build/bazel/remote/execution/v2/remote_execution.proto
moroten: I thought about it but realized it might be up to the server side to decide. We would mention that in the description.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1407,6 +1412,11 @@ message ExecuteOperationMetadata {
   // being executed.
   Digest action_digest = 2;
 
+  // A unique identifier for an ongoing execution of the action, which will
+  // also be returned as `ActionResult.execution_metadata.execution_uuid`.
```
bergsieker, line 1416: Does the execution_metadata.execution_uuid have to match this one? Can the identifier change, either during the lifetime of the ExecuteOperationMetadata (e.g., if the server merges this action with another one) or at the end of execution?

build/bazel/remote/execution/v2/remote_execution.proto
bergsieker: I'd be inclined towards a semantic of generating an ID when the action starts to execute and not before.

bergsieker, line 1418: I think it's better to stick to the generic "id" rather than the specific "uuid". I don't think we need to constrain implementors specifically to UUIDs.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -926,6 +926,11 @@ message Digest {
 
 // ExecutedActionMetadata contains details about a completed execution.
 message ExecutedActionMetadata {
+  // A unique identifier for the execution of an action. This makes it possible
+  // to distinguish different action executions from each other, e.g. to
```
bergsieker, line 930: I'm curious about the use case for this. My gut feeling is that there are a lot of bad ways that clients could start to depend on something like knowing whether than action was executed or cached.

build/bazel/remote/execution/v2/remote_execution.proto
moroten: Yes, the identifier can change, e.g. through merges or restarts of the action. It should therefore match `execution_metadata.execution_uuid`.

build/bazel/remote/execution/v2/remote_execution.proto
moroten: Correct, I'll change that. The description of "unique id" is enough.

moroten: I would like to leave it open for the server to decide, but put an expectation on it being provided during `EXECUTING` and `COMPLETED`. Should this field be mandatory in v3 but marked as optional using "if set, ..." for v2?

moroten:
As discussed on [RE API forum 2022-02-08](https://docs.google.com/document/d/1EtQMTn-7sKFMTxIMlb0oDGpvGCMAuzphVcfx58GWuEM/#heading=h.l0ia2k6779xu), the use case looks good regarding logging and cross-referencing between client and server. It’s useful to propagate this information on the server side. It might also be useful on the client side, but no concrete use case at the moment.

The conclusion is to add it as a server specific message in `ExecutedActionMetadata.auxiliary_metadata` instead. If multiple users are interested in the future, we can reopen this PR again.

PR #216, opened 2022-03-08
Specify acceptable optimizations for input roots

EricBurnett:
Clarify semantics for how files in the input root should appear to actions, and how implementations may deviate from this ideal for performance reasons.

This wording is proposed as one possible direction we could specify, for further discussion following conversation at the Remote Execution WG monthly.

@EdSchouten please go ahead and propose your preferred alternative. This is not intended to be committed until/unless we've had a chance to see the alternatives and attain consensus here.

### DIFF
--- a/build/bazel/remote/execution/v2/remote_execution.proto
+++ b/build/bazel/remote/execution/v2/remote_execution.proto
@@ -465,6 +465,16 @@ message Action {
   // directory, as well as every subdirectory and content blob referred to, MUST
   // be in the
   // [ContentAddressableStorage][build.bazel.remote.execution.v2.ContentAddressableStorage].
+  //
+  // Implementations SHOULD make it appear as if all files in the input root are
+  // actually present in the specified locations in the input filesystem, and are
+  // unique copies of the input blobs (unless otherwise specified in the Directory
+  // tree). However, for optimization reasons implementations MAY break this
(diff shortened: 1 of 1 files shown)


EdSchouten (approved):

jmillikin:
Related: https://github.com/bazelbuild/bazel/issues/10299 is a Bazel issue that requests the ability to mark files as non-symlinkable, to be more compatible with JavaScript ecosystem tools. The workaround I used in that case was inelegant: forcing non-symlink placement for any files that are executable or end in `.js`.

It would be nice if clients could tag specific files as needing special treatment, even if that tag were as general-purpose as "do not optimize placement or storage of this file at all".

ulfjack:
It might make sense to have more of a discussion around what the options and expectations are:
1. There are four ways of making files appear: symlink, hardlink, copy, and clonefile (not all of these are available on all platforms or file systems).
2. There are also questions around permissions (are files read-only or read-write? who owns the files?), which aren't covered so far.

I think most people would expect that the files are standalone (copy or clonefile), and that they can be modified. However, in our experience, copy is the slowest option, and hardlinked files have to be read-only to avoid corruption. We have an undocumented platform option to select between these options as an escape hatch (can be set on a per-action basis), but haven't actually needed it so far.

PR #218, opened 2022-04-12
Update README.md to reference Remote Asset API in additon to REAPI

jeremiahbonney:
When looking for a brief description of the Remote Asset API, I realized that it was only discoverable by traversing the repo and mentioned anywhere. This PR reworks the intro to the `README.md` a bit to add a link and simple description of the Remote Asset API to make it more discoverable.

Happy to rework the description or do something like have a dedicated section for the other APIs in this repo that aren't the REAPI, let me know what y'all think!

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,12 +1,38 @@
-# Remote Execution API
+# remote-apis
 
-The Remote Execution API is an API that, at its most general, allows clients to
-request execution of binaries on a remote system. It is intended primarily for
-use by build systems, such as [Bazel](bazel.build), to distribute build and test
-actions through a worker pool, and also provide a central cache of build
-results. This allows builds to execute faster, both by reusing results already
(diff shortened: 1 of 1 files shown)


bergsieker (approved):

bergsieker (changes requested):
Actually, we should probably also include the logstream API here, too. Maybe an intro paragraph tying all of these together as "things related to remote execution" and then a paragraph/bullet for each API?

EdSchouten (approved):

sstriker (approved):

sstriker:
> Actually, we should probably also include the logstream API here, too. Maybe an intro paragraph tying all of these together as "things related to remote execution" and then a paragraph/bullet for each API?

@bergsieker I believe this is now done and should resolve your request for changes :-)

PR #225, opened 2022-05-12
Document errors returned by Execution/WaitExecution

Yannic:


Yannic:
@ulfjack PTAL

EdSchouten (approved):

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -120,6 +121,21 @@ service Execution {
...
+  // `status` field of
+  // [ExecuteResponse][build.bazel.remote.execution.v2.ExecuteResponse].
+  // The possible errors include:
+  //
+  // * `NOT_FOUND`: The client requested an
+  //   [Operation][google.longrunning.Operation] that is not known by the server
+  //   (e.g., because
+  //   [Execute][build.bazel.remote.execution.v2.Execution.Execute] never
+  //   returned that value, or because the action was already executed and
+  //   returned the result).
```
EdSchouten, line 136: Buildbarn automatically terminates execution of actions in case it has not observed any clients during the last minute that have called Execute() or WaitExecution() on it. This means that a client could also get `NOT_FOUND` in that case.

Maybe worth calling it out? Or is that too oddly specific?

EdSchouten (approved):

build/bazel/remote/execution/v2/remote_execution.proto
bergsieker: I agree that the case of cancelled actions should be explicitly documented here.

I'd like some clarification around separating the Operation error code from the Execution error code. If a client asks for an operation that was completed (or cancelled) in the past, the server may respond with NOT_FOUND if it has already cleared the state related to that Operation, but it is also allowed to respond with the completed Operation.

Yannic: Sorry for the super long delay here. PTAL

PR #226, opened 2022-05-16
Make it explicit that there are no method-specific errors for GetCapabilities

Yannic:


Yannic:
Ping?

EdSchouten (approved):
Sorry for the delays; overlooked this notification.

bergsieker (approved):

Issue #229, opened 2022-08-22
Replace message Tree with a topologically sorted varint delimited stream of Directory messages


EdSchouten:
I know that on numerous occasions we've had discussions within the working group on whether the Tree vs. Directory dichotomy is truly necessary, and whether there are ways to eliminate it. Let me make it clear up front that this issue is filed under the assumption that we do want to keep these separate like we have today, which is why I'm not posting this under #159 or #170. It's just low hanging fruit.

The last couple of days I've worked on optimising handling of large Tree objects in [bb-clientd](https://github.com/buildbarn/bb-clientd). Whereas some of the other Protobuf messages can sometimes become hundreds of kilobytes in size (e.g., Command), it's not impossible for Trees to become dozens of megabytes. While working on optimising this area, I realised that my life would have been a lot easier if Trees were stored as a varint delimited stream of Directory messages (just like Bazel `--build_event_binary_file`). My suggestion would be to make it topologically sorted (parents before children), so that the root of the Tree comes first. Some advantages:

1. Because of the topological sorting, you can instantiate the contents of a Tree on a native file system in a streaming manner. Memory consumption of such an extraction tool would only be proportional to the number of referenced child directories that have not yet been observed.
2. Similarly, you can do one-shot path lookups (for paths without "..") by doing a forward scan against the stream, only unmarshaling directory objects that match the length and hash yielded by the parent.
3. Right now you see that consumers of Tree first unmarshal the message, and then marshal each of the child directories again, just so that their hash can be computed. This implicitly assumes that the producer and consumer use the same strategy for marshaling. By using a varint delimited stream, consumers have direct access to directories in marshaled form, meaning that both this assumption and the redundant marshaling are eliminated.
4. In bb-clientd I need to be able to have fast random access to Tree objects. To facilitate this, I'm basically storing the Tree object in the client-side CAS just once, but creating n+1 entries in the index. One for the Tree as a whole, and one for each of the child directories, pointing to byte ranges in the Tree's bytestream that correspond to the individual directory messages. That way these directory messages become directly addressable. This happens to be possible today, only because Protobuf messages are encoded identically regardless of the depth at which they are placed. The format I'm proposing not only allows me to eliminate this assumption, it means I can also compute the n additional entries in a streaming manner.

EdSchouten:
It looks like I've been able to solve (3) and (4) on the Buildbarn side without making any protocol changes, as it turns out that Protobuf's wire format has some interesting properties:

[...]

PR #230, opened 2022-09-30
Add a hint for indicating that a Tree is topologically sorted

EdSchouten:
I'm currently trying to improve the performance of handling of large    output directories (Tree messages), having sizes in the order of    hundreds of megabytes. In the process, I have realised that there is a    lot of value in enforcing that the Directory messages contained in them    are topologically sorted. Two practical use cases:
    
- When instantiating the contents of a Tree on a local file system,      having the Tree be topologically sorted allows you to immediately      create files and directories in the right place.
    
- When needing to resolve the properties of a single file by path, a      topologically sorted Tree permits resolution (of paths without ..) by doing a simple forward      scan.
    
Especially when other features like compression are taken into account,    it's useful if Tree messages can be processed in a streaming manner.
    
One practical issue is that most Protobuf libraries don't offer APIs for    processing messages in a streaming manner. This means that implementors    who want to achieve these optimisations will need to write their own    message parsers; at least for the Tree itself. To make this as    painless as possible, we also require that the Tree is stored in some    normal form.
    
Fixes: #229


EdSchouten:
Cc @werkt, as he also wanted to make improvements to Tree in #159.

EdSchouten:
Also cc @peterebden, as he made changes in this area as well in #223

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1235,6 +1235,33 @@ message OutputDirectory {
...
+  //
+  // Additionally, the Tree MUST be encoded as a stream of records,
+  // where each record has the following format:
+  //
+  // - A single tag byte, having one of the following values:
+  //   - First record (the root directory): 0x0a
+  //   - Any subsequent records (child directories): 0x12
+  // - The size of the directory, encoded as a base 128 varint.
+  // - The contents of the directory, encoded as a binary serialized
+  //   Protobuf message.
```
mostynb, line 1260: Do all these requirements come from the proto3 encoding?

EdSchouten: > Do all these requirements come from the proto3 encoding?

Yes, but they have been strengthened. For example, Protobuf field tags are also encoded as base 128 varints, meaning that you could also write tag 0x0a (for the root directory) as one of the following:

- 0x8a 0x00
- 0x8a 0x80 0x00
- 0x8a 0x80 0x80 0x00
- 0x8a 0x80 0x80 0x80 0x00
- 0x8a 0x80 0x80 0x80 0x80 0x80 0x80 0x80 ... 0x80 0x00

Also, plain Protobuf would permit placing the root directory anywhere you want; it doesn't need to come before the child directories. It may even be placed in between two of them. Both of these things are of course completely undesirable, which is why I decided to give a complete description of what the binary format looks like.

> I think it's worth documenting where these values come from.

Yes! I've just made two improvements:

1. I've added derivations of where 0x0a and 0x12 come from. They are basically `(x << 3) | 2`, where `x` is the field number.
2. I've added a link to the Protobuf encoding format.

EdSchouten:
Thanks everyone for your time discussing this proposal today. I've made the change suggested by @sstriker, documenting that manual construction of Tree objects without using the Protobuf library is recommended.

mostynb (approved):
You mentioned that this implies a limit of 16(?) possible fields in this message- I think it might be worth mentioning that so we don't forget about it. Otherwise this lgtm.

EdSchouten:
> You mentioned that this implies a limit of 16(?) possible fields in this message- I think it might be worth mentioning that so we don't forget about it. Otherwise this lgtm.

I think this was @bergsieker asking whether or not we should require that the tag is encoded as a single byte. In my opinion this is acceptable, for the following reason.

The tag is essentially just a variable length integer, having value `(fieldNumber << 3) | fieldType`. In my proposal I'm stating that the field number can only be 1 or 2, with no room for future extensions. This means that the tag can only be a 5-bit number, which can always be encoded in a single byte.

I have just extended the phrasing to clarify this.

EdSchouten:
Draft PR against Bazel to start generating topologically sorted trees: https://github.com/bazelbuild/bazel/pull/16463

I will mark it ready for review as soon as this PR lands.

EdSchouten:
@bergsieker Would you be interested in merging this PR for me? Thanks!

bergsieker (approved):

PR #236, opened 2022-11-17
Support multiple remote execution digest functions

EdSchouten:
Even though the cache can announce support for multiple digest functions, the remote execution system cannot. Let's address this inconsistency by adding a new repeated field that should in the (very) long term replace the singular field.


EdSchouten:
Draft PR to let Bazel respect this: https://github.com/bazelbuild/bazel/pull/16791

sstriker:
As per the meeting yesterday, awaiting changes by @EdSchouten

EdSchouten:
Hey @sstriker, @EricBurnett, @bergsieker,

As discussed during Tuesday's meeting, I have revised this PR, taking two things into consideration:

- Ensuring that none of the phrasing in the specification assumes a single digest function: I looked through the spec, and I wasn't able to find such assumptions/contradictions, so I think we're good. Be sure to point any out if you find them!
- Tighten the phrasing to state that even if multiple digest functions are supported, that only a single one may be used by individual actions.

PTAL!

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1854,6 +1857,20 @@ message ExecutionCapabilities {
...
+  // If this field is set, it MUST also contain digest_function.
+  //
+  // Even if the remote execution system announces support for multiple
+  // digest functions, individual execution requests may only reference
+  // CAS objects using a single digest function. For example, it is not
+  // permitted to execute actions having both MD5 and SHA-256 hashed
+  // files in their input root.
+  //
+  // The CAS objects referenced by action results generated by the
+  // remote execution system will use the same digest function as the
```
mostynb, line 1871: s/will/MUST/ here ?

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1843,7 +1843,10 @@ message CacheCapabilities {
 
 // Capabilities of the remote execution system.
 message ExecutionCapabilities {
-  // Remote execution may only support a single digest function.
+  // Legacy field for indicating which digest function is supported by
+  // the remote execution system. Implementations should consider the
+  // repeated digest_functions field first, falling back to this
+  // singular field if unset.
   DigestFunction.Value digest_function = 1;
```
mostynb, line 1850: Reading through https://developers.google.com/protocol-buffers/docs/encoding#optional it looks like it might actually be OK to upgrade a field to be repeated:

> Normally, an encoded message would never have more than one instance of a non-repeated field. However, parsers are expected to handle the case in which they do. For numeric types and strings, if the same field appears multiple times, the parser accepts the last value it sees.

Could we upgrade this field to repeated, and state that the preferred value for legacy clients should be specified last?

EdSchouten: I thought about this, but there's also a bit more risk to that. Back in #205 I made a very minor change to the protocol to pluralize the name of a repeated field. After we merged that, we got a report or two from people who were dependent on the Protobuf JSON encoding of these capabilities and ran into issues because of that change. My fear is that if we make an in-place change like that, we're going to run into the same kinds of issues again.

Happy to be persuaded otherwise, as I don't really depend on the JSON encoding of messages like these. Just want to make sure I don't make life of others harder.

mostynb: IIRC the problem in #205 was due to renaming the field, whereas we could keep the existing singular name and just make the field repeated (and state that servers should report the preferred value for legacy clients as the last one).

EdSchouten: I don't think you can if you want to keep compatibility, both for the binary and the JSON format. Here, consider this code in the Protobuf library:

https://github.com/protocolbuffers/protobuf-go/blob/master/encoding/protojson/decode.go#L506-L508

Whenever the descriptor states that a message field is repeated, it only allows the JSON representation to be a list. It doesn't fall back to interpreting it as a scalar. Conversely, `unmarshalScalar()` doesn't permit values to be lists.

peterebden: Yep, making a field repeated is a breaking change in the JSON serialisation - at least for most clients I'm aware of.

mostynb: Ah, ok. Scratch that idea then.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1843,7 +1843,10 @@ message CacheCapabilities {
 
 // Capabilities of the remote execution system.
 message ExecutionCapabilities {
-  // Remote execution may only support a single digest function.
+  // Legacy field for indicating which digest function is supported by
+  // the remote execution system. Implementations should consider the
+  // repeated digest_functions field first, falling back to this
+  // singular field if unset.
```
mostynb, line 1849: Should we be explicit and say "This field MUST be set." or "This field MUST be set to a value other than UNKNOWN." here?

EdSchouten: I'm not sure we should go that far. Once we're a couple of years down the line and most clients have logic to process both `digest_function` and `digest_functions`, I guess it wouldn't be too bad if a clean slate server implementation doesn't set this field.

mostynb: I think that would be a breaking change which needs a version bump.

In the meantime we already say that digest_function must be set if digest_functions is set, it would be strange to allow neither to be set IMO.

EdSchouten: Gotcha! Just changed the description of `digest_function` to make it required. PTAL!

mostynb (approved):

EdSchouten:
Thanks for the review, @mostynb!

EdSchouten:
@sstriker @bergsieker @EricBurnett Hey there! Would you folks be interested in reviewing this PR, and PR #235? Thanks!

EricBurnett (approved):

EricBurnett:
@EdSchouten apologies for the long delay!

@sstriker I see you added yourself; are you wanting to give input on this? Otherwise I'll give it a couple days and merge, as I think it's as discussed in the monthly.

tylerwilliams:
Another vote for merging this :)

sstriker (approved):

sstriker:
> @EdSchouten apologies for the long delay!
> 
> @sstriker I see you added yourself; are you wanting to give input on this? Otherwise I'll give it a couple days and merge, as I think it's as discussed in the monthly.

My intent was to land this somewhere after last meeting.  Then my 'round tuits went up in thin air.  Apologies for the wait.

PR #237, opened 2023-01-08
platform.md: Add LoongArch ISA name `la64v100`

cheese:
This adds the base ISA [1] of the LoongArch architecture to the platform lexicon.

[1] https://loongson.github.io/LoongArch-Documentation/LoongArch-toolchain-conventions-EN.html

### DIFF
--- a/build/bazel/remote/execution/v2/platform.md
+++ b/build/bazel/remote/execution/v2/platform.md
@@ -33,6 +33,7 @@ The following standard property `name`s are defined:
     - `arm-a64-be` (big endian)
     - `arm-t32` (little endian)
     - `arm-t32-be` (big endian)
+    - `la64v100` (little endian)
     - `power-isa-be` (big endian)
     - `power-isa-le` (little endian)
     - `rv32g` (little endian)


EdSchouten (approved):

nanonyme:
RE: legal on arches, isn't a change like this so small that independent of CLA it should not really be problematic?

EdSchouten:
Hi @nanonyme,
This is a fair question, whether the CLA should be required for this repository. We brought it up during today's remote execution working group meeting, but unfortunately nobody from Google attended it. I have already placed it on the agenda for next month's meeting. Will keep you posted!

EricBurnett:
Apologies for missing the meeting today. I checked the CLA policy documentation internally and I don't see any provision for bypassing it on account of PR size. And logistically, I would not really expect so - a sequence of small changes could become a large change, which the auditing tooling would have to account for, etc...for auditability purposes if nothing else I think Google-run repos (including everything in `bazelbuild`) do require CLAs on all PRs. Sorry :(.

EdSchouten:
I think the question we had in the monthly meeting was whether the CLA should apply to this repo to begin with, considering it merely captures a specification/standard, not an implementation.

nanonyme:
Note that is now a repeating Achille's heel of remote execution ecosystem at large. No one who contributes ISA names (which I don't think are original work, they are just ISA names; if these are original work then they are original work of the creator of the ISA) into the documentation has signed Google CLA resulting in many cases the PR's dragging on for years. While clearly simplest option is @cheese simply signs CLA, this same problem seems to happen every time. Eg it took *two years* to get someone who had signed Google CLA to submit riscv64 ISA names. https://github.com/bazelbuild/remote-apis/pull/176

cheese:
I have signed Google CLA.

PR #238, opened 2023-01-13
Embed partial ExecutedActionMetadata in ExecuteOperationMetadata

roitk:
The `ExecutedActionMetadata` contains a lot of interesting information about the work that was done, such as the worker the job was run on, timestamps for various execution milestones, and so on. This information is currently only available after the work is complete, in the `ActionResult`. However, I think some of these things would be useful to look at while the execution is running as well, especially in the context of UI tools that can show the status of ongoing work (e.g. [bgd-browser](https://gitlab.com/BuildGrid/bgd-browser)).

Unfortunately, there isn't currently a place to put this in the longrunning `Operation` that clients get back when waiting for a job. The only metadata field is used by the `ExecuteOperationMetadata` message, which has a very limited set of fields. So this PR just adds an `ExecutedActionMetadata` field in `ExecuteOperationMetadata`.

### DIFF
--- a/build/bazel/remote/execution/v2/remote_execution.proto
+++ b/build/bazel/remote/execution/v2/remote_execution.proto
@@ -1471,6 +1471,10 @@ message ExecuteOperationMetadata {
   // [ByteStream.Read][google.bytestream.ByteStream.Read] to stream the
   // standard error from the endpoint hosting streamed responses.
   string stderr_stream_name = 4;
+
+  // Additional implementation-specific metadata to send along with the
+  // Operation.
+  repeated google.protobuf.Any auxiliary_metadata = 5;
 }
 
 // A request message for
--- a/build/bazel/remote/execution/v2/remote_execution.proto
+++ b/build/bazel/remote/execution/v2/remote_execution.proto
@@ -1472,11 +1472,9 @@ message ExecuteOperationMetadata {
   // standard error from the endpoint hosting streamed responses.
   string stderr_stream_name = 4;
 
-  // Additional implementation-specific metadata to send along with the
-  // Operation. Servers MAY choose to populate this field with a partially
-  // complete [ExecutedActionMetadata][build.bazel.remote.execution.v2.ExecutedActionMetadata]
-  // message to provide updates on the ongoing execution.
-  repeated google.protobuf.Any auxiliary_metadata = 5;
+  // The client can read this field to view details about the ongoing
+  // execution.
+  ExecutedActionMetadata partial_execution_metadata = 5;
 }
 
 // A request message for
--- a/build/bazel/remote/execution/v2/remote_execution.proto
+++ b/build/bazel/remote/execution/v2/remote_execution.proto
@@ -1473,7 +1473,9 @@ message ExecuteOperationMetadata {
   string stderr_stream_name = 4;
 
   // Additional implementation-specific metadata to send along with the
-  // Operation.
+  // Operation. Servers MAY choose to populate this field with a partially
+  // complete [ExecutedActionMetadata][build.bazel.remote.execution.v2.ExecutedActionMetadata]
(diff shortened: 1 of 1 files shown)


build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1471,6 +1471,10 @@ message ExecuteOperationMetadata {
   // [ByteStream.Read][google.bytestream.ByteStream.Read] to stream the
   // standard error from the endpoint hosting streamed responses.
   string stderr_stream_name = 4;
+
+  // Additional implementation-specific metadata to send along with the
+  // Operation.
+  google.protobuf.Any additional_metadata = 5;
```
EdSchouten, line 1477: Can we make it repeated and call it auxiliary_metadata? That way it’s consistent with the one we have in ExecutedActionMetadata.

roitk: Makes sense! Updated.

EdSchouten (approved):

peterebden:
I like the idea of this - it seems useful to find out where an action is executing (presumably this can be found out on the server side but this makes it easier to see it on the client).

I'm a little dubious about the Any though. It seems tricky for a client to know how to interpret it - there is prior art with the `auxiliary_metadata` field, but it seems like that is more likely to be both populated & read by the server side, which probably agrees about what it wants to put in there.
Maybe it will be sufficient to have some de facto options for this?

peterebden:
Yup, I agree on those. I think `ExecutedActionMetadata` is fairly close, although I also agree with your points before about where it's not so ideal.
On balance maybe that's a good place to start since it already exists and we don't have to define a new message? So just something like "Servers MAY choose to populate this field with a partially complete `ExecutedActionMetadata` message" would give a hint on how to interpret the contents.

roitk:
> On balance maybe that's a good place to start since it already exists and we don't have to define a new message? So just something like "Servers MAY choose to populate this field with a partially complete `ExecutedActionMetadata` message" would give a hint on how to interpret the contents.

I think that's a good compromise. Updated the comment.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1471,6 +1471,12 @@ message ExecuteOperationMetadata {
   // [ByteStream.Read][google.bytestream.ByteStream.Read] to stream the
   // standard error from the endpoint hosting streamed responses.
   string stderr_stream_name = 4;
+
+  // Additional implementation-specific metadata to send along with the
+  // Operation. Servers MAY choose to populate this field with a partially
+  // complete [ExecutedActionMetadata][build.bazel.remote.execution.v2.ExecutedActionMetadata]
```
EdSchouten, line 1477: In that case, please just embed `ExecutedActionMetadata` into `ExecuteOperationMetadata` directly, without using the `repeated google.protobuf.Any` field:
```proto
ExecutedActionMetadata partial_execution_metadata = 5;
```
The current approach introduces an ambiguity, where there's both an `auxiliary_metadata` field in the `ExecuteOperationMetadata` and  `ExecutedActionMetadata` message. Which one should be used for which purpose?

EdSchouten (changes requested):

build/bazel/remote/execution/v2/remote_execution.proto
roitk: Yeah that's fair enough, and `partial_execution_metadata` is a name that makes it clear enough that the fields may not all be present.

Should we explicitly mark this data as optional with a comment like "The server MAY omit this data"?

EdSchouten (approved):

EricBurnett:
This LGTM as well. @sstriker I see you assigned yourself: anything more needed here?

Issue #240, opened 2023-02-21
REv3: Let exit_code be better aligned with C/POSIX


EdSchouten:
I am glad that REv2 already uses `int32` as its data type for `exit_code`. Even though most UNIX systems historically only had 8-bit exit codes, modern versions of POSIX provide interfaces for obtaining the full integer value. Using functions like [`waitid()`](https://pubs.opengroup.org/onlinepubs/9699919799/functions/waitid.html), you can obtain a [`siginfo_t`](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/signal.h.html) whose `si_status` field contains the full value.

Where we do deviate from C/POSIX is that there are no facilities for indicating that a process did not terminate by calling `_Exit()`, but through a signal. This leads to confusing situations where people wonder why their build actions terminate with exit code 139, while in reality it terminated through `SIGSEGV` (signal 11, giving exit code 128+11=139). Even worse are modern versions of Go's `os/exec` package, which will always return an exit code of -1 in case of signal delivery. This means that users can't even figure out why their actions terminated.

One solution to this would be to change the existing field:
```proto
  // The exit code of the command.
  int32 exit_code = 4;
```
To something like this:
```proto
  oneof termination_reason {
    // The command terminated by calling the operating system's equivalent
    // of `_Exit()` with the code provided.
    int32 exit_code = 4;

    // The command terminated by the delivery of a signal.
    int32 signal_number = 13;
 }
```
Even though virtually all operating systems use 9 for `SIGKILL` and 15 for `SIGTERM`, C/POSIX make no such requirement. The only place where that does happen, is in the definition of [the `kill` utility](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/kill.html), where as part of the X/Open System Interfaces (XSI) they provide a hardcoded list of numerical command line options.

My recommendation would thus be to use this instead:
```proto
  oneof termination_reason {
    // The command terminated by calling the operating system's equivalent
    // of `_Exit()` with the code provided.
    int32 exit_code = 4;

    // The command terminated by the delivery of a signal. This field contains the name of
    // the signal, with the "SIG" prefix removed (e.g., "SEGV" for segmentation violations).
    string signal_name = 13;
 }
```

[...]

PR #244, opened 2023-03-08
Add ProcessTerminationResult, for better alignment with POSIX

EdSchouten:
[POSIX issue 7, 2018 edition, section 3.303 "Process Termination"](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_303) states
the following:

> There are two kinds of process termination:
>
> 1. Normal termination occurs by a return from main(), when requested
>    with the exit(), _exit(), or _Exit() functions; or when the last
>    thread in the process terminates by returning from its start
>    function, by calling the pthread_exit() function, or through
>    cancellation.
>
> 2. Abnormal termination occurs when requested by the abort() function
>    or when some signals are received.

Both these cases can be detected by parent processes separately.
Unfortunately, REv2 currently doesn't allow this, as we only provide an
exit code. Implementations such as Buildbarn tend to set the exit code
for abnormal termination to 128+signum, meaning that if a segmentation
fault occurs, an action terminates with exit code 139. This tends to be
fairly confusing for non-expert users.

The goal of this change is to make all of this less confusing. It adds a
new message type named ProcessTerminationResult, which contains the
termination result in a more structured way. It can now either be an
exit code or a signal number. If it turns out that other operating
systems provide even more ways, we can add additional `oneof` cases.

For the exit code in case of normal termination I decided to use an
int64 instead of int32. The reason being that POSIX allows you to get
the full exit code back, using APIs like waitid(). On ILP64 systems, the
exit code may thus be 64 bits.

For the signal in case of abnormal termination I decided to use a
string. The reason being that signal numbers are not covered by POSIX.
Everybody always assumes that 9 is SIGKILL, but POSIX only documents
this as a convention for the command line flags of the `kill` utility.
Not the the actual `SIG*` constants.

Fixes: #240


EdSchouten:
Hi @EricBurnett @juergbi,

Could you folks please take a look at the latest version of this PR? I think I have processed your feedback accordingly:
- The `exit_code` field is no longer marked as deprecated. Instead, each of the `oneof` arms of `ProcessTerminationResult.reason` now documents how its value maps to `exit_code`.
- What @juergbi brought up during yesterday's meeting is correct: exit codes may also be exception values. But what's interesting is that they seemingly don't live in a distinct space from exit code. A process terminating because of an exception will under the hood merely call `TerminateProcess()` with the exception value as the exit code. This means that the `Signaled` arm in `oneof` will remain exclusively used by POSIX-like systems.

Thanks!

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1312,6 +1323,57 @@ message OutputSymlink {
   NodeProperties node_properties = 4;
 }
 
+// The termination result of a process, as reported by the operating
+// system to the parent process (e.g., using wait*() on POSIX-like
+// systems, or GetExitCodeProcess() on Microsoft Windows).
```
EricBurnett, line 1328: I'd suggest removing or weakening the statement "as reported by the operating system..." . This structure of exit_code + process_termination_result I think would actually extend to other concepts as well in the future.

For example, a pattern I expect to need at some point is "flake due to intermediate on-worker infrastructure", where there is a logical test that we care about the result of, but also some intermediate scaffolding that's been injected to run it where we'd like to be able to distinguish flakes due to the scaffolding itself (that doesn't necessarily imply a problem with the test) from flakes of the test.

(To be clear, don't want to extend your PR to cover this; that can be a future discussion. I'd just say that the "process termination result" need not be _only_ OS-provided information.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1312,6 +1323,57 @@ message OutputSymlink {
...
+  }
+
+  message Signaled {
+    // The name of the signal that was delivered that caused the process
+    // to terminate, without the "SIG" prefix (e.g., "BUS", "KILL",
+    // "SEGV").
+    string signal_name = 1;
+  }
+
+  oneof reason {
```
EricBurnett, line 1353: How would you feel about not using a oneof and simply having a set of optional fields? signal_name implying the lack of exit_code makes sense in this particular case, but I'm not sure every future result is going to follow the pattern of being mutually exclusive. E.g. if we also provided info from the syslog or Windows Event Log so you could know _why_ the OS killed the process, say, where you'd get both a SIGKILL and some indication that it was killed due to OOM or too many handles or whatever.

So e.g.
```
message ProcessTerminationResult {
    //...
    int64 exit_code = 1;
    //...
    string signal_name = 2;
    //...
    string human_readable_explanation_for_signal = 3; 
}
```

With comments detailing the relationship between the various fields.

EricBurnett (changes requested):

build/bazel/remote/execution/v2/remote_execution.proto
mostynb: Would that open the door to problems with default values? I'm thinking about `exit_code` in particular.

EricBurnett: Hmm, good point. For many the empty field is trivially uninteresting (e.g. string signal, empty==missing) but you're right about exit_code.

A few options:
 1. Do nothing, since exit_code can be distinguished fairly readily: the only case of concern is when it's 0, where we shouldn't actually need to distinguish if its set or not. (Do we really need to provide a ProcessTerminationResult just to provide a 0 exit_code? Given that the top-level field already exists, it seems just as reasonable to omit the whole message entirely in that case. So "0 means unset" for exit_code is just fine, since it introduces no real ambiguity in practice.)
 1. Do something special for exit_code, since so far it's the only one that has an ambiguous default. A second "has_exit_code" field would do, for example.
 1. If you know a set that will definitely be mutually exclusive, a oneof suffices, and then we can use optional fields for the rest. So e.g. exit_code and signal_name within the oneof, but with the intent for augmenting fields (like the human_readable_explanation_for_signal example) to be optionals outside of it. Which I guess is...what you already have, just with the understanding that maybe not everything will go in the oneof?
 1. Make these structs with a single field instead of raw fields, since you _can_ check for the presence of a struct. E.g.

```
message ProcessTerminationResult {
    message ExitCode {
        //...
        int64 code = 1;
    }
    ExitCode exit_code = 1;
    message Signal {
        //...
        string name = 1;
        string human_readable_explanation_for_signal = 2; 
    }
    Signal signal = 2;
}
```

I think I vote for (1): doesn't really feel like there's "real" ambiguity being introduced in this case, even for exit_code.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1312,6 +1323,57 @@ message OutputSymlink {
...
+    // code. Functions such as wait() and waitpid() tend to truncate the
+    // exit code to the bottom eight bits.
+    int64 exit_code = 1;
+  }
+
+  message Signaled {
+    // The name of the signal that was delivered that caused the process
+    // to terminate, without the "SIG" prefix (e.g., "BUS", "KILL",
+    // "SEGV").
+    string signal_name = 1;
```
bduffany, line 1350: Is there a rationale behind choosing a signal name vs. a signal number? (I don't have much of a preference, just curious)

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1312,6 +1323,57 @@ message OutputSymlink {
...
+    //
+    // If this field is set, [ActionResult.exit_code][build.bazel.remote.execution.v2.ActionResult.exit_code]
+    // MUST have the same value as [Exited.exit_code][[build.bazel.remote.execution.v2.ProcessTerminationResult.Exited.exit_code],
+    // clamped to the range of int32.
+    Exited exited = 1;
+
+    // Abnormal process termination occurred.
+    //
+    // On POSIX-like systems, this is achieved by calling abort(), or by
+    // receiving a signal for which no signal handler is installed.
```
bduffany, line 1369: nit: SIGKILL can't be handled, so maybe this wording is slightly clearer:

```suggestion
    // receiving a signal that is not handled by the process.
```

bduffany (commented):
+1 for this change; our impl currently returns `-1` for exit code in this case, mostly as a consequence of using `exec.Command` in Golang which returns `-1` whenever the process doesn't `exit()`. It would be nice to surface more details as to whether it's an `abort()` or something else.

PR #248, opened 2023-03-15
Add value for BLAKE3 hash function

tylerwilliams:
Add BLAKE3 hash function.

### DIFF
--- a/build/bazel/remote/execution/v2/remote_execution.proto
+++ b/build/bazel/remote/execution/v2/remote_execution.proto
@@ -1885,6 +1885,10 @@ message DigestFunction {
     // Test vectors of this digest function can be found in the
     // accompanying sha256tree_test_vectors.txt file.
     SHA256TREE = 8;
+
+    // The BLAKE3 hash function.
+    // See https://github.com/BLAKE3-team/BLAKE3.
+    BLAKE3 = 9;
   }
 }
 


tylerwilliams:
@EricBurnett mind taking a look at this? Thank you!

EricBurnett:
LGTM!

EdSchouten:
@tylerwilliams I was wondering, what's the reason you folks are pursuing using BLAKE3 and not the recently added SHA256TREE digest function (#235)? Sure, at its peak BLAKE3 is faster than SHA256TREE, but the difference isn't that big in the grand scheme of things, right? SHA256TREE has some added advantages:

- Performance is still adequate on other CPU architectures that provide SHA-256 CPU extensions. For example, BLAKE3 only has 1/10th the throughput of SHA-256 and SHA256TREE on ARM64.
- SHA256TREE is built with chunking in mind, allowing you to decompose large CAS objects into smaller ones. BLAKE3 makes this hard, because of their use of Chunk Counters.

I mean, I'm fine with adding support for BLAKE3, but I'm merely interested in seeing whether we can prevent Balkanization in this area.

tylerwilliams:
@EdSchouten -- Definitely agree re balkanization and I'm very excited to see SHA256TREE land for chunking support. We will likely support this too when it lands in Bazel :) 

That said, BLAKE3 is incredibly fast and this is useful when hashing large files, which we do a lot of (docker images, snapshots, large binaries, etc) on CPUs with AVX* support. I don't think it's a bad thing to support this hash in the protocol, or in Bazel, so that these kinds of blobs can be hashed quickly today.

PR #256, opened 2023-05-26
Add justbuild to the list of clients and servers

oreiche:
It [officially](https://github.com/just-buildsystem/justbuild/blob/master/share/man/just.1.org#compatibility-options) supports the protocol in its client and server implementations.


EdSchouten (approved):

mostynb:
I wonder if it's worth mentioning the need for the `–compatible` flag here? Just users might find this confusing otherwise.

oreiche:
> I wonder if it's worth mentioning the need for the `–compatible` flag here? Just users might find this confusing otherwise.

Agreed, I added a note in parentheses to both list entries.

bergsieker (approved):

PR #260, opened 2023-06-09
Add BLAKE3 to Go code

sluongng:
This is a follow up to #248 to ensure the git hook is running correctly.


EdSchouten (approved):
Fine by me. But feel free to collapse it into #261.

sluongng:
I think this is easier to merge first as it's a simple fixup.

#261 could potentially trigger some discussions, so I will keep it separate.

bergsieker (approved):

PR #276, opened 2023-10-13
remote_exec.proto: support ZSTD with dictionary

### LINKED ISSUE #272: Support compression with external dictionary
Both Brotli and ZSTD support speeding up compression with a set of external dictionaries, trained from sample data.

It would be nice if our API could be made aware that (a) there is a dictionary involved in the compression and (b) how to upload/download that dictionary from/to client to/from server.

sluongng:
Closes #272


sluongng:
cc: @mostynb @werkt @EdSchouten

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1961,9 +1961,30 @@ message Compressor {
...
+  // Each file respresent a single Zstandard dictionary
+  // with name being the `dictId` that was added to the header
+  // of the compressed file.
+  //
+  // A special file with name `default` is used to represent the
+  // default dictionary that is used when no `dictId` is specified.
+  // Clients SHOULD prefer to use the default dictionary when possible.
+  //
+  // The `default` dictionary MUST be present.
+  repeated FileNode dictionaries = 1;
```
EdSchouten, line 1985: As I mentioned during our monthly meeting, I don't see why we must shoehorn this into `Directory`/`FileNode`/... related messages. We're trying to store properties about compression dictionaries. Not a file system hierarchy. There is never going to be a scenario where a build action wants to call `ls` against a set of compressor dictionaries. Dictionaries also don't need an `is_executable` or `node_properties`.

Why permit the use case where no dictionary ID is present? Isn't that just encouraging our users to paint themselves into a corner? For what is a brand new feature I see no reason to be so liberal here.

Please do something like this instead.
```proto
message ZstdDictionaryRegistry {
  // The dictionary ID that clients SHOULD use when uploading
  // objects into the Content Addressable Storage. An entry for
  // this dictionary ID MUST be present in `dictionaries`.
  uint32 upload_dictionary_id = 1;

  // Dictionaries that may be used by the server when returning
  // objects stored in the Content Addressable Storage. The key
  // corresponds to a dictionary ID, as described in RFC 8878,
  // section 3.1.1.1.3, while the value refers to an dictionary
  // as described in RFC 8878, chapter 6.
  map<uint32, Digest> dictionaries = 2;
}
```

A downside of `map<X, Y>` is that Protobuf libraries tend to be inconsistent in the order in which they get marshalled. I can imagine that may be annoying for us, as they get stored in a CAS. If that's a concern, we can use a `repeated` field instead.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1961,9 +1961,30 @@ message Compressor {
...
+    //
+    // When this is used, the server MUST advertise the dictionaries by
+    // including
+    // [ZstdDictionaryRegistry][build.bazel.remote.execution.v2.ZstdDictionaryRegistry]
+    // digest in CacheCapabilities.
+    ZSTD_DICT = 4;
   }
 }
 
+message ZstdDictionaryRegistry {
```
EdSchouten, line 1975: Any specific reason we have "Registry" part of the message & digest field name? I'd just call this "ZstdDictionaries" or "ZstandardDictionaries".

build/bazel/remote/execution/v2/remote_execution.proto
mostynb, line 1985: FileNode has a bunch of unrelated and unspecified fields.

I wonder if it would be simpler to have a repeated Digest field, and a separate "Default" field which is an index into the array/slice/whatever of the digests? Or maybe just a separate "Default" digest?

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1961,9 +1961,30 @@ message Compressor {
 
     // Brotli compression.
     BROTLI = 3;
+
+    // Zstandard compression with dictionary.
```
mostynb, line 1965: Is the dictionary optional?

If a server supports this mode, should we also require that it support ZSTD (ie without dictionary)?

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1961,9 +1961,30 @@ message Compressor {
...
+    //
```
mostynb, line 1966: Should we provide advice on which dictionary to use for a given blob, and which not to use a dictionary for at all? eg maybe large blobs over a certain size aren't worth using a dictionary for. And maybe we could use file type specific dictionaries?

build/bazel/remote/execution/v2/remote_execution.proto
mostynb: I think I would prefer a repeated field and an index-into-that field.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1997,6 +2018,11 @@ message CacheCapabilities {
   // [BatchUpdateBlobs][build.bazel.remote.execution.v2.ContentAddressableStorage.BatchUpdateBlobs]
   // requests.
   repeated Compressor.Value supported_batch_update_compressors = 7;
+
+  // The digest of the
+  // [ZstdDictionaryRegistry][build.bazel.remote.execution.v2.ZstdDictionaryRegistry]
+  // that contains all the dictionaries supported by the remote cache.
+  Digest zstd_dictionary_registry = 8;
```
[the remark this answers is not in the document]
sluongng: Oh this is brilliant!
QualifiedDigest is exactly what I wanted.

Will add this into the next iteration.

build/bazel/remote/execution/v2/remote_execution.proto
sluongng: I was fixated on FileNode initially as I want to give more flexibility in naming the dictionary for other compression algorithms. However since then, I have decided to make this specifically for Zstandard dictionary compression so I think using `uint32` is a great suggestion here.

```proto
message ZstdDictionaryRegistry {
  message Dictionary {
    uint32 id 1;
    Digest digest 2;
  }

  // The dictionary ID that clients SHOULD use when uploading
  // objects into the Content Addressable Storage. An entry for
  // this dictionary ID MUST be present in `dictionaries`.
  uint32 default_dictionary_id = 1;

  // Dictionaries that may be used by the server when returning
  // objects stored in the Content Addressable Storage. The key
  // corresponds to a dictionary ID, as described in RFC 8878,
  // section 5, while the value refers to a dictionary
  // as described in RFC 8878, chapter 6.
  repeated Dictionary dictionaries = 2;
}
```

Would something like this make sense?

Note: I do want to `s/upload/default/` to make sure that we generalize the use of a default dictionary for unforeseeable use cases.

I also think the dictionary ID is described in section 5, not 3.1.1.3 https://datatracker.ietf.org/doc/html/rfc8878#name-dictionary-format

build/bazel/remote/execution/v2/remote_execution.proto
sluongng: I was unsure about using plural forms, so I sidestepped with `registry` which has been relatively popular in the naming convention recently.

I don't mind going back to ZstdDictionaries.

build/bazel/remote/execution/v2/remote_execution.proto
sluongng: Currently, I prefer to think of this as a separate compressor and not related to ZSTD at all. It helps simplify the change I am proposing. Using a mix of ZSTD and ZSTD_DICT is a non-goal for me at this moment.

build/bazel/remote/execution/v2/remote_execution.proto
sluongng: We could discuss the mixed usage of different compressors separately in another change, but at least for now, I would prefer to treat this the same way as treating `ZSTD vs Brotli`. Let's leave the decision to the actual implementations to decide.

I think once this is implemented by a set of clients and servers, we will have a better idea regarding performance implications, as well as whether a mixed use of compressors is preferred in certain scenarios.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1997,6 +2018,11 @@ message CacheCapabilities {
   // [BatchUpdateBlobs][build.bazel.remote.execution.v2.ContentAddressableStorage.BatchUpdateBlobs]
   // requests.
   repeated Compressor.Value supported_batch_update_compressors = 7;
+
+  // The digest of the
```
EdSchouten, line 2022: Giving this part some more thought: you may also need to extend GetCapabilitiesRequest to contain a list of supported digest functions. Otherwise this field may return a digest that is incompatible with the client. There need to be some negotiation here.

sluongng: Then it might need to be a combination of Digest and Compressor. Because the dictionary registry blob could also be compressed with a non-dictionary compressor right? 🤔

EdSchouten: There is no need to account for that, as it's always possible to fetch blobs with the IDENTITY compressor. I think that most clients would actually do that when fetching a dictionary. With digest functions it's a different story, as a client may not actually be able to construct a bytestream URL if it does not provide support for a given digest function.

sluongng: If Digest functions are unavailable in GetCapRequest, we should assume that it's SHA256 right?

EdSchouten: No. There is nothing in the Remote Execution protocol right now that gives SHA-256 a special status over the others.

sluongng: On one hand, I do agree that some negotiation seems to be needed here for the client/server to agree on the digest. On the other hand, I am a bit afraid of the scope creep of having to extend GetCapabilitiesRequest. For example: if the client were to only support `SHA256` digest function, should the server still return other digest functions in CacheCapabilities? What if the server does not support SHA256? should it just err out?

I guess we could just left it undecided in the spec?

EdSchouten: > On the other hand, I am a bit afraid of the scope creep of having to extend GetCapabilitiesRequest. For example: if the client were to only support SHA256 digest function, should the server still return other digest functions in CacheCapabilities?

It still may. We're not going to adjust anything there. The digest functions would only be used to control the format of digests returned as part of GetCapabilities(). Not any of the other fields.

Another option would be to simply add a dedicated call for negotiating compression dictionaries, as opposed to having it part of GetCapabilities().

> What if the server does not support SHA256? should it just err out?

You mean if the client only announces support for SHA-256 and the server does not support it? Then it has no way to return the dictionary in that format. So in that case it may return a capabilities response that does not announce the support for ZSTD_DICT.

sluongng:
This PR is/will be a bit stale because I am a bit occupied with other tasks for the time being.

I could revisit this if there are more demands for such capability, or if somebody else wants to pick up the flag, please feel free to.

sluongng:
Since we might have some use case for this via https://github.com/bazelbuild/bazel/issues/18997, I have pushed a new version of this PR according to review feedbacks.

As for @EdSchouten suggestion about enhancing `GetCapabilitiesRequest` for better negotiation. I think that should be done in a separate PR from this one? Wdyt?

EdSchouten:
> As for @EdSchouten suggestion about enhancing `GetCapabilitiesRequest` for better negotiation. I think that should be done in a separate PR from this one? Wdyt?

Why can't we just do it right as part of this PR? What's preventing us from doing that?

If we don't add full negotiation to `GetCapabilities` itself, what are your thoughts on adding a dedicated RPC method to `ContentAddressableStorage`? For example, a `GetCompressionDictionaries()`.

sluongng:
> Why can't we just do it right as part of this PR? What's preventing us from doing that?

Initially, I thought it could make it harder to get this merge. But I looked at it and found it to be less disruptive than I initially thought. So I pushed a commit that does just that, the language could use some help though 🙏 .

> what are your thoughts on adding a dedicated RPC method to ContentAddressableStorage? For example, a GetCompressionDictionaries().

I think it would still require us to pass a minimum boolean value in `ServerCapabilities` to indicate that the server does support that RPC.
And then, the additional RPC simply returns what could be a simple CAS entry... so a glorified ByteStream.Read call.

So instead of sending the boolean value in ServerCapabilties, let's just send the digest and use the existing ByteStream.Read RPC instead.

The caveat is ofc, the server would have to store that CAS value somewhere, which could potentially limit how fast the server could swap out the dictionaries. Realistically, I don't expect the dictionary list to change too frequently for most use cases as it will change the delta calculation, leading to more frequent cache invalidation.

Lastly, I think we could reserve the dedicated RPC for when the dictionary negotiation gets more complicated. But for now, this would do.

EdSchouten:
No, it doesn't. The presence of Zstd with

sluongng:
Discussed with @EdSchouten offline. As some server/client/proxy are caching the `ServerCapabilities` based on the `GetCapabilitiesRequest`, adding more fields to `GetCapabilitiesRequest` could further complicate the existing caching mechanism.

Sent a change to move the Dictionary fetching to a dedicated RPC in hope that this change has minimal impact on existing implementations. I still want to make this rpc ZSTD specific, but if there is a strong opinion on making it generic for future compressors, suggestions are welcome.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1997,6 +2037,16 @@ message Compressor {
 
     // Brotli compression.
     BROTLI = 3;
+
+    // Zstandard compression with dictionary.
+    //
+    // Servers which support this compressor MUST implements
```
EdSchouten, line 2043: s/implements/implement/

EdSchouten (approved):

build/bazel/remote/execution/v2/remote_execution.proto
sluongng: fixed

EdSchouten: I still see this in the latest version of the PR. Sure you pushed the latest version?

sluongng: Sorry, I pushed it to a dedicated ref on my fork remote. Did not realize this PR was based on the `main` branch of my fork.

I re-pushed the latest version

build/bazel/remote/execution/v2/remote_execution.proto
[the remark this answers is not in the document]
bduffany: IMO, this is not the right time to introduce `QualifiedDigest`. For better or for worse, the existing convention is to make `DigestFunction.Value` a sibling field of `Digest`. Without concrete plans to migrate the other existing usages, I think this will add unnecessary inconsistency to the API, introduce a new "busywork" task of everyone having to migrate from Digest => QualifiedDigest, and cause confusion for both clients and servers (e.g. "why do we sometimes use QualifiedDigest and other times we use a regular Digest?").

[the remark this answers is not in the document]
bduffany: NVM, I see that this is outdated. How embarrassing :sweat_smile:

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1814,6 +1821,39 @@ message GetTreeResponse {
...
 
+message GetZstdDictionariesRequest {
+  // The instance of the execution system to operate against. A server may
+  // support multiple instances of the execution system (with their own workers,
+  // storage, caches, etc.). The server MAY require use of this field to select
+  // between them in an implementation-defined fashion, otherwise it can be
+  // omitted.
+  string instance_name = 1;
+
+  // The digest functions supported by the client.
```
bduffany, line 1832:
```suggestion
  // The digest function in use by the client, which will determine the
  // digest function for all Digests in the response.
```

bduffany:
> I still want to make this rpc ZSTD specific, but if there is a strong opinion on making it generic for future compressors, suggestions are welcome.

Seems to me that compression dictionaries are simple enough where they wouldn't vary much across compressor types, so if it were compressor-specific then there would eventually wind up being a bunch of different RPCs with the same Request/Response signature? (I don't feel super strongly either way, just something to consider)

EdSchouten (approved):

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1814,6 +1821,40 @@ message GetTreeResponse {
...
+  // The digest function in use by the client, which determines the
+  // digest function for all Digests in the response.
+  // Must be one of the server's supported digest functions listed in
+  // [CacheCapabilities.digest_functions][build.bazel.remote.execution.v2.CacheCapabilities.digest_functions].
+  DigestFunction.Value digest_function = 2;
+}
+
+message GetZstdDictionariesResponse {
+  message Dictionary {
+    uint32 id = 1;
```
bergsieker, line 1841: It seems like we could make dictionaries reasonably generic by replacing this uint32 with an Any. That wouldn't account for session-specific dictionaries, but it should allow for ~any static dictionary. The unpacking of the Any, and understanding what blobs it applies to, is very much compressor-specific.

sluongng: Since this is an "id" specific field, I would say we should use `string` instead of `Any` at the minimum. `Any` delegates the responsibility down to implementations to validate the data format which could be a pain to work with.

If generic is a requirement here, I think something like this could be a good alternative

```protobuf
message GetDictionariesResponse {
  message DefaultDictionaryId {
    oneof id {
      uint32 zstd_dictionary_id = 1;
    }
  }
  // The dictionary ID that clients SHOULD use when uploading
  // objects into the Content Addressable Storage. An entry for
  // this dictionary ID MUST be present in `dictionaries`.
  DefaultDictionaryId default_dictionary_id = 1;

  message ZstdDictionary {
    uint32 id = 1;
    Digest digest = 2;
  }
  message ZstdDictionaries {
    repeated ZstdDictionaries zstd_dictionaries = 1;
  }
  message Dictionaries {
    oneof dictionaries {
      ZstdDictionaries zstd_dictionaries = 1;
    }
  }

  // Dictionaries that may be used by the server when returning
  // objects stored in the Content Addressable Storage. The key
  // corresponds to a dictionary ID, as described in RFC 8878,
  // section 5, while the value refers to a dictionary
  // as described in RFC 8878, chapter 6.
  Dictionaries dictionaries = 2;
}
```

This would enable future Dictionary types (if there were any) could be added incrementally. WDYT?

bergsieker: I would also be fine with a string, but given the conversation yesterday regarding people wanting more structured types to avoid the need to shove json into string fields, thought that Any might be appropriate. Using a string for an ID field would be more consistent with the rest of the API. In any case, I don't think that using string vs Any absolves the client from the need to validate the input. In this case of zstd, for example, the client will need to parse the string into an int ID to match with the ID encoded in the compressed blob.

At some level what I'm trying to avoid is needing to modify this API with each dictionary-compression function that we add, which means trying to remove anything zstd-specific. It's not realistic to design an API that will work in *all* cases without investing significantly in understanding, but there are only so many realistic ways to get dictionary-based compression to work, and I have to believe that "encode a dictionary ID in the compressed blob" is reasonably common.

EdSchouten: > Since this is an "id" specific field, I would say we should use string instead of Any at the minimum. Any delegates the responsibility down to implementations to validate the data format which could be a pain to work with.

But doesn't a string also require that an implementation validates the data format? There are also strings that are not valid integers.

PR #278, opened 2023-10-16
Add max_cas_blob_size_bytes to CacheCapabilities message

diegohavenstein:
We need this field to improve the performance of Bazel when using remote cache, since by doing this we can determine the CAS entry upper limit size, thus skipping uploads that would fail due to size constraints.

The issue on Bazel side is https://github.com/bazelbuild/bazel/issues/18449


build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1997,6 +1997,10 @@ message CacheCapabilities {
   // [BatchUpdateBlobs][build.bazel.remote.execution.v2.ContentAddressableStorage.BatchUpdateBlobs]
   // requests.
   repeated Compressor.Value supported_batch_update_compressors = 7;
+
+  // Maximum size a single CAS entry can have in the remote cache system.
+  // A value of 0 means no limit is set
+  int64 max_cas_entry_size_bytes = 8;
```
tjgq, line 2003: I'd like to see something here describing how implementations should interpret this limit:

1. SHOULD implementations not attempt to upload a blob larger than this? MUST they not? (note the latter is a breaking change and would require a version bump)
2. MAY implementations return a blob larger than this size?

diegohavenstein: Good points

I'm for hint semantics for the limit here, i.e.
1. Implementations SHOULD not attempt to upload blobs over the threshold. If they do, it is implementation defined what happens. E.g. a remote cache server may accept the request anyway, or reject it as invalid.
2. Implementations MAY return blobs larger than the threshold. The threshold may be set after the CAS in use was created and have larger blobs, in which case we don't want to enforce skipping those blobs.
3. The value 0 means the server implementations do not have special requirements when it comes to the size of CAS entries.

diegohavenstein: @tjgq do you agree with the new description?
@bergsieker any issues from your side?

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1997,6 +1997,22 @@ message CacheCapabilities {
...
+  //   blobs with size larger than the limit. If they do, it is up to the
+  //   cache implementation what to do, e.g. accept the request anyway or reject
+  //   it as invalid.
+  //   If the cache implementation return a given limit, it *may* still serve
+  //   blobs larger than this limit. This is again implementation defined behaviour.
+  //
+  // Overall, treat this limit as a hint from the cache. The limit may be set for
+  // performance reasons mainly, since when limit is known beforehand some upload
+  // requests can then be skipped completely and avoid roundtrips.
+  int64 max_cas_entry_size_bytes = 8;
```
EdSchouten, line 2015: s/entry/object/ maybe?

Looking at remote_execution.proto, it looks like 'entry' or 'cache entry' mainly refers to 'Action Cache entries'. For things in the CAS, we frequently seem to call them 'blobs' or 'objects'.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1997,6 +1997,22 @@ message CacheCapabilities {
...
+  //   may be uploaded to and downloaded from the cache.
+  // - If it is larger than 0, implementations *should* not attempt to upload
+  //   blobs with size larger than the limit. If they do, it is up to the
+  //   cache implementation what to do, e.g. accept the request anyway or reject
+  //   it as invalid.
+  //   If the cache implementation return a given limit, it *may* still serve
+  //   blobs larger than this limit. This is again implementation defined behaviour.
+  //
+  // Overall, treat this limit as a hint from the cache. The limit may be set for
+  // performance reasons mainly, since when limit is known beforehand some upload
```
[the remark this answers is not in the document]
diegohavenstein: Dropped

build/bazel/remote/execution/v2/remote_execution.proto
diegohavenstein: I'm fine with blob. Changing it

EdSchouten (approved):
Though I approve of this change, I do think we're still leaving some things undefined that are worth specifying.

Notably: what's the maximum individual/combined size of Tree messages UpdateActionResult()/GetActionResult() is willing to tolerate to perform completeness checking on? Relatedly, what's the maximum size of a Directory message that may be contained in a Tree?

Buildbarn has the ability to limit these through its `maximumTotalTreeSizeBytes` and `maximumMessageSizeBytes` options.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1997,6 +1997,18 @@ message CacheCapabilities {
   // [BatchUpdateBlobs][build.bazel.remote.execution.v2.ContentAddressableStorage.BatchUpdateBlobs]
   // requests.
   repeated Compressor.Value supported_batch_update_compressors = 7;
+
+  // Maximum size a single CAS blob can have in the remote cache system.
+  // Here is how implementations should interpret this value:
+  // - If it is 0, it means there is no limit is set. Arbitrarily large blobs
```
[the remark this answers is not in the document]
diegohavenstein: agree

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1997,6 +1997,18 @@ message CacheCapabilities {
...
+  //   may be uploaded to and downloaded from the cache.
+  // - If it is larger than 0, implementations SHOULD not attempt to upload
```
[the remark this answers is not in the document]
diegohavenstein: thx for the link. Fixed

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1997,6 +1997,18 @@ message CacheCapabilities {
   // [BatchUpdateBlobs][build.bazel.remote.execution.v2.ContentAddressableStorage.BatchUpdateBlobs]
   // requests.
   repeated Compressor.Value supported_batch_update_compressors = 7;
+
+  // Maximum size a single CAS blob can have in the remote cache system.
+  // Here is how implementations should interpret this value:
```
tjgq, line 2002: Nit: this sentence doesn't add any information; it's clear that the remainder describes how to interpret the value.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1997,6 +1997,18 @@ message CacheCapabilities {
...
+  // - If it is 0, it means there is no limit is set. Arbitrarily large blobs
+  //   may be uploaded to and downloaded from the cache.
+  // - If it is larger than 0, implementations SHOULD not attempt to upload
+  //   blobs with size larger than the limit. If they do, it is up to the
+  //   cache implementation what to do, e.g. accept the request anyway or reject
+  //   it with INVALID_ARGUMENT.
+  //   If the cache implementation return a given limit, it MAY still serve
+  //   blobs larger than this limit. This is again implementation defined behaviour.
```
tjgq, line 2010: "This is again implementation defined behaviour" doesn't add any information.

build/bazel/remote/execution/v2/remote_execution.proto
tjgq: LGTM from the Bazel side, but I'd suggest letting it linger for a week or so, to give other implementors a chance to comment.

diegohavenstein:
> Though I approve of this change, I do think we're still leaving some things undefined that are worth specifying.
> 
> Notably: what's the maximum individual/combined size of Tree messages UpdateActionResult()/GetActionResult() is willing to tolerate to perform completeness checking on? Relatedly, what's the maximum size of a Directory message that may be contained in a Tree?
> 
> Buildbarn has the ability to limit these through its `maximumTotalTreeSizeBytes` and `maximumMessageSizeBytes` options.

(Sorry for my lack of deep knowledge in this area)
Does it make sense to have fine-granular settings for these? Not sure how they exactly relate to max blob size in the CAS
Are you referring to UpdateActionResult()/GetActionResult() in ActionCacheServer.java?

EdSchouten:
> Are you referring to UpdateActionResult()/GetActionResult() in ActionCacheServer.java?

Is that ActionCacheServer.java in bazel-buildfarm? If so... yes?

GetActionResult() is only permitted to return ActionResults if all of the output files it references exist. ActionResults may also refer to output directories (Tree objects), which in its turn point to files. If `max_cas_blob_size_bytes` is set to 100 TB, that still doesn't mean that it's okay to upload a 100 TB Tree object and provide it to a call to `UpdateActionResult()`. There would be no way for a remote cache to validate it in a reasonable amount of time/space.

bergsieker:
> > Are you referring to UpdateActionResult()/GetActionResult() in ActionCacheServer.java?
> 
> Is that ActionCacheServer.java in bazel-buildfarm? If so... yes?
> 
> GetActionResult() is only permitted to return ActionResults if all of the output files it references exist. ActionResults may also refer to output directories (Tree objects), which in its turn point to files. If `max_cas_blob_size_bytes` is set to 100 TB, that still doesn't mean that it's okay to upload a 100 TB Tree object and provide it to a call to `UpdateActionResult()`. There would be no way for a remote cache to validate it in a reasonable amount of time/space.

I agree that there are more types of limits that could be specified here--total blobs in a tree, total tree depth, max files/subdirectories in a level, etc. However, I think this limit stands alone, provides reasonable value, and is easy to understand. I think committing this on its own is good, and we can have a separate conversation about whether specifying additional limits would be valuable.

sluongng:
Is it worth introducing a new ServerLimits/ServerConstraints message that contains various different limits to be introduced in the future?

diegohavenstein:
> Is it worth introducing a new ServerLimits/ServerConstraints message that contains various different limits to be introduced in the future?

That sounds like a good idea to keep CacheCapabilities stable, yes. I'll push a patch

diegohavenstein:
> Is it worth introducing a new ServerLimits/ServerConstraints message that contains various different limits to be introduced in the future?

I had a look again, I don't think it's good in this context, because many things in the CacheCapabilities could arguably be in the new proposed message (e.g. max_batch_total_size_bytes or supported_compressors), since they describe limits/constraints. As such I think the max_cas_blob_size_bytes is in the right place as of now.

Do you agree @sluongng ?

sluongng:
Sound reasonable to me. Thanks for investigating! 🤗

diegohavenstein:
> > > Are you referring to UpdateActionResult()/GetActionResult() in ActionCacheServer.java?
> > 
> > 
> > Is that ActionCacheServer.java in bazel-buildfarm? If so... yes?
> > GetActionResult() is only permitted to return ActionResults if all of the output files it references exist. ActionResults may also refer to output directories (Tree objects), which in its turn point to files. If `max_cas_blob_size_bytes` is set to 100 TB, that still doesn't mean that it's okay to upload a 100 TB Tree object and provide it to a call to `UpdateActionResult()`. There would be no way for a remote cache to validate it in a reasonable amount of time/space.
> 
> I agree that there are more types of limits that could be specified here--total blobs in a tree, total tree depth, max files/subdirectories in a level, etc. However, I think this limit stands alone, provides reasonable value, and is easy to understand. I think committing this on its own is good, and we can have a separate conversation about whether specifying additional limits would be valuable.

I'm good to ship this as-is. What is the process? I can't merge commits here it seems @bergsieker

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1997,6 +1997,17 @@ message CacheCapabilities {
   // [BatchUpdateBlobs][build.bazel.remote.execution.v2.ContentAddressableStorage.BatchUpdateBlobs]
   // requests.
   repeated Compressor.Value supported_batch_update_compressors = 7;
+
+  // Maximum size a single CAS blob can have in the remote cache system.
+  // - If it is 0, it means there is no limit is set. Arbitrarily large blobs
+  //   may be uploaded to and downloaded from the cache.
+  // - If it is larger than 0, implementations SHOULD not attempt to upload
+  //   blobs with size larger than the limit. If they do, it is up to the
+  //   cache implementation what to do, e.g. accept the request anyway or reject
```
bergsieker, line 2006: I'm wondering if we should actually tighten this up to improve the possibility of interop between implementations. Being permissive sounds nice in theory, but if you're using a system that advertises limit X (but actually accepts larger blobs), and move to another system that advertises the same limit (and enforces it), you'd get surprising failures.

(Note that the carve-out below for *serving* larger blobs is still important, for dealing with legacy conditions.)

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1997,6 +1997,17 @@ message CacheCapabilities {
...
+  //   it with INVALID_ARGUMENT.
```
mostynb, line 2007:
```suggestion
  //   blobs with size larger than the limit. Servers SHOULD reject blob
  //   uploads over the `max_cas_blob_size_bytes` limit with response code
  //   `INVALID_ARGUMENT`.
```

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1997,6 +1997,17 @@ message CacheCapabilities {
...
+  //   If the cache implementation return a given limit, it MAY still serve
```
mostynb, line 2008:
```suggestion
  //   If the cache implementation returns a given limit, it MAY still serve
```

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1997,6 +1997,17 @@ message CacheCapabilities {
   // [BatchUpdateBlobs][build.bazel.remote.execution.v2.ContentAddressableStorage.BatchUpdateBlobs]
   // requests.
   repeated Compressor.Value supported_batch_update_compressors = 7;
+
+  // Maximum size a single CAS blob can have in the remote cache system.
+  // - If it is 0, it means there is no limit is set. Arbitrarily large blobs
+  //   may be uploaded to and downloaded from the cache.
+  // - If it is larger than 0, implementations SHOULD not attempt to upload
```
mostynb, line 2004:
```suggestion
  // - If it is larger than 0, implementations SHOULD NOT attempt to upload
```

axeluhlig:
@diegohavenstein left our team to help cats and dogs in need: https://palou.app/en/ 
I'll continue his work here to help overloaded remote endpoints in need :-) 
@mostynb Appreciate any help. I'll push a commit addressing your comments

axeluhlig:
Here is a new version addressing the review comments: https://github.com/axeluhlig/remote-apis/tree/issue-275-cont

It's also rebased onto upstream `main` and I re-generated the `remote_execution.pb.go` file. Are there any other required steps?

Given that we - to my knowledge - won't be able to change the source / head of this PR I created a new one: https://github.com/bazelbuild/remote-apis/pull/328

dhavenstein:
Thx for taking it over @axeluhlig !
I'll close the issue now

PR #280, opened 2023-11-15
Add trexe and buildbox-casd to the list of RE clients/servers

juergbi:


README.md
```diff
@@ -50,6 +50,7 @@ These tools use the Remote Execution API to distribute builds to workers.
 * [Recc](https://gitlab.com/bloomberg/recc)
 * [Reclient](https://github.com/bazelbuild/reclient)
 * [Siso](https://chromium.googlesource.com/infra/infra/+/refs/heads/main/go/src/infra/build/siso/)
+* [trexe](https://gitlab.com/BuildGrid/trexe) (open source)
```
sluongng, line 53: Should we remove the `(open source)` part here? because it would imply that the other clients are not open source (which they are).

README.md
```diff
@@ -58,6 +59,7 @@ workers implement the Remote Worker API.
 
 * [bazel-remote](https://github.com/buchgr/bazel-remote) (open source, cache only)
 * [Buildbarn](https://github.com/buildbarn) (open source)
+* [buildbox-casd](https://gitlab.com/BuildGrid/buildbox/buildbox-casd) (open source)
 * [BuildBuddy](https://www.buildbuddy.io/) (commercial & open source)
 * [Buildfarm](https://github.com/bazelbuild/bazel-buildfarm) (open source)
 * [BuildGrid](https://buildgrid.build/) (open source)
```
sluongng, line 65: Could you help elaborate how to differentiate `buildbox-casd` vs BuildGrid? Aren't they the same project?

README.md
juergbi: Makes sense, changed.

README.md
juergbi: buildbox-casd is a caching proxy, designed to be mainly used locally on worker machines and client systems. It provides additional functionality that only makes sense for a local server such as integration with buildbox-fuse (used by buildbox-run for local execution).

The proxy function is optional, which means that it can theoretically also be deployed as server for remote clients. This can be useful for tiny deployments. However, as it only supports local filesystem storage, this is not a real alternative for larger deployments that BuildGrid and other servers support.

It supports proxying of Execute requests and should soon also support local sandboxed execution without a remote execution service.

sluongng: SGTM. Thanks!

EdSchouten:
Is it really worth calling out all of these projects separately? For example, for Buildbarn we also don't list tools like bb_clientd separately.

juergbi:
> Is it really worth calling out all of these projects separately? For example, for Buildbarn we also don't list tools like bb_clientd separately.
> 
> Don't they just fall below the Buildgrid umbrella?

trexe is an REAPI client (written in C++) that is not tied to the BuildGrid server (written in Python) at all, except that both are mainly developed by the same team. If someone is interested to explore what REAPI clients are available, it makes sense to find trexe in the list. I would say BuildGrid and trexe technically don't have more in common than e.g. Buildfarm and Reclient (both in the same GitHub org).

buildbox-casd may be a slightly less clear case as it is commonly used together with other REAPI software that is already on the list (BuildStream, recc, BuildGrid). However, it also does not depend on BuildGrid at all and it is actively being used without BuildGrid (for local-only operation of BuildStream and recc, and as simple remote cache server).

I don't think it makes sense to list the other BuildBox components such as buildbox-worker, buildbox-run and buildbox-fuse as they are exclusively used as dependencies of BuildGrid, BuildStream, and buildbox-casd. However, I don't see a reason to omit trexe and I think it makes sense to list buildbox-casd as well.

sluongng:
How about turning https://buildgrid.build/ into the landing page that introduces these sub-projects?

The reason is: At BuildBuddy, our `bb` CLI (1) could also be used as a standalone RE-API client.
We support `bb upload`, `bb download` and `bb execute`.

```
> bb execute --help

usage: bb execute [ options ... ] -- <executable> [ args ... ]

Runs a remote execution request against a remote execution service backend
using a given command as input.

Args that modify execution can be placed before '--', and the command executable
and arguments should come afterwards.

Example of running a simple bash command:
  $ bb execute -- bash -c 'echo "Hello world!"'

Example of running a bash command with runner recycling:
  $ bb execute --exec_properties=recycle-runner=true -- bash -c 'echo "Runner uptime:" $(uptime)'

All options:
  -action_env NAME=VALUE
        Action environment variable, as a NAME=VALUE pair. Can be specified more than once.
  -digest_function "sha256" or "blake3"
        Digest function used for content-addressable storage. Can be "sha256" or "blake3". (default "sha256")
  -exec_properties NAME=VALUE
        Platform exec property, as a NAME=VALUE pair. Can be specified more than once.
  -input_root string
        Input root directory. By default, the action will have no inputs.
  -invocation_id string
        If set, set this value as the tool_invocation_id in RequestMetadata.
  -remote_executor string
        Remote execution service target. (default "grpcs://remote.buildbuddy.io")
  -remote_header NAME=VALUE
        Header to be applied to all outgoing gRPC requests, as a NAME=VALUE pair. Can be specified more than once.
  -remote_instance_name string
        Value to pass as an instance_name in the remote execution API.
  -remote_timeout duration
        Timeout used for the action. (default 1h0m0s)
```

These could be used against other grpc RBE backends and not just ours.
So if we start to go down this road, we would do a follow-up to add our CLI to this list as well and I'm sure other projects will start to do the same thing. At that point, I think it will make new users, who rely on this doc to navigate the ecosystem, very confused.

Another angle I could see is that we could create some sort of grouping for these entries with sub-bullet points.

(1): https://www.buildbuddy.io/docs/cli

juergbi:
The grouping in the README into clients and servers doesn't seem ideal if individual projects include both clients and servers. However, we're planning to proceed with improvements to https://buildgrid.build/ for now, closing this.

Issue #178, opened 2020-10-16
V3 idea: Drop support for offset reads & writes. Decompose large CAS objects into shallow Merkle trees


EdSchouten:
The size of objects stored in the CAS has a pretty large spread. We can have tiny Directory objects that are less than 100 bytes in size, but we can also have output files that are gigabytes in size. Dealing with these large files is annoying:

- When using the digest as a key for sharding the CAS across multiple backends, you notice that you get an almost perfect distribution in terms of request count. Below is an example read count, 1h average, of a setup that has eight shards:

<img width="1284" alt="Screenshot 2020-10-16 at 13 43 58" src="https://user-images.githubusercontent.com/736085/96254486-a6fc9380-0fb5-11eb-87ee-978dc6efc179.png">

But if you take a look at the amount of data read in bytes, again 1h average, it's a lot less balanced. We sometimes see that the busiest shard receives 50% traffic than the one that's least busy.

<img width="1282" alt="Screenshot 2020-10-16 at 13 44 38" src="https://user-images.githubusercontent.com/736085/96254549-be3b8100-0fb5-11eb-8c17-cad267ed45a9.png">

- Many storage systems out there put some limits on the maximum size of an object. Redis can only store 512 MB objects. Cassandra has a 2 GB limit on column values. Ceph's Rados has a maximum object size of 128 MB. (Buildbarn's LocalBlobAccess also has certain limits.) It is possible to decompose objects into chunks on the storage infrastructure side, but that removes the ability to perform data integrity checking at the chunk level, as multiple chunks need to be concatenated to perform checksum validation. When chunks are also spread across shards, an individual shard loses the ability to validate its own data entirely.
- Connection drops may occur while uploading files. When this happens, two things may happen: A) the client may attempt to re-upload the file from the beginning. B) the client somehow requests that the upload continues at the last received position. This would require the server to hold on to partial uploads, and negotiate with the client at which offset the upload needs to resume. That isn't fun to implement.
- The current approach doesn't make it possible to do validated reads of chunks of a file. If workers implement lazy loading of file contents (e.g., using a worker FUSE file system), it may be preferable to not read large files as a whole, but only read the parts that are actually used by build actions. This can't be done in a validated way if the only thing you have is the SHA-sum of the file as a whole.
- Relatedly, in #168 we're discussing how support for compressing CAS contents can be added to REv2. As we've noticed, the need for supporting partial uploads & downloads complicates that a lot.

[...]

PR #282, opened 2023-11-23
Adding blob split API

roloffs:
This is a proposal of a conservative extension to the `ContentAddressableStorage` service, which allows to reduce traffic when blobs are fetched from the remote CAS to the host for local usage or inspection. With this extension it is possible to request a remote-execution endpoint to split a specified blob into chunks of a certain average size. These chunks are then stored in the CAS as blobs and the ordered list of chunk digests is returned. The client can then check, which blob chunks are available locally from earlier fetches and fetch only the missing chunks. By using the digest list, the client can splice the requested blob from the locally available chunk data.

This extension could especially help to reduce traffic if large binary files are created at the remote side and needed locally such as executables with debug information, comprehensive libraries, or even whole file system images. It is a conservative extension, so no client is forced to use it. In our build-system project [justbuild](https://github.com/just-buildsystem/justbuild/), we have implemented this protocol extension for [server](https://github.com/just-buildsystem/justbuild/blob/3b1095f7e3584f37c984a79ad7a2b94ebaa0700f/src/buildtool/execution_api/execution_service/cas_server.cpp#L187-L269) and [client](https://github.com/just-buildsystem/justbuild/blob/3b1095f7e3584f37c984a79ad7a2b94ebaa0700f/src/buildtool/execution_api/remote/bazel/bazel_api.cpp#L105-L177) side.


EdSchouten:
Though this proposal may solve certain annoyances, it does not fix some other issues:

- The inability to upload files by uploading missing chunks. Given the fact that uploading is often slower than downloading (using typical internet connections provided by ISPs), I would argue that this is more important than downloading.
- The inability for clients to access files at random, without sacrificing the ability to check file integrity.

I would rather see that we try to solve these issues as part of REv3, by working towards eliminating the existence of large CAS objects in their entirety. See this section in the proposal I've been working on:

https://docs.google.com/document/d/1FxpdOzOhzOCTjjn2loppMlBzjqjU9WpYF4E1K6opxVI/edit#heading=h.3ix2ngn3uabo

sluongng:
This is quite similar to https://github.com/bazelbuild/remote-apis/pull/233/files. @roloffs have you had a chance to review the PR prior and related discussions?

I think both PRs approaching this by adding a separate RPC, which is good and V2 compatible.
The V3 support for this could be discussed separately and should not block V2 improvement as long as it’s backward compatible?

roloffs:
@EdSchouten, thanks for sharing the link to your REv3 discussion document and your comments about this proposal. Sorry for not being aware of this document. After going through it, I agree with you that this API extension would not make much sense in REv3 given your proposal in the "Elimination of large CAS objects" section. However, as also @sluongng stated, since this extension is conservative and backwards compatible to REv2 and the release of REv3 is very uncertain right now, it would not harm people that do not use it, but would provide advantages already now for people who use it and could also lead to insights for your content-defined chunking ideas for REv3, since we also used such an algorithm to split blobs.

I also agree with your concerns that uploading large files is something not covered by this proposal, while there exist relevant use-cases for this. However, I can think of a symmetric SpliceBlob rpc to allow for splitting a large blob on the client side, uploading only those parts of this blob that are missing on the server side, and then splicing there. This could be added in this PR as well.

@sluongng, thanks for pointing out this PR. Despite the fact they look very similar they actually target complementary goals. Let me explain why. While the PR from @EdSchouten introduces split and combine blobs rpcs, the goal is not to safe traffic but to introduce a blob splitting scheme, which allows to verify the integrity of a blob by validating the digests of its chunks without actually reading the whole chunk data. In order to achieve this, he introduced a new digest function SHA256TREE, which allows recursive digest calculation. I hope I did not completely misunderstood your intention @EdSchouten. In contrast, the presented splitting scheme targets reuse as much as possible with the final goal of traffic reduction between client and server. E.g., if a large binary in the remote CAS was just modified slightly and you want to use it locally, you would have to download it completely. Using the presented extension, only the binary differences between the two versions determined by content-defined chunking would have to be downloaded, which is typically much less than the whole data. As I said both splitting schemes are actually complementary and follow different goals.

roloffs:
While the actual choice of the splitting algorithm is mainly an implementation detail of the remote-execution endpoint (which of course affects the quality of the split result), the essential property of a server is to provide certain guarantees to a client if it successfully answers a `SplitBlob` request:

1. The blob chunks are stored in CAS.
2. Concatenating the blob chunks in the order of the digest list returned by the server results in the original blob.

Besides this guarantee, in order to increase the reuse factor as much as possible between different versions of a blob, it makes sense to implement a content-defined chunking algorithm. They typically result in chunks of variable size and are insensitive to the data-shifting problem of fixed-size chunking.

Such content-defined chunking algorithms typically rely on a rolling-hash function to efficiently compute hash values of consecutive bytes at every byte position in the data stream in order to determine the chunk boundaries. Popular algorithms for content-defined chunking are:

- [Polynomial rolling hash](https://ieeexplore.ieee.org/document/5390135)
- [Rabin fingerprint](http://www.cs.cmu.edu/~15-749/READINGS/optional/rabin1981.pdf)
- [Cyclic polynomial rolling hash](https://dl.acm.org/doi/abs/10.1145/256163.256168) (also called Buzhash)
- [Fast rolling Gear hash algorithm](https://www.usenix.org/conference/atc16/technical-sessions/presentation/xia) (also called FastCDC)

I have selected FastCDC as chunking algorithm for the endpoint implementation in our build system, since it has been proven to be very compute efficient and faster than the other rolling-hash algorithms while achieving similar deduplication ratios as the Rabin fingerprint. We already observed reuse factors of 96-98% for small changes, when working with big file-system images (around 800 MB) and also of 75% for a 300 MB executable with debug information.

Maybe, you want to have a look at our internal [design document](https://github.com/just-buildsystem/justbuild/blob/b7b438d5bba82a391d581537d144c2af6ac233ac/doc/concepts/blob-splitting.md) for more information about this blob-splitting API extension.

sluongng:
Ah I think I have realized what's missing here. Your design seems to be focusing on splitting the blob on the server side for the client to download large blobs. While I was thinking that blob splitting could happen to both the client side and the server side.

For example: a game designer may work on some graphic assets, say a really large picture. Subsequent versions of a picture may get chunked on the client side, compare the chunk list with what's available and only upload the parts that are missing.

So in the case where both client and server have to split big blobs for efficient download _AND_ upload, it's beneficial for 2 sides to agree upon how to split (and put back together) big blobs.

roloffs:
Yes, you are right, this design currently focuses on splitting on the server side and downloading large blobs, but as mentioned in a comment above, I am willing to extend this design proposal by a `SpliceBlob` rpc to handle chunked uploads. This allows splitting a blob on the client side, uploading chunks that are missing at the server, and then splicing the original blob there.

Maybe it is worth to mention that in this case, it is not necessarily required for client and server to agree upon the same splitting algorithm, since after the first round-trip overhead, the chunking algorithm for each direction anyway ensures an efficient reuse.

I will update this proposal to handle uploads for you to review. Thank you very much for your interest and nice suggestions.

sluongng:
Do keep in mind that there could be mixed usage of clients (a) with chunking support and clients (b) without chunking support.

So I do believe a negotiation via the initial GetCapability RPC, similar to the current Digest and Compressor negotiation, is much desirable. As the server would need to know how to put a split blob upload, from (a), back together to serve it to (b).

I would recommend throwing the design ideas into https://github.com/bazelbuild/remote-apis/issues/178. It's not yet settled whether chunking support needs to be a V3 exclusive feature, or we could do it as part of V2. Discussion to help nudge the issue forward would be much appreciated.

sluongng:
Do give https://github.com/bazelbuild/remote-apis/issues/272 and my draft PR a read on how client/server could negotiate for a spec leveraging GetCapabilities rpc. Could be useful if you want to have a consistent splitting scheme between client and server.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1778,6 +1851,53 @@ message GetTreeResponse {
...
+  // omitted.
+  string instance_name = 1;
+
+  // Expected digest of the spliced blob.
+  Digest blob_digest = 2;
+
+  // The ordered list of digests of the chunks which need to be concatenated to
+  // assemble the original blob.
+  repeated Digest chunk_digests = 3;
+}
```
sluongng, line 1893: I wonder if we should design this `SpliceBlobRequest` as part of `BatchUpdateBlobsRequest`.

The problem is that we are not pushing the chunk blobs in this request, only sending the CAS server metadata regarding combining some blobs into a larger blob. There could be a delay between the BatchUpdateBlobs RPC call and the SpliceBlob RPC call. That delay could be a few mili-seconds, or it could be weeks, or months after some of the uploaded chunks have expired from the CAS server. There is no transaction guarantee between the 2 RPCs.

So a way to have some form of transaction guarantee would be to send this as part of the same RPC that uploads all the blobs.

roloffs:
Hello @sluongng, I have updated the proposal by using the capabilities service as you have proposed. It is now possible for a client to determine the supported chunking algorithms at the server side and select one at a SplitBlob request.  By this means, the client can select one that it also uses locally so that both communication directions benefit from the available chunking data on each side. Furthermore, I have added some comments about lifetime of chunks. Thanks for your time reviewing this PR!

roloffs:
Hello all,

after spending quite some time working on this proposal and its implementation, I have finished incorporating all suggestions made by the reviewers and that came up during the working group meeting. Finally, the following high-level features would be added to the REv2 protocol:

1. Optional split operation at the CAS service to reduce traffic for the download direction.
2. Optional splice operation at the CAS service to reduce traffic for the upload direction.
3. Negotiation mechanism between client and server to agree about the used chunking algorithm.

This whole proposal is fully implemented in our own remote-execution implementation in justbuild:

- [Split](https://github.com/just-buildsystem/justbuild/blob/967cc82bfd3ae57fa942f2bcaca780db0d918f60/src/buildtool/execution_api/execution_service/cas_server.cpp#L226) operation provided at server side
- [Splice](https://github.com/just-buildsystem/justbuild/blob/967cc82bfd3ae57fa942f2bcaca780db0d918f60/src/buildtool/execution_api/execution_service/cas_server.cpp#L309) operation provided at server side

and used by the just client:
- [Split](https://github.com/just-buildsystem/justbuild/blob/967cc82bfd3ae57fa942f2bcaca780db0d918f60/src/buildtool/execution_api/remote/bazel/bazel_api.cpp#L116) operation used at client side
- [Splice](https://github.com/just-buildsystem/justbuild/blob/967cc82bfd3ae57fa942f2bcaca780db0d918f60/src/buildtool/execution_api/remote/bazel/bazel_api.cpp#L139) operation used at client side

From my side, this proposal is finished and ready for final review. What I would like to know from you is what needs now to be done that this proposal finally gets merged into main. I can also summarize it again at the next working group meeting and at best would like to know a decision how to proceed with this proposal. Thank you very much for your efforts.

@EdSchouten, @sluongng, @bergsieker

mostynb (commented):
Do we already effectively have the ability to splice blobs by using the bytestream API with read_offset and read_limit?

roloffs:
@mostynb, as far as I have understood the protocol, no. While the bytestream API with read_offset and read_limit allows you to partially read the content of a blob, it does not allow you to create a new blob from a batch of other blobs (its chunks) at the remote CAS.

The goal of blob splicing is that if a client regularly uploads slightly different large objects to the remote CAS, only the binary differences between the versions are needed to be uploaded and not the entire block of binary data every time. To achieve this, the client needs to split the large object into reusable chunks (which is typically done by content-defined chunking) and just uploads the chunks (handled as blobs) that are missing at the remote CAS, which are normally a lot when uploading the first time. If the client needs to upload this large object again, but a slightly different version of it (meaning only a percentage of the binary data has been changed), he again splits it into chunks and tests which chunks are missing at the remote CAS. Normally, content-defined chunking splits the binary data that hasn't been changed into the same set of chunks, only where binary differences occur, different chunks will be created. This means, only a fraction of the whole set of chunks need to be uploaded to the remote CAS in order to be able to reconstruct the second version of the large object at the remote CAS. The actual reconstruction of a large blob at the remote side is done using the splice command with a description of which chunks need to be concatenated (a list of chunk digests available at the remote CAS).

The split operation works exactly the other way around, when you regularly download an ever changing large object from remote CAS. Then, the server splits the large object into chunks, the client fetches only the locally missing chunks and reconstructs the large object locally from the locally available chunks.

Finally, to exploit chunking for both directions at the same time, it makes sense that the client and the server agree on a chunking algorithm to allow reusing chunks created on both sides. For this, we added a negotiation mechanism to agree on the chunking algorithm used on both sides.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1814,6 +1911,60 @@ message GetTreeResponse {
...
+  // support multiple instances of the execution system (with their own workers,
+  // storage, caches, etc.). The server MAY require use of this field to select
+  // between them in an implementation-defined fashion, otherwise it can be
+  // omitted.
+  string instance_name = 1;
+
+  // The digest of the blob to be splitted.
+  Digest blob_digest = 2;
+
+  // The chunking algorithm to be used. Must be IDENTITY (no chunking) or one of
```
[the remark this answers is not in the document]
roloffs: Not sure about that, I have basically copied the pattern from PR #276 to include a sane default value. I leave that open for your decision, I have no objections to change this.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1814,6 +1911,60 @@ message GetTreeResponse {
...
+message SplitBlobRequest {
+  // The instance of the execution system to operate against. A server may
+  // support multiple instances of the execution system (with their own workers,
+  // storage, caches, etc.). The server MAY require use of this field to select
+  // between them in an implementation-defined fashion, otherwise it can be
+  // omitted.
+  string instance_name = 1;
+
+  // The digest of the blob to be splitted.
+  Digest blob_digest = 2;
```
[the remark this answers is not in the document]
roloffs: Sure, makes sense, same for your remaining remarks. I will add a DigestFunction field to the respective messages. Thanks.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -2000,6 +2151,40 @@ message Compressor {
...
+    // following 64-bit polynomial of degree 53: 0x003DA3358B4DC173. The window
+    // size to be used is 64 bits.
+    RABINCDC = 1;
+
+    // Content-defined chunking using the FastCDC algorithm. The algorithm is
+    // described in this paper https://ieeexplore.ieee.org/document/9055082
+    // (Algorithm 2, FastCDC8KB). The algorithm is configured to have the
+    // following properties on resulting chunk sizes.
+    //  - Minimum chunk size: 2 KB
+    //  - Average chunk size: 8 KB
```
ulrfa, line 2179: I understand that using small chunk sizes, such as 8 KB, can increase the likelihood of deduplication and may also reduce the risk of disk storage fragmentation. However, have you considered if there is potential performance overhead of having too many fine-grained CAS blobs?

I can envision that a feature like this could also be beneficial in distributing the load more evenly across multiple CAS shards. But for such use cases, it might make sense to use much larger chunks, perhaps 8 MB? Should we somehow accommodate also for larger chunks in this PR?

luluz66: We experimented with FastCDC on approximately 5TB of real Bazel data from many different codebases and found that 0.5MB is a good trade-off between space savings and metadata overhead. Too small of a chunk size means the metadata for all chunks becomes very large / numerous, while too large of a chunk size means poor space savings.

ulrfa: Thanks @luluz66, I think 0.5 MB is more reasonable than 8 KB.

Do you think there is one value that would fit all, or should a size like this be allowed to be tuned in a flexible way?

build/bazel/remote/execution/v2/remote_execution.proto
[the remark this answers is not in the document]
roloffs: I have updated this field from IDENTITY to DEFAULT, because as you mentioned IDENTITY does not really makes sense being requested by a client. Instead to provide a proper default value for the chunking algorithm enum, I have introduced DEFAULT, which means the client does not care, which exact chunking algorithm is used by the server, just use the default implementation. If a client wants to negotiate more explicitly about the used chunking algorithm, it should specify one of the other enum values that are supported and advertised by the server.

I hope, this resolves your concerns? @mostynb

build/bazel/remote/execution/v2/remote_execution.proto
sluongng: 0.5 MB was the ideal range for us based on the Bazel-specific data set we were testing against. However, there would be no telling whether that number would be different for a different client/server pair, or a different data set.

So I think we would want a discovery mechanism for the FastCDC configuration on the server side. The client should follow the server's advertised setting in order to achieve the best result. WDYT?

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -2000,6 +2188,42 @@ message Compressor {
...
+  enum Value {
+    // No special constraints about the used chunking algorithm. If a client
+    // selects this value this means he does not care about which exact chunking
+    // algorithm is used. Thus, the server is free to use any chunking algorithm
+    // at its disposal. A server does not need to advertise this value.
+    DEFAULT = 0;
+
+    // Content-defined chunking using Rabin fingerprints. An implementation of
+    // this scheme in presented in this paper
+    // https://link.springer.com/chapter/10.1007/978-1-4613-9323-8_11. The final
```
EdSchouten, line 2202: This document is behind a paywall. Any chance we can link to a spec that is freely accessible?

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -2000,6 +2188,42 @@ message Compressor {
...
+    // implementation of this algorithm should be configured to have the
+    // following properties on resulting chunk sizes.
+    //  - Minimum chunk size: 128 KB
+    //  - Average chunk size: 512 KB (0x000000000007FFFF bit mask)
+    //  - Maximum chunk size: 2048 KB
+    // The irreducible polynomial to be used for the modulo divisions is the
+    // following 64-bit polynomial of degree 53: 0x003DA3358B4DC173. The window
+    // size to be used is 64 bits.
```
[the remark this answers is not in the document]
roloffs: You are right, that was a mistake from myside, what was meant is 64 bytes window size. I took this value from an existing implementation. Thanks for this finding!

The window size is a general parameter of a rolling hash function and the hash value or also fingerprint for a specific byte position is calculated for that window of bytes. Then, you move forward by one byte and calculate the hash value for this new window of bytes again. Thanks to the rolling-hash property, this process can be done very efficiently. So, the window size influences the hash value for a specific byte position and thus, the locations of actual chunk boundaries. Theoretically, we could use a window size of the minimum chunk size of 128 KB, but it is not common to use such a large window size in the implementations of content-defined chunking algorithms I have seen so far.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -2000,6 +2188,42 @@ message Compressor {
...
+    DEFAULT = 0;
+
+    // Content-defined chunking using Rabin fingerprints. An implementation of
+    // this scheme in presented in this paper
+    // https://link.springer.com/chapter/10.1007/978-1-4613-9323-8_11. The final
+    // implementation of this algorithm should be configured to have the
+    // following properties on resulting chunk sizes.
+    //  - Minimum chunk size: 128 KB
+    //  - Average chunk size: 512 KB (0x000000000007FFFF bit mask)
+    //  - Maximum chunk size: 2048 KB
```
EdSchouten, line 2207: Assuming my understanding of the algorithm is correct, doesn't that mean that there is a...
```
((2^19-1)/(2^19))^((2048-128)*1024) = 2.352...%
```
... probability that chunks end up reaching the maximum size? This sounds relatively high, but is arguably unavoidable.

A bit odd that these kinds of algorithms don't attempt to account for this, for example by repeatedly rerunning the algorithm with a smaller bit mask (18, 17, 16, [...] bits) until a match is found. That way you get a more even spread of data across such chunks, and need to upload fewer chunks in case data is injected into/removed from a sequence of chunks that would currently reach the 2 MB limit.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -2000,6 +2188,42 @@ message Compressor {
...
+    // size to be used is 64 bits.
+    RABINCDC = 1;
+
+    // Content-defined chunking using the FastCDC algorithm. The algorithm is
+    // described in this paper https://ieeexplore.ieee.org/document/9055082
+    // (Algorithm 2, FastCDC8KB). The algorithm should be configured to have the
+    // following properties on resulting chunk sizes.
+    //  - Minimum chunk size: 128 KB
+    //  - Average chunk size: 512 KB
+    //  - Maximum chunk size: 2048 KB
```
EdSchouten, line 2219: Looking at algorithm 2 in that paper, I see that I can indeed plug in these values to MinSize, MaxSize, and NormalSize. But what values should I use for MaskS and MaskL now? (There's also MaskA, but that seems unused by the algorithm.)

roloffs: Good question, as far as I have understood the algorithm, it is the normalized chunking technique that they use, which allows to keep these mask values as they are, but change the min/max/average chunk sizes according to your needs.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -2000,6 +2188,42 @@ message Compressor {
...
+    // at its disposal. A server does not need to advertise this value.
+    DEFAULT = 0;
+
+    // Content-defined chunking using Rabin fingerprints. An implementation of
+    // this scheme in presented in this paper
+    // https://link.springer.com/chapter/10.1007/978-1-4613-9323-8_11. The final
+    // implementation of this algorithm should be configured to have the
+    // following properties on resulting chunk sizes.
+    //  - Minimum chunk size: 128 KB
+    //  - Average chunk size: 512 KB (0x000000000007FFFF bit mask)
```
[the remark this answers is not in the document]
roloffs: All of the 19 lowest bits of the fingerprint need to be zero for the expression (fp & mask) to become 0, in which case you found a chunk boundary. Regarding your question about the average chunk size, you are right, the actual average chunk size = expected chunk size (512 KB) + minimum chunk size (128 KB). I will state this more clearly in the comments, thanks for pointing this out!

build/bazel/remote/execution/v2/remote_execution.proto
EdSchouten: Looking at algorithm 2 in that paper that attempts to compute 8 KB chunks, I think the probability of a random chunk having size `x` should be as follows ([link](https://www.desmos.com/calculator/2d2gqajtqb)):

desmos-graph

Now if I change that graph to use the minimum/average/maximum chunk sizes that you propose while leaving the bitmasks unaltered, I see this ([link](https://www.desmos.com/calculator/h92klnucmo)):

desmos-graph-2

If I change it so that `MaskS` has 21 bits and `MaskL` has 17, then the graph starts to resemble the original one ([link](https://www.desmos.com/calculator/zpuf6rhubw)):

desmos-graph-3

So I do think the masks need to be adjusted as well.

build/bazel/remote/execution/v2/remote_execution.proto
roloffs: Actually, the paper about FastCDC https://ieeexplore.ieee.org/document/9055082 also contains an algorithmic description of the Rabin fingerprinting technique (Algorithm 3. RabinCDC8KB). The only thing that is missing here is a precise description of the precomputed U and T arrays. Even though, they provide links to other papers, where these arrays are supposedly defined, I could not find any definition in these papers.

build/bazel/remote/execution/v2/remote_execution.proto
EdSchouten: Hmm... Making the graphs above was meaningful. I think it gives good insight in why FastCDC was designed the way it is. They are essentially trying to mimic a normal distribution ([link](https://www.desmos.com/calculator/v91cuknzrs)):

desmos-graph-4

Using a literal normal distribution would not be desirable, because it means that the probability that a chunk is created at a given point not only depends on the data within the window, but also the size of the current chunk. And this is exactly what CDC tries to prevent. So that's why they emulate it using three partial functions.

Yeah, we should make sure to set `MaskS` and `MaskL` accordingly.

roloffs: Wow, thanks @EdSchouten for this great analysis. I have to admit, I did not look into these mask values in that detail, because it also was not really explained in detail in the paper, but your concerns are absolutely right and we have to set the mask values accordingly, when the min/max/average chunk sizes are changed. I am just asking myself, since the paper authors mentioned they derived the mask values empirically, how should we adapt them?

@luluz66 , @sluongng since you mentioned you were working with the FastCDC algorithm with chunk sizes of 500 KB, I am wondering how you handled the adaption of the mask values or whether you did it at all. Can you share your experience here? Thank you so much.

sluongng: We have our fork of CDC implementation here https://github.com/buildbuddy-io/fastcdc-go/blob/47805a2ecd550cb875f1b797a47a1a648a1feed1/fastcdc.go#L162-L172

This work is very much in progress and not final. We hope that the current API will come with configurable knobs so that downstream implementation could choose what's best for their use cases and data.

EdSchouten: > I am just asking myself, since the paper authors mentioned they derived the mask values empirically, how should we adapt them?

I think that the authors of the paper just used the following heuristics:

1. Don't use the bottom 16 bits, because those likely contain low quality data.
2. Don't use the top 16 bits to get an equal comparison against many common Rabin fingerprinting implementations that use a 48 byte window.

(1) sounds like a good idea, but (2) can likely be ignored. So just pick 21 out of the top 48 bits, and then another 17 bits that area subset of the former.

build/bazel/remote/execution/v2/remote_execution.proto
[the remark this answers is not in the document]
EdSchouten: I just did some measurements, and it looks like using `fp & mask == 0` is actually a pretty bad choice. The reason being that it's pretty easy to craft windows for which `fp == 0`, namely a sequence consisting exclusively of 0-bytes. This has also been observed here:

https://ieeexplore.ieee.org/document/9006560

> Although using a cut value of zero seems to be a natural choice in theory, it turns out to be a poor choice in practice (Figure 2).

I noticed this especially when creating chunks of a Linux kernel source tarball (decompressed), for the same reason as stated in the article:

> It turns out that tar files use zeroes to pad out internal file structures. These padded structures cause an explosion of 4-byte (or smaller) length chunks if the cut value is also zero. In fact, over 98% of the chunks are 4-bytes long (Figure 2, (W16,P15,M16,C0) Table I).

I didn't observe the 4-byte chunks, for the reason that I used a minimum size as documented.

Results look a lot better if I use `fp & mask == mask`.

EdSchouten:
Hi @roloffs,

Earlier today I spent some time experimenting with FastCDC. In addition to implementing Algorithm 2 "FastCDC8KB" from the paper, I also wrote something corresponding to the following:

```
Input: data buffer, src; buffer length, n
Output: chunking breakpoint bestOffset

/* The same initialization as the original algorithm. */
MinSize = ...;
MaxSize = ...;
fp = 0;
if n <= MinSize then
    return n;
if n >= MaxSize then
    n = MaxSize;

/* Compute the rolling hash at the start of the cutoff range. */
i = MinSize - 64;
for ; i < MinSize; i++; do
    fp = (fp << 1) + Gear[src[i]];

/* Test if the cutoff range contains points for which the rolling hash is higher. */
bestFP = fp
bestOffset = MinSize
for ; i < MaxSize; i++; do
    fp = (fp << 1) + Gear[src[i]];
    if bestFP < fp then
        bestFP = fp;
        bestOffset = i + 1;
return bestOffset;
```

The following thought has gone into this variant:

- As we saw above, the expected chunk size of plain FastCDC8KB resembles a normal distribution. I fail to see why this should be preferred over a uniform distribution. It means that if cuts start to diverge due to differing file contents, it also takes potentially longer for the cuts to converge after the file contents become equal again. This due to the existence of a potential period/cadence in cuts.

- With FastCDC8KB, if no suitable location can be found to make a cut, the algorithm simply returns $MaxSize$. For Rabin fingerprinting the probability of hitting $MaxSize$ tends to be relatively high, but FastCDC's use of $MaskL$ acts somewhat as a safeguard. That said, if we do reach the maximum it stops being "content defined" chunking altogether. The poor selection of the cut could potentially trickle through into subsequent chunks/cuts. By always placing the cut at the point with the highest $fp$ instead of hoping to see a point where the desired number of bits is clear, the chunking remains "content defined".

If implemented naively, the algorithm above will have a worse worst-case running time. Namely, we always compute $fp$ over the full $MaxSize$ bytes of input, even though the resulting chunk may be close to $MaxSize$. This means that the worst-case running time of the algorithm above is $O(n * \frac{MaxSize}{MinSize})$ instead of just $O(n)$. Fortunately, this can easily be addressed by making an array of $bestFP$ and $bestOffset$ that is $\lceil \frac{MaxSize}{MinSize} \rceil$ elements in size. This array can preserve rolling hashes, to be carried over to subsequent rounds. Using this approach I managed to end up with an implementation with an amortized worst-case running time of $O(n)$. Its throughput is nearly indistinguishable from plain FastCDC8KB, due to the hot path still being the rolling hash computation, which is unaltered.

With regards to performance of the chunking performed, I downloaded some different versions of the Linux kernel source code.  Because the upstream tarballs contain timestamps, I unpacked them and concatenated all of the files contained within, which gave me some ~1.4 GB files, consisting mostly of text. I cut these files into ~10 KB chunks using both FastCDC8KB and the algorithm described above, giving me ~140k chunks per kernel. Comparing Linux 6.7.10 with Linux 6.8.1, I see that:

- Using FastCDC8KB, 12.29% of the chunks change.
- Using the algorithm above, only 11.44% of the chunks change.

This means that the algorithm described above performs $1 - 11.44\\% / 12.29\\% = 6.92\\%$ better at eliminating redundancy for this specific dataset.

Would you by any chance be interested in trying to reproduce these results? I'd be interested in knowing whether these savings hold in general.

sluongng:
@EdSchouten what do you think would be the next step here for this PR?

To me, it seems like we have established the value of using FastCDC as one of the potential chunking algorithms.
FastCDC comes with several configuration knobs that could be finetuned against the data to improve the efficiency of runtime and deduplication hits.

I think this means that we gonna need a mechanism for the server to advertise the desired FastCDC config, and the client to comply accordingly.

As for the modified-FastCDC, if we cannot expose it as an discoverable configuration knobs, then we could add a new chunking algorithm after FastCDC is merged.

WDYT?

EdSchouten:
I have no opinion whatsoever what should happen here. As I mentioned during the last working group meeting, I have absolutely no intent to implement any of this on the Buildbarn side as part of REv2. I don't think that there is an elegant way we can get this into the protocol without making unreasonable sacrifices. So with regards to what the next steps are, that's for others within the working group to decide.

EdSchouten:
FYI: If other people want to do some testing in this area, I have just released the source code for the algorithm described above: https://github.com/buildbarn/go-cdc

sluongng:
Hey @roloffs ,

We had a monthly REAPI meeting this week and the maintainers have concluded that we should push this PR forward.
As folks discussed on the current state of this PR, notable blockers that need some additional work:

- Clarification on which chunking algorithm should be included and why they should be included. In the specific case of CDC, the maintainers have asked whether we could provide a "sane default" for the set of tuning involved in setting it up.

- A test vector should be provided for each chunking algorithm. This should help folks verify their implementations against the spec more easily.

- Some document improvement is needed. Specifically, the maintainers suggested that this new API set should be marked as an "optional extension" of REAPI v2. The purpose is to highlight that this is a temporary solution for the large blobs problem for REAPI v2. I also suspect using the word "Experimental" here would help lower the expectations from the end users and help us merge this PR earlier. The maintainers hope that REAPI v3 will provide a more concrete solution to this problem.

With that said, please let me know if you still have the capacity to work on this PR.
If not, I could give it a try in a few weeks to see if I could help drive it to the finish line.

cc: @buchgr @EdSchouten

roloffs:
Hello @sluongng, sorry for not being responsive for a longer time, I was on parental leave from work for three months and will catch up everything during the next days. I am willing to finish this PR and also have the capacity to do this from now on. Still, if you are willing to support, it would be appreciated since I have to consider and incorporate all great comments from @EdSchouten. Today, there is a Remote Execution API Working Group Meeting, however, I won't attend since there is not much to report. I will do my best to finish everything until the next meeting in August.

roloffs:
After our dedicated meeting to discuss the different large-blobs proposals, we agreed on that we would like to have both proposals in the REv2 protocol, because they have different pros and cons. According to the agreement, I have removed any negotiation parts from the split/splice proposal and have it ready for a final review. A proper negotiation proposal for performance improvement will be a follow-up PR.

It would be nice, if the key players would give their final remarks or approval to merge this PR. Thank you very much.
@sluongng , @EdSchouten , @bergsieker , @mostynb

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -430,6 +430,105 @@ service ContentAddressableStorage {
...
+
+  // Splice a blob from chunks.
+  //
+  // This is the complementary operation to the
+  // [ContentAddressableStorage.SplitBlob][build.bazel.remote.execution.v2.ContentAddressableStorage.SplitBlob]
+  // function to handle the chunked upload of large blobs to save upload
+  // traffic.
+  //
+  // If a client needs to upload a large blob and is able to split a blob into
+  // chunks locally according to some content-defined chunking algorithm, it can
```
juergbi, line 495: Minor comment as this explicitly mentions content-defined chunking. There may be clients that construct a large blob from semantically meaningful, smaller blobs, in which case it may actually be better to use that semantic chunking also for the upload optimization (or possibly a combination of semantic chunking and CDC). As far as I can tell, the splice API should just work with such semantic chunking.

Is there a reason to discourage semantic chunking even in cases where it would be easily possible? It may be too niche to mention this in the doc comment but I'd be interested if anyone else has any thoughts on this aspect.

aehlig: The wording is indeed a bit misleading. What is meant is that the blob should be split in such a way thyat it produce reusable chunks. Doing content-defined splitting is a generic way to do this, as it does not require any understanding of the semantics of the blob. Therefore, it is a very popular method, but if the client is aware of the semantics of the blob and hence can do a better splitting based on this knowledge, this is fine as well.

@roloffs We probably should reword the comment, e.g., saying `... and is able to split a blob into chunks in such a way that reusable chunks are obtained, e.g., by means of content-defined chunking, ...`.

roloffs: I have changed the wording.

aehlig:
@sluongng Does this pull request now reflect what we agreed upon during the dedicated large-blob meeting (so that we can merge it), or do you think further changes are necessary?

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -430,6 +430,106 @@ service ContentAddressableStorage {
...
+  // However, clients need to check first the server capabilities, whether blob
+  // splitting is supported by the server.
+  //
+  // Hints:
+  //
+  // * For clients, it is recommended to verify whether the digest of the blob
+  //   assembled by the fetched chunks results in the requested blob digest.
+  //
+  // * Since the generated chunks are stored as blobs, they underlie the same
+  //   lifetimes as other blobs. However, their lifetimes are extended if they
```
bergsieker, line 470: I think it's clearest to express that the chunk lifetimes are independent from the lifetime of the original blob. In particular:

* A blob and any chunks derived from it may be evicted from the CAS at different times.
* A call to Split extends the lifetime of the original blob, and sets the lifetime of the resulting chunks (or extends the lifetime of an already-existing chunk).
* Touching a chunk extends its lifetime, but does not extend the lifetime of the original blob.
* Touching the original blob does not extend the lifetime of chunks derived from it (except in the case of a subsequent Split call).

roloffs: Thanks for that hint, I have adapted the wording and incorporated your suggestions.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -439,6 +439,115 @@ service ContentAddressableStorage {
...
+  //
+  // Clients are free to use this functionality, it is just an optimization to
+  // reduce download network traffic, when downloading large blobs from the CAS.
+  // However, clients need to check first the server capabilities, whether blob
+  // splitting is supported by the server.
+  //
+  // Hints:
+  //
+  // * For clients, it is recommended to verify whether the digest of the blob
+  //   assembled by the fetched chunks results in the requested blob digest.
```
tjgq, line 476: Clients SHOULD verify ...

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -439,6 +439,115 @@ service ContentAddressableStorage {
...
+  // reduce upload traffic, when uploading large blobs to the CAS. However,
+  // clients need to check first the server capabilities, whether blob splicing
+  // is supported by the server.
+  //
+  // Hints:
+  //
+  // * In order to ensure data consistency of the CAS, the server will verify
+  //   the spliced result whether digest calculation results in the provided
+  //   digest from the request and will reject a splice request if this check
+  //   fails.
```
tjgq, line 534: Is this intended to be required of a conforming server implementation, or optional? Please use one of MUST or SHOULD to clarify.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -439,6 +439,115 @@ service ContentAddressableStorage {
...
+  //
+  // Servers are free to implement this functionality, but they need to declare
+  // whether they support it or not by setting the
+  // [CacheCapabilities.blob_splice_support][build.bazel.remote.execution.v2.CacheCapabilities.blob_splice_support]
+  // field accordingly.
+  //
+  // Clients are free to use this functionality, it is just an optimization to
+  // reduce upload traffic, when uploading large blobs to the CAS. However,
+  // clients need to check first the server capabilities, whether blob splicing
+  // is supported by the server.
```
tjgq, line 527: Clients MAY use ... MUST first check the server capabilities

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -439,6 +439,115 @@ service ContentAddressableStorage {
...
+  // If a split request is answered successfully, a client can expect the
+  // following guarantees from the server:
+  //  1. The blob chunks are stored in CAS.
+  //  2. Concatenating the blob chunks in the order of the digest list returned
+  //     by the server results in the original blob.
+  //
+  // Servers are free to implement this functionality, but they need to declare
+  // whether they support it or not by setting the
+  // [CacheCapabilities.blob_split_support][build.bazel.remote.execution.v2.CacheCapabilities.blob_split_support]
+  // field accordingly.
```
tjgq, line 466: Servers MAY implement this functionality, but they MUST declare ...

(The MUST is backwards compatible because in proto3 unset is the same as false for boolean fields)

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -439,6 +439,115 @@ service ContentAddressableStorage {
...
+  //
+  // Clients are free to use this functionality, it is just an optimization to
+  // reduce download network traffic, when downloading large blobs from the CAS.
+  // However, clients need to check first the server capabilities, whether blob
+  // splitting is supported by the server.
+  //
+  // Hints:
```
tjgq, line 473: I'd remove `Hints:` both here and below, because AFAICT the following text is still spelling out requirements for implementers, not just making suggestions.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -439,6 +439,115 @@ service ContentAddressableStorage {
...
+  //
+  // * In order to ensure data consistency of the CAS, the server will verify
+  //   the spliced result whether digest calculation results in the provided
+  //   digest from the request and will reject a splice request if this check
+  //   fails.
+  //
+  // * When blob splitting and splicing is used at the same time, the clients
+  //   and the server should out-of-band agree upon a chunking algorithm used by
+  //   all parties to benefit from each others chunk data and avoid unnecessary
+  //   data duplication.
```
tjgq, line 539: s/should/SHOULD/
s/all/both/

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -439,6 +439,115 @@ service ContentAddressableStorage {
...
+  //      already-existing chunks).
+  //    * Touching a chunk extends its lifetime, but does not extend the
+  //      lifetime of the original blob.
+  //    * Touching the original blob extends its lifetime, but does not extend
+  //      the lifetimes of chunks derived from it.
+  //
+  // * When blob splitting and splicing is used at the same time, the clients
+  //   and the server should out-of-band agree upon a chunking algorithm used by
+  //   all parties to benefit from each others chunk data and avoid unnecessary
+  //   data duplication.
```
tjgq, line 494: s/should/SHOULD/
s/all/both/

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1846,6 +1955,91 @@ message GetTreeResponse {
...
+  // the order of the digests given by this list.
+  repeated Digest chunk_digests = 1;
+
+  // The digest function of the chunks.
+  //
+  // If the digest function used is one of MD5, MURMUR3, SHA1, SHA256,
+  // SHA384, SHA512, or VSO, the client MAY leave this field unset. In
+  // that case the server SHOULD infer the digest function using the
+  // length of the blob digest hashes and the digest functions announced
+  // in the server's capabilities.
```
tjgq, line 1995: s/server/client/, surely?

On the other hand - would it make more sense to specify that the digest function for `chunk_digests` must be the same as the request? I believe most clients stick to a single digest function, and it seems unreasonable to (in theory) let the server force a different one on the client.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1846,6 +1955,91 @@ message GetTreeResponse {
...
+  repeated Digest chunk_digests = 3;
+
+  // The digest function of the blob to be spliced as well as of the chunks to
+  // be concatenated.
+  //
+  // If the digest function used is one of MD5, MURMUR3, SHA1, SHA256,
+  // SHA384, SHA512, or VSO, the client MAY leave this field unset. In
+  // that case the server SHOULD infer the digest function using the
+  // length of the blob digest hashes and the digest functions announced
+  // in the server's capabilities.
```
tjgq, line 2023: Should it be allowed for `chunk_digests` to contain digests of different lengths (and thus different digest functions)?

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1846,6 +1955,91 @@ message GetTreeResponse {
...
+  Digest blob_digest = 1;
+
+  // The digest function of the spliced blob.
+  //
+  // If the digest function used is one of MD5, MURMUR3, SHA1, SHA256,
+  // SHA384, SHA512, or VSO, the client MAY leave this field unset. In
+  // that case the server SHOULD infer the digest function using the
+  // length of the blob digest hashes and the digest functions announced
+  // in the server's capabilities.
+  DigestFunction.Value digest_function = 2;
```
tjgq, line 2040: Same as for the splitting case - would it be better to specify that the response must use the same digest function as the request, so that the server is not allowed to force a different one on the client? (This also requires disallowing multiple functions in the request, see comment above)

tjgq (changes requested):
LGTM overall, just a few comments.

I also think it would be worthwhile to specify a negotiation procedure for the chunking algorithm before usage becomes too widespread and it gets harder to do; but it's fine to start without it.

roloffs (commented):
Thanks for the review, I have updated the pull request according to your suggestions. Awaiting your final decision to move on with this PR.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1846,6 +1955,78 @@ message GetTreeResponse {
...
+message SpliceBlobRequest {
+  // The instance of the execution system to operate against. A server may
+  // support multiple instances of the execution system (with their own workers,
+  // storage, caches, etc.). The server MAY require use of this field to select
+  // between them in an implementation-defined fashion, otherwise it can be
+  // omitted.
+  string instance_name = 1;
+
+  // Expected digest of the spliced blob.
+  Digest blob_digest = 2;
```
fmeum, line 2004: Could you clarify whether this must be set? Since the response contains the digest, we should probably also include the rationale for adding this field to the request.

roloffs: Hi @fmeum, thanks for that hint. I agree with the comments from @sluongng and @aehlig and have improved the description of the rationale behind this field accordingly. Please check, if it now meets your expectations.

fmeum (approved):

PR #283, opened 2023-12-22
Introduce attempt_number to RequestMetadata

hoj-stripe:
I'd like to propose adding a field to RequestMetadata that indicates if this request was a retry from the client. This allows the remote server to choose their behavior differently when this invocation had to be retried from the client, and also helps to track the stats with retried invocations.

More motivation in: issue https://github.com/bazelbuild/bazel/issues/20664 and PR https://github.com/bazelbuild/bazel/pull/20665

### DIFF
--- a/build/bazel/remote/execution/v2/remote_execution.proto
+++ b/build/bazel/remote/execution/v2/remote_execution.proto
@@ -2119,7 +2119,7 @@ message RequestMetadata {
   // or equality across invocations, though some client tools may offer these guarantees.
   string configuration_id = 7;
 
-  // If Bazel automatically retries an invocation, this value is set to the
+  // If the client automatically retries an invocation, this value is set to the
   // number of invocation attempts that have been started so far. For example, a
   // value of 2 would indicate that it is the second attempt.
   int32 attempt_number = 8;
--- a/build/bazel/remote/execution/v2/remote_execution.proto
+++ b/build/bazel/remote/execution/v2/remote_execution.proto
@@ -2118,4 +2118,9 @@ message RequestMetadata {
   // There is no expectation that this value will have any particular structure,
   // or equality across invocations, though some client tools may offer these guarantees.
   string configuration_id = 7;
+
+  // If Bazel automatically retries an invocation, this value is set to the
+  // number of invocation attempts that have been started so far. For example, a
+  // value of 2 would indicate that it is the second attempt.
+  int32 attempt_number = 8;
 }


build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -2118,4 +2118,9 @@ message RequestMetadata {
   // There is no expectation that this value will have any particular structure,
   // or equality across invocations, though some client tools may offer these guarantees.
   string configuration_id = 7;
+
+  // If Bazel automatically retries an invocation, this value is set to the
```
brentleyjones, line 2122: The API should be agnostic to the client, so it shouldn't call the client "Bazel".

EdSchouten:
Instead of an attempt number, why not include the name of the previous operation?

build/bazel/remote/execution/v2/remote_execution.proto
hoj-stripe: Oops, thanks for catching!

hoj-stripe:
@EdSchouten 

> Instead of an attempt number, why not include the name of the previous operation?

That sounds like another good way of indicating retries, and I actually like that approach more! The only concern is: I'm not sure if tracking that would be relevant with the future of Bazel, given potential discussion: `With attempt_number, whether Bazel should use the same invocation id for the retries?` from https://github.com/bazelbuild/bazel/issues/20605. 

> Would RequestMetadata be the right spot for this, or should we put it in ExecuteRequest?

The main purpose (for us) of this field was to tell remote cache server that the cache may have been evicted. I think it makes this field simpler and cleaner to use here and there, and not only from `Execute` calls, by putting this into `RequestMetadata`.

EdSchouten:
Why would you need to tell a remote cache server that a cached entry may have been evicted?

hoj-stripe:
We are actually suffering from way too large input trees. `GetTree` is quite slow for us to be able to inspect every single blob in the input tree whenever `GetActionResult` is called, and thus we can't guarantee that the TTL is properly updated for all the depending blobs for the action cache on `GetActionResult` :( So the idea is to only run that when the invocation is retried with potentially evicted cache, which is way simpler than somehow optimizing tree traversal for large input trees and also optimizing blob existence check so it's fast enough to run for every `GetActionResult`.

EdSchouten:
GetActionResult should not bump the TTL on input roots. Only on individual output files. This only becomes annoying if an action yields a large number of output files or very large Tree objects. But those cases tend to be rare during typical builds. The vast majority of build actions only yield <5 output files.

Servers can also easily cache the digests of objects whose TTLs were recently bumped, so that GetActionResult can skip repeated renewal.

hoj-stripe:
Yep I agree that it's not an expectation to update the input tree on `GetActionResult`, that was just an alternative option to work with massive input root with potentially evicted blobs. I think I slightly misunderstood your previous message, and we're less worried of the output blobs referenced by the `ActionResult` being evicted during a build, but more of that output being used as an input but then it got GC'ed before an invocation.

The main issue we're facing is shown in [this repro](https://github.com/hoj-stripe/cas-delete-error-2/blob/master/repro.sh#L85-L95). Since we can't easily track all the actions that depend on a specific blob, when we GC a blob we likely will have some actions that will always fail until we "fix" the action cache. However, if this field is present then we can automatically trigger the "fix" and make it not fail.

EdSchouten:
> Since we can't easily track all the actions that depend on a specific blob, when we GC a blob we likely will have some actions that will always fail until we "fix" the action cache.

Reading this, I think the root cause of your problems is that the garbage collector of your CAS is racing against GetActionResult() and FindMissingBlobs() bumping the TTL on your objects. So essentially this sequence of operations:

1. Garbage collector finds that object X stored in the CAS is potentially unused.
2. Some client calls GetActionResult() against an Action having object X as an output, or FindMissingBlobs() against object X, bumping its TTL.
3. Garbage collector issues a delete operation against object X.

If this is indeed the case, then your CAS ought to be fixed. One very easy way to work around this is by storing your data in two places. For example, two S3 buckets. Let's call them "Storage A" and "Storage B".

- When you need to read objects, you do a coin toss.
  - When you get heads, you first try storage A. If that returns a NOT_FOUND, you fall back to reading from storage B. If that one does return success, you may optionally repair storage A while serving the data.
  - When you get tails, you do the opposite. First storage B and then storage A.
- When you need to write objects, you multiplex and write the data to both storage backends.
- When GetActionResult() or FindMissingBlobs() is called, you check for the existence of objects in both storage backends. If the results are inconsistent (i.e., the object only exists in storage A or B), you repair by exchanging prior to returning results to the client.

Now garbage collection is really easy. Week 1 you run a GC job against storage A. This can either be implemented gracefully (i.e. taking into consideration when an object was last used), or destructive (i.e. deleting everything that's there). Week 2 you run the same GC job against storage B. And then you repeat. With such a system in place, you guarantee that GetActionResult() and FindMissingBlobs() extend the lifetime of objects by at least one week. You can change the interval if you want a different retention period.

Buildbarn has been using this approach for almost exactly four years now (see [this ADR](https://github.com/buildbarn/bb-adrs/blob/master/0002-storage.md#adding-fault-tolerance) and [this implementation](https://github.com/buildbarn/bb-storage/blob/master/pkg/blobstore/mirrored/mirrored_blob_access.go)) and it's been working great. Considering how easy it was to design this, I tend to have very little symphathy for proposals that add more quirks to the protocol, simply to support sub par CAS implementations.

hoj-stripe:
Thanks for the interest and the detailed write-up on a potential strategy! Buildbarn's approach was very interesting when I first read but it wasn't preferred due to our limitations. We heavily rely on the remote cache, and the size of our CAS is extremely large. Duplicating the storage would mean that we would pay twice the cost of already-expensive cost (millions of dollars).

The solutions for the problem set we're facing can be generalized into choosing the balance between:
1. Designing a fault-free system that updates TTL very sophistically, in a way that it eliminates the race
2. Designing a fault-fixing system that lets the cache go bad and fix afterwards, when we detect that it's bad

We've been optimizing the storage's TTL strategy design as much as possible to achieve the solution 1. We anticipate that even with simple TTL with minimal updates, we can eliminate most of the faulty cases and only leave some race conditions, which is currently still quite bad to the developers' experience since we can't easily recover.

However, we determined that resolving the race conditions costs us either another layer of cache (second cache as proposed above, holding soon-to-be-removed blobs table, or caches with generations, etc.) and/or another step of operation (holding a lock, or checking back after scheduled GC, etc.).

In contrast, with the added field of `attempt_number`, solution 2 will be trivial. For those few cases of race conditions, we can just simply retry the invocation while inspecting all the `ActionResult` to invalidate the ones with evicted blobs. It will let us maintain our cache to be simpler with less layers and less steps of operations with the cost of slower invocation for those few cases. I really wish that we could buy the sympathy with our intention on maintaining simpler and less expensive cache.


I also think this field can be interesting to broader audience, at least to the Bazel users who want to debug their remote instances' behavior followed by or that led to cache eviction. I guess for this purpose though, including the name of the previous operation would be more effective?

bergsieker:
I'm not a fan of leveraging the RequestMetadata in this way. The metadata is intended to be an optional piece of information that the server can use for logging/debugging, not something that affects the way an RPC is processed.

I'd be more open to a field within the GetActionResultRequest that provided similar functionality, but I'm not sure that it actually buys you anything--there's still a race between the GC and the TTL extension from GetActionResult--and I also feel like it's yoking the RE API with the specific semantics of how Bazel handles retrying an invocation.

Probably the correct thing in this case is for Bazel to retry the invocation (or the Action?) with skip_cache_lookup=true. This will cause the action to be re-executed automatically (and the existing cache entry should be rewritten with the new result).

EdSchouten:
@hoj-stripe can you also let us know whether this PR is still relevant? I remember that we had a pretty long discussion on Slack in response to this PR, where we discussed options for you folks to make your storage comply to the spec without requiring an intrusive change like this.

hoj-stripe:
Ah yep sorry about that, I forgot to update in this PR :(. Based on the discussion, I agree on closing this PR. Thanks for your opinions!

PR #285, opened 2024-01-15
Clarify how Digest.hash is encoded generally.

tjgq:
It seems bad form to special-case SHA-256 while not saying anything about other digest functions. (I welcome corrections if my proposed wording does not correspond to actual usage in the wild.)

### DIFF
--- a/build/bazel/remote/execution/v2/remote_execution.proto
+++ b/build/bazel/remote/execution/v2/remote_execution.proto
@@ -981,7 +981,8 @@ message SymlinkNode {
 // serializing, but care should be taken to avoid shortcuts. For instance,
 // concatenating two messages to merge them may produce duplicate fields.
 message Digest {
-  // The hash, represented as a lowercase hex string including any leading zeroes.
+  // The hash, represented as a lowercase hexadecimal string, padded with
+  // leading zeroes up to the hash function length.
   string hash = 1;
 
   // The size of the blob, in bytes.
--- a/build/bazel/remote/execution/v2/remote_execution.proto
+++ b/build/bazel/remote/execution/v2/remote_execution.proto
@@ -981,8 +981,7 @@ message SymlinkNode {
 // serializing, but care should be taken to avoid shortcuts. For instance,
 // concatenating two messages to merge them may produce duplicate fields.
 message Digest {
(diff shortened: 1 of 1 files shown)


build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -981,8 +981,7 @@ message SymlinkNode {
 // serializing, but care should be taken to avoid shortcuts. For instance,
 // concatenating two messages to merge them may produce duplicate fields.
 message Digest {
-  // The hash. In the case of SHA-256, it will always be a lowercase hex string
-  // exactly 64 characters long.
+  // The hash, represented as a lowercase hex string including any leading zeroes.
```
EdSchouten, line 984: LGTM, though I do think the "including any leading zeroes" part might be ambiguous. Does this mean that I can take an arbitrary SHA-256 hash and add as many leading zeroes to it as I want? Likely not. Maybe phrase it the other way around. "Leading zeroes MUST NOT be removed."

bergsieker (approved):

build/bazel/remote/execution/v2/remote_execution.proto
bergsieker: I'm not convinced that it's clearer one way vs the other.

mostynb: What if we were to say "In the case of SHA-256, it will always be a string containing exactly 64 hex characters (digits 0 through 9 or the letters a through f)." ?

tjgq: My point is that saying anything specifically about SHA-256 doesn't technically clarify how to proceed for the other hash functions; it seems bad form to describe something that applies to all hash functions in terms of one of them.

How about: "The hash, represented as a lowercase hexadecimal string, padded with leading zeroes up to the hash function length"?

EdSchouten: Sgtm!

bergsieker: That wording SGTM.

mostynb: sgtm3

bergsieker (approved):

PR #286, opened 2024-01-23
Support digest functions in remote_asset API

sluongng:
This will allow client(like Bazel) to specify the `digest_function` to the remote server (so that `--experimental_remote_downloader` could use non-SHA256 digest functions).

This is a recreation of https://github.com/bazelbuild/remote-apis/pull/273/ with reviewers' comment addressed

### DIFF
--- a/build/bazel/remote/asset/v1/remote_asset.proto
+++ b/build/bazel/remote/asset/v1/remote_asset.proto
@@ -240,14 +240,16 @@ message FetchBlobResponse {
   // The digest of the file's contents, available for download through the CAS.
   build.bazel.remote.execution.v2.Digest blob_digest = 5;
 
-  // This field SHOULD be set to the digest function that was used by the server
-  // to compute [FetchBlobResponse.blob_digest].
-  // Clients could use this to determine whether the server honors
-  // [FetchBlobRequest.digest_function] that was set in the request.
+  // Indicates if the server used the client-requested digest function for
+  // computing [FetchBlobResponse.blob_digest].
   //
-  // If unset, clients SHOULD default to use SHA256 regardless of the requested
-  // [FetchBlobRequest.digest_function].
-  build.bazel.remote.execution.v2.DigestFunction.Value digest_function = 6;
+  // If `true`, the server respected the [FetchBlobRequest.digest_function]
+  // choice made by the client. Clients can rely on this to verify that the
+  // server computed `blob_digest` using the requested digest function.
+  //
+  // If this field is unset, clients SHOULD assume SHA256 was used for digest
+  // computation, regardless of the function requested in [FetchBlobRequest.digest_function].
+  bool respect_digest_function = 6;
 }
 
 // A request message for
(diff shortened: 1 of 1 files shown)


sluongng:
cc: @mostynb @EdSchouten

EdSchouten (approved):

build/bazel/remote/asset/v1/remote_asset.proto
```diff
@@ -201,6 +201,13 @@ message FetchBlobRequest {
   //
   // Specified qualifier names *MUST* be unique.
   repeated Qualifier qualifiers = 5;
+
+  // The digest function the server must use to compute the digest.
+  //
+  // If unset, the server SHOULD select one of MD5, MURMUR3, SHA1, SHA256,
```
bergsieker, line 207: I like Mostyn's suggestion in https://github.com/bazelbuild/remote-apis/pull/273/files#r1344582116 to add a field to the response indicating what digest was actually used, as a way of differentiating servers that honour vs ignore this field.

sluongng: Added

build/bazel/remote/asset/v1/remote_asset.proto
```diff
@@ -234,6 +239,15 @@ message FetchBlobResponse {
   // The result of the fetch, if the status had code `OK`.
   // The digest of the file's contents, available for download through the CAS.
   build.bazel.remote.execution.v2.Digest blob_digest = 5;
+
+  // This field SHOULD be set to the digest function that was used by the server
+  // to compute [FetchBlobResponse.blob_digest].
+  // Clients could use this to determine whether the server honors
+  // [FetchBlobRequest.digest_function] that was set in the request.
```
EdSchouten, line 246: Couldn't it be a simple boolean if it's only used by the client to validate that it was respected by the server?

bergsieker: I suppose it could, but it seems weird to have a field that explicitly says, "I honoured this part of your request," vs. one that says "Here's how to interpret this other part of the response (which has the incidental effect of confirming that the system honoured the digest_function of the request." Honestly it seems like an oversight that the current API doesn't qualify blob_digest with a digest_function, given the way this has evolved in the RE API.

sluongng: Yes, it's a bit unfortunate.

I will go ahead with the boolean suggestion. I too think it's ugly, but alternative routes would require more disruptive changes.

bergsieker: I still think that including digest_function is the correct path here. There are already many places in the API where we pair a Digest with a digest_function so that the receiver knows how to interpret the field, and a digest_function field is useful even if the client *doesn't* specify a digest_function in the request (e.g., the server can encode SHA256 in the response to remove any ambiguity).

Ed, WDYT?

EdSchouten: I'm fine either way. 😀

sluongng: Reverted from `bool -> DigestFunction.Value`.

This PR is now ready to be merged.

@bergsieker you could use the Github UI to squash the commits into 1 single commit. Or I could manually squash them and push them as well.

bergsieker (approved):

PR #287, opened 2024-02-13
inline hinting for ExecuteRequest

werkt:
Match the capabilities of GetActionRequest to support hints for composing ActionResults for Execute

### DIFF
--- a/build/bazel/remote/execution/v2/remote_execution.proto
+++ b/build/bazel/remote/execution/v2/remote_execution.proto
@@ -1433,6 +1433,20 @@ message ExecuteRequest {
   // length of the action digest hash and the digest functions announced
   // in the server's capabilities.
   DigestFunction.Value digest_function = 9;
+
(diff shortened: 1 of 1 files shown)


bergsieker (approved):

EdSchouten (commented):
Instead of doing this, what are our thoughts on having a nested message for specifying all these hints, so that it's usable from within both GetActionResultRequest and ExecuteRequest?

bergsieker:
> Instead of doing this, what are our thoughts on having a nested message for specifying all these hints, so that it's usable from within both GetActionResultRequest and ExecuteRequest?

Conceptually that's fine, but given that the messages already exist in one place, I think it's better to just duplicate them here than to force another migration from moving fields around. The practical implication of switching to a nested message would be that clients end up needing to set both fields in GetActionRequest indefinitely.

werkt:
@bergsieker good for a merge? no privs here

PR #289, opened 2024-03-07
Add readme notes for worker apis

cal-pratt:
Adding a note into the readme, aimed at new contributors and groups looking to design their own REAPI system.

The main point I wanted to add, is that there is no standardization on how schedulers/servers and workers must communicate. I provided some reference documentation which I found while doing my own research on scheduling implementations.

If there are more links from other implementations we should add them as well!

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,8 +75,8 @@ Typically, workers are domain specific, optimized for custom scheduling
 protocols defined by the server implementation.
 
 * [Remote Worker API](https://goo.gl/oojM5H)
-* [Buildfarn](https://bazelbuild.github.io/bazel-buildfarm/docs/architecture/queues/)
-* [Buildbarn](https://nluug.nl/bestanden/presentaties/2023-05-11-ed-schouten-buildbarn-a-distributed-build-cluster.pdf)
+* [Buildfarm](https://bazelbuild.github.io/bazel-buildfarm/docs/architecture/queues/)
+* [Buildbarn](https://github.com/buildbarn/bb-remote-execution/blob/master/pkg/proto/remoteworker/remoteworker.proto)
 * [BuildGrid](https://buildgrid.build/developer/data_model.html#rwapi)
 
 ## API Community
--- a/README.md
+++ b/README.md
@@ -53,8 +53,7 @@ These tools use the Remote Execution API to distribute builds to workers.
 
 ### Servers
 These applications implement the Remote Execution API to serve build requests
-from the clients above. These are then distributed to workers; some of these
-workers implement the Remote Worker API.
+from the clients above. Client requests are then distributed to workers.
 
 * [bazel-remote](https://github.com/buchgr/bazel-remote) (open source, cache only)
 * [Buildbarn](https://github.com/buildbarn) (open source)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -68,6 +67,18 @@ workers implement the Remote Worker API.
...
+### Workers
+Servers are responsible for managing a pool of workers.
+The Remote Worker API defines a reference protocol for worker and server
+communication, although, adhering to this protocol is not a requirement.
+Typically, workers are domain specific, optimized for custom scheduling
+protocols defined by the server implementation.
+
+* [Remote Worker API](https://goo.gl/oojM5H)
+* [Buildfarn](https://bazelbuild.github.io/bazel-buildfarm/docs/architecture/queues/)
+* [Buildbarn](https://nluug.nl/bestanden/presentaties/2023-05-11-ed-schouten-buildbarn-a-distributed-build-cluster.pdf)
```
EdSchouten, line 79: Please link to this instead: https://github.com/buildbarn/bb-remote-execution/blob/master/pkg/proto/remoteworker/remoteworker.proto

EdSchouten (approved):

cal-pratt:
Still figuring out the CLA stuff. Might have to remake this PR on my company email.

README.md
```diff
@@ -68,6 +67,18 @@ workers implement the Remote Worker API.
...
 * [Turbo Cache](https://github.com/allada/turbo-cache) (open source)
 
+### Workers
+Servers are responsible for managing a pool of workers.
+The Remote Worker API defines a reference protocol for worker and server
+communication, although, adhering to this protocol is not a requirement.
+Typically, workers are domain specific, optimized for custom scheduling
+protocols defined by the server implementation.
+
+* [Remote Worker API](https://goo.gl/oojM5H)
```
mostynb, line 77: I think it might be worth using the full (if ugly) google doc URL, in case goo.gl goes away. Also, could this be replaced with a link to the first mention of "Remote Worker API" above?

README.md
```diff
@@ -68,6 +67,18 @@ workers implement the Remote Worker API.
...
+* [Buildfarm](https://bazelbuild.github.io/bazel-buildfarm/docs/architecture/queues/)
+* [Buildbarn](https://github.com/buildbarn/bb-remote-execution/blob/master/pkg/proto/remoteworker/remoteworker.proto)
+* [BuildGrid](https://buildgrid.build/developer/data_model.html#rwapi)
```
mostynb, line 80: I think the link texts should be more specific than just the server name, since you are referring to a specific part of the implementation.

README.md
```diff
@@ -53,8 +53,7 @@ These tools use the Remote Execution API to distribute builds to workers.
 
 ### Servers
 These applications implement the Remote Execution API to serve build requests
-from the clients above. These are then distributed to workers; some of these
-workers implement the Remote Worker API.
+from the clients above. Client requests are then distributed to workers.
```
mostynb, line 56: I would drop the last sentence ("Client requests are then distributed to workers." - see the comment below.

README.md
```diff
@@ -68,6 +67,18 @@ workers implement the Remote Worker API.
 * [Scoot](https://github.com/twitter/scoot) (open source)
 * [Turbo Cache](https://github.com/allada/turbo-cache) (open source)
 
+### Workers
+Servers are responsible for managing a pool of workers.
```
bergsieker, line 71: I don't think I'd phrase it in this way. Not all server implementations "manage" the worker pools, and it would be a legitimate use of the API to multiplex work from many clients onto a single, powerful/rare machine (and thus not a pool, per se). Maybe "Servers generally distribute work to a fleet of workers. While there is no standard API for communication between the server and workers, links to the APIs from some existing implementations are provided as a reference below. Adhering to any one of these protocols is not a requirement."

README.md
```diff
@@ -68,6 +67,18 @@ workers implement the Remote Worker API.
...
+The Remote Worker API defines a reference protocol for worker and server
+communication, although, adhering to this protocol is not a requirement.
+Typically, workers are domain specific, optimized for custom scheduling
+protocols defined by the server implementation.
```
bergsieker, line 75: ..."Because workers are domain-specific, there is little value in standardizing on a single API for workers."

README.md
```diff
@@ -68,6 +67,18 @@ workers implement the Remote Worker API.
 * [Scoot](https://github.com/twitter/scoot) (open source)
 * [Turbo Cache](https://github.com/allada/turbo-cache) (open source)
 
+### Workers
+Servers are responsible for managing a pool of workers.
+The Remote Worker API defines a reference protocol for worker and server
+communication, although, adhering to this protocol is not a requirement.
```
bergsieker, line 73: "It was proposed as a standard at one point, but was deemed too heavyweight for most use-cases."

bergsieker (commented):
Thanks for providing this! I've left some suggested wording updates. Mostly, I think it's valuable to be more explicit that having a standard API for workers has relatively little value, given that we expect workers to be specific to the server implementation.

cal-pratt:
Reopened as #290

PR #290, opened 2024-03-11
Add readme notes for worker apis

cpratt34:
Replaces #289 
CLA should be happy now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,8 +53,7 @@ These tools use the Remote Execution API to distribute builds to workers.
 
 ### Servers
 These applications implement the Remote Execution API to serve build requests
-from the clients above. These are then distributed to workers; some of these
-workers implement the Remote Worker API.
+from the clients above.
 
 * [bazel-remote](https://github.com/buchgr/bazel-remote) (open source, cache only)
 * [Buildbarn](https://github.com/buildbarn) (open source)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -68,6 +67,26 @@ workers implement the Remote Worker API.
...
+### Workers
+Servers generally distribute work to a fleet of workers. While there is no
+standard API for communication between the server and workers, links to the
+APIs from some existing implementations are provided as a reference below.
+
+The [Remote Worker API](https://docs.google.com/document/d/1s_AzRRD2mdyktKUj2HWBn99rMg_3tcPvdjx3MPbFidU) 
+defines a generic protocol for worker and server communication, although, 
+this API is considered too heavyweight for most use-cases.
+
+*Adhering to any one of these protocols is not a requirement.*
```
EdSchouten, line 79: Nit: I agree with all of the sentences under this heading, but I think that the order in which they are provided can be improved. What's your thought on the following?

> Servers generally distribute work to a fleet of workers. The [Remote Worker API](https://docs.google.com/document/d/1s_AzRRD2mdyktKUj2HWBn99rMg_3tcPvdjx3MPbFidU) defines a generic protocol for worker and server communication, although, this API is considered too heavyweight for most use-cases. Because of that, many implementations have designed their own protocols. Links to these APIs are provided as a reference below. *Adhering to any one of these protocols is not a requirement.*
>
> - List of protocols goes here.

EdSchouten (approved):

README.md
cpratt34: I like it. Updated!

EdSchouten (approved):
@bergsieker PTAL!

bergsieker (approved):

PR #294, opened 2024-04-18
remote_asset: link digest_function & checksum qual

sluongng:
Ensure that the digest function in requests is consistent with the
digest function used in `checksum.sri` qualifier, this way server
implementations do not have to support replicating blob from one digest
function to another.

Such functionality is not the intended use case of Remote Asset APIs
and, instead, could be achieved via a different client-side API call.

### DIFF
--- a/build/bazel/remote/asset/v1/remote_asset.proto
+++ b/build/bazel/remote/asset/v1/remote_asset.proto
@@ -205,6 +205,8 @@ message FetchBlobRequest {
   // The digest function the server must use to compute the digest.
   //
   // If unset, the server SHOULD default to SHA256.
+  // If set, the function MUST be the same with the digest function used in
+  // "checksum.sri", the subresource integrity qualifier.
   build.bazel.remote.execution.v2.DigestFunction.Value digest_function = 6;
 }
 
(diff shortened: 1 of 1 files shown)


juergbi:
I agree with Ed. And being able to fetch assets from servers that are not tied to the CAS server at all (and thus, may make their own choice of checksum algorithm) absolutely is a valid use case for the Remote Asset API as it was envisioned.

> this way server implementations do not have to support replicating blob from one digest function to another.

Remote Asset API servers are not required to support fetching arbitrary files from arbitrary (HTTP) servers. I.e., it can be acceptable to only have limited support for HTTP servers, as long as that's clearly documented to potential clients/users. However, we shouldn't restrict the protocol in such a way, in my opinion.

sluongng:
Our current draft implementation does support that behavior: fetching with a SHA256 checksum qualifier while expecting a BLAKE3 hash.

However, such logic is a bit harder to maintain and such a use case could be relatively rare in reality. It would require a client to know and use both digest functions at the same time: Requested blob with SHA256 hash, downloaded blob using BLAKE3 digest, and re-hash blob against to validate SHA256 on the client side. Such a workflow looks like a convoluted footgun, making it harder for other client/server implementations to follow the spec.

EdSchouten:
> Our current draft implementation does support that behavior: fetching with a SHA256 checksum qualifier while expecting a BLAKE3 hash.
> 
> However, such logic is a bit harder to maintain and such a use case could be relatively rare in reality.

Is it really that rare? Bazel WORKSPACE/MODULE.bazel files often contain SHA-256 hashes. But the user has the freedom to invoke Bazel with any remote digest function it wants.

PR #298, opened 2024-06-05
Update go dependencies to remove use of deprecated googleapi go modules

bentekkie:
"google.golang.org/genproto/googleapis/longrunning" is deprecated in favor of "cloud.google.com/go/longrunning/autogen/longrunningpb"
"google.golang.org/genproto/googleapis/api" and "google.golang.org/genproto/googleapis/rpc" are separate go module now
Use of these old dependencies is complicating updating dependencies in remote-apis-sdks and reclient


mostynb:
I think this might have been done in #307.

PR #302, opened 2024-07-03
chore: add Aspect Build to server providers

CaerusKaru:
Aspect Build offers remote server resources providing both remote cache and remote execution capabilities.


CaerusKaru:
cc @alexeagle @gregmagolan

alexeagle:
LGTM thanks!

PR #303, opened 2024-07-09
remote_execution: add routing header to ByteStream

sluongng:
Add optional routing header for ByteStream RPCs.

### DIFF
--- a/build/bazel/remote/execution/v2/remote_execution.proto
+++ b/build/bazel/remote/execution/v2/remote_execution.proto
@@ -325,12 +325,12 @@ service ActionCache {
 // each of the compression formats that the server supports, as well as in
 // uncompressed form.
 //
-// Additionally, ByteStream RPCs MAY come with an additional plain text header
+// Additionally, ByteStream requests MAY come with an additional plain text header
 // that indicates the `resource_name` of the blob being sent.  The header, if
 // present, MUST follow the following convention:
 // * name: `build.bazel.remote.execution.v2.resource-name`.
 // * contents: the plain text resource_name of the request message.
-// The contents of the header MUST match the `resource_name` of the request
+// If set, the contents of the header MUST match the `resource_name` of the request
 // message.  Servers MAY use this header to assist in routing requests to the
 // appropriate backend.
 //
--- a/build/bazel/remote/execution/v2/remote_execution.proto
+++ b/build/bazel/remote/execution/v2/remote_execution.proto
@@ -325,6 +325,15 @@ service ActionCache {
 // each of the compression formats that the server supports, as well as in
 // uncompressed form.
 //
+// Additionally, ByteStream RPCs MAY come with an additional plain text header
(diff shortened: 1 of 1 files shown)


build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -2133,3 +2133,14 @@ message RequestMetadata {
...
+
+// An optional Metadata to attach to any RPC request to tell the server about the blob
+// inside, or referenced by, a request.  The server may use this for better load 
+// balancing, or other purposes.
+//
+// To use it, the client attaches the header to the call using the canonical proto
+// serialization:
+//
+// * name: `build.bazel.remote.execution.v2.blob-digest`
+// * contents: the plained text digest of the blob in the form of "{hash}/{size}".
```
EdSchouten, line 2145: Instead of reinventing our own format for this string, any chance we can just reuse the existing bytestream path string? So prefixed with the instance name, potentially containing the digest function, etc.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -2133,3 +2133,14 @@ message RequestMetadata {
...
+message BlobDigest {}
```
EdSchouten, line 2146: Is this message type actually needed?

build/bazel/remote/execution/v2/remote_execution.proto
EdSchouten: In that case probably best to call this `build.bazel.remote.execution.v2.resource-name`?

sluongng:
@EdSchouten I have changed the PR to indicate that we want a resource-name header instead. Let me know what you think.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -325,6 +325,15 @@ service ActionCache {
 // each of the compression formats that the server supports, as well as in
 // uncompressed form.
 //
+// Additionally, ByteStream RPCs MAY come with an additional plain text header
```
EdSchouten, line 328: s/RPCs/requests?

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -325,6 +325,15 @@ service ActionCache {
...
+// that indicates the `resource_name` of the blob being sent.  The header, if
+// present, MUST follow the following convention:
+// * name: `build.bazel.remote.execution.v2.resource-name`.
+// * contents: the plain text resource_name of the request message.
+// The contents of the header MUST match the `resource_name` of the request
```
EdSchouten, line 333: s/The contents/If set, the contents/

EdSchouten (approved):

build/bazel/remote/execution/v2/remote_execution.proto
sluongng: Good call. Fixed!

EdSchouten (approved):

PR #307, opened 2024-08-13
Migrate remote-apis to bzlmod, drop legacy support

philwo:
I initially just wanted to upgrade the dependencies in this project to their latest version, thinking it'll be a ~5 minute task.

After spending a whole day on trying to do it in a way that didn't feel completely hacky and dirty, I figured that migrating the total random chaos that is the WORKSPACE file and its "first one wins" and "remember, X comes before Y comes before my library's patched version of rules_proto, comes after protobuf's outdated copy of it, but not if your rules_go's commit hash divided by 3.141 is an even number [...]" semantics to bzlmod is actually easier.

In summary, migrating from WORKSPACE to bzlmod removes a lot of complexity:
- All of the manual tinkering with WORKSPACE dependencies is gone and replaced with just a short MODULE.bazel file. No more wondering about the correct ordering of declaring repositories. Fully automatic reliable version resolution.
- Removed remote_apis_deps.bzl and repository_rules.bzl files, as they are no longer necessary (and incompatible with bzlmod).
- Reverted the split into language-specific subdirectories. I understand that this was done to avoid pulling in Go as an unnecessary dependency for C++ and Java users. However, I'm not sure if it's worth the maintenance cost of having our own grpc_cc_library rule - which broke after upgrading gRPC to the latest version, and I couldn't figure out how to fix it. Now that bzlmod removes the pain of adding and maintaining rules_go etc. as a dependency to one's project, maybe it's no longer worth the trouble? Let me know if you think this is a problem, though.

I also upgraded all Bazel and go.mod dependencies to the latest versions and migrated all deprecated dependencies to their replacements (such as longrunningpb from go-genproto to its new home in cloud.google.com and the api and rpc stuff to their new independent modules).

Finally, I added a .bazelversion for the project to pin it to the latest Bazel LTS 7.3.0 version.

There were a couple of incompatible changes I also took care of, like having to specify `--output_groups=go_generated_srcs` in our hook script that generates the *.pb.go files, because otherwise the files were missing after the build (I guess Bazel doesn't copy them out of the sandbox unless we explicitly request them as output files). have a PR for remote-apis-sdks, too, which migrates that to bzlmod as well and updates it to use the latest version of this repo.


build/bazel/semver/BUILD
```diff
@@ -1,30 +1,37 @@
...
+# Go
+go_proto_library(
     name = "semver_go_proto",
-    actual = "//build/bazel/semver/go:semver_go_proto",
+    importpath = "github.com/bazelbuild/remote-apis/build/bazel/semver",
+    proto = ":semver_proto",
 )
 
-alias(
+go_library(
```
EdSchouten, line 33: What's the reason we're not using alias() here? This means that if you end up depending on both `semver_go_proto` and `go_default_library`, you end up getting a bunch of warnings along the lines of:
```
DEBUG: /private/var/tmp/_bazel_ed/9dcffd914bcef91974dd3105239de9ed/external/rules_go~/go/private/actions/link.bzl:59:18: Multiple copies of github.com/bazelbuild/remote-apis/build/bazel/remote/execution/v2 passed to the linker. Ignoring bazel-out/darwin_arm64-fastbuild/bin/external/bazel_remote_apis~/build/bazel/remote/execution/v2/remote_execution_go_proto.a in favor of bazel-out/darwin_arm64-fastbuild/bin/external/bazel_remote_apis~/build/bazel/remote/execution/v2/go_default_library.a
```

philwo: @EdSchouten I don't fully remember. This was the best result I could come up with after spending two days tinkering with this stuff, and trying to make it as clean as possible. There was also a related change for the `remote-apis-sdks` side of things, but I might have forgotten to `git push` that.. 🙈

If you have an idea how to improve upon this, by all means, please go ahead. Happy to review and/or try something out if that helps!

EdSchouten: No worries; On the Buildbarn side I'm simply adding the following directives to BUILD:
```
# gazelle:resolve go github.com/bazelbuild/remote-apis/build/bazel/remote/execution/v2 @bazel_remote_apis//build/bazel/remote/execution/v2:go_default_library
# gazelle:resolve go github.com/bazelbuild/remote-apis/build/bazel/semver @bazel_remote_apis//build/bazel/semver:go_default_library
# gazelle:resolve proto build/bazel/remote/execution/v2/remote_execution.proto @bazel_remote_apis//build/bazel/remote/execution/v2:remote_execution_proto
# gazelle:resolve proto go build/bazel/remote/execution/v2/remote_execution.proto @bazel_remote_apis//build/bazel/remote/execution/v2:remote_execution_go_proto
```
This ensures that both proto rules and plain Go libraries depend on the non-`go_default_library` target.

PR #309, opened 2024-09-10
rename module to remote-apis

mmorel-35:
Hi @philwo , @keith,
As the module is not yet published to the bcr, I would to suggest to keep the name aligned with the project in github.
Also it’s not yet really recommanded to persist the MODULE.bazel.lock so I removed it added it to gitignore


mostynb:
> Also it’s not yet really recommanded to persist the MODULE.bazel.lock so I removed it added it to gitignore

Do you have more info about why this is not recommended? https://bazel.build/external/lockfile lists the opposite in its best practices:
> Include the lockfile in version control to facilitate collaboration and ensure that all team members have access to the same lockfile, promoting consistent development environments across the project.

keith:
I think as of bazel 7.2.x+ it's ok to check in the lockfile, but previously as Matthieu said this file had a lot of thrash and might change across platforms, there's also probably not a ton of value in maintaining it in a repo like this, since its really the end user's lockfile that matters

EdSchouten:
I'd actually be in favor of keeping the "bazel_" prefix. As the name implies, BCR is a central registry. There are many systems apart from Bazel that have "remote APIs".

bergsieker:
Adding Tiago from the Bazel team to help us figure out what the best practices for BCR are. Browsing the existing modules, I see some with relatively generic names, but none quite as likely to be overloaded as "remote-apis." At the same time, I like *not* having the "bazel" prefix as a way to emphasize that this API is not specific to or defined by Bazel.

tjgq:
@Wyverald Can you advise on BCR module naming?

Wyverald:
I like `bazel_remote_apis`. Just `remote_apis` seems too generic.

If this is not Bazel-specific, I'm less sure what it should be named. `remote-build-apis`? How are these APIs referred to usually, outside of a Bazel-specific context?

(BTW hypher vs underscore is irrelevant, both are fine.)

bergsieker:
I suggest that we stick with bazel_remote_apis. It's not ideal to have the module and repo name not match, but as far as I can tell it's a supported path.

Practically, the remote-apis repo lives under bazelbuild. Unless and until we move the repo out of that organization, there's already an explicit association with Bazel, so there's not much point to avoiding "bazel" in the module name.

PR #312, opened 2024-09-24
Improve phrasing of digest_function added in #311

EdSchouten:
- Unlike the other messages where a digest_function field was added, ExecuteActionMetadata is returned by the server to the client. This means that the words "client" and "server" need to be swapped around in some but not all places.

- To ensure backward compatibility, we permit digest functions that were defined at the time this field was added to remain unset. When the digest_function fields were added to the other messages, the last one to be added was MURMUR3. In the meantime we've added BLAKE3 and SHA256TREE, so for this specific field we must list those as well.

### DIFF
--- a/build/bazel/remote/execution/v2/remote_execution.pb.go
+++ b/build/bazel/remote/execution/v2/remote_execution.pb.go
@@ -2799,11 +2799,11 @@ type ExecuteOperationMetadata struct {
 	PartialExecutionMetadata *ExecutedActionMetadata `protobuf:"bytes,5,opt,name=partial_execution_metadata,json=partialExecutionMetadata,proto3" json:"partial_execution_metadata,omitempty"`
 	// The digest function that was used to compute the action digest.
 	//
-	// If the digest function used is one of MD5, MURMUR3, SHA1, SHA256,
-	// SHA384, SHA512, or VSO, the client MAY leave this field unset. In
-	// that case the server SHOULD infer the digest function using the
-	// length of the action digest hash and the digest functions announced
-	// in the server's capabilities.
+	// If the digest function used is one of BLAKE3, MD5, MURMUR3, SHA1,
+	// SHA256, SHA256TREE, SHA384, SHA512, or VSO, the server MAY leave
+	// this field unset. In that case the client SHOULD infer the digest
+	// function using the length of the action digest hash and the digest
+	// functions announced in the server's capabilities.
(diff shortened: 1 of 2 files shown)


EdSchouten:
@werkt @bergsieker I spotted a couple of small mistakes in the way the new `digest_function` was documented. PTAL!

werkt:
The reason to list omittable digest functions is not just legacy - sha256 cannot be distinguished from blake3 by digest length, so you can't omit both.

EdSchouten:
That also applies to existing cases, as `MURMUR3` cannot be distinguished from `MD5`, due to both of them being 128 bits in size. We already discussed this issue during one of the monthly meetings when we originally added `digest_functions`, and the outcome was that we preferred to remain backward compatible, even if it meant that ambiguity remained if sender doesn't set this field. The way this field is documented right now, we are breaking compatibility.

werkt:
If MURMUR3 is also 128 bits, then yes, it should be removed from the omittable list as well.

EdSchouten:
No, it should not, because that would completely break backward incompatibility between existing clients and servers.

There are already servers out there that support BLAKE3. And those can already return ExecuteActionMetadata messages that have `action_digest.hash` set to a BLAKE3 hash, while having the newly added `digest_function` field left unset. Are those now suddenly in violation of the protocol? Should a client that fully complies to the latest protocol simply fail to parse these messages? And should they also fail if the server supports nothing else than BLAKE3?

werkt:
Not having written the original copy, I can't say for sure what the intent was. Based on the specification it seems to be a tenured, bit-count matching exclusion of later-numbered additions to DigestFunction.Value - newer sequence numbers without matching bit-counts for lower numbers MUST be omitted, but that is certainly not a formula expressed in docs or anything else I've seen.

Of note, the language actually doesn't require those things NOT in the omitted list to NOT be omitted (i.e. specified). It is within spec (as much as our comments are spec) for someone to omit DigestFunction.Value.BLAKE3 from a message with a relevant blake3 hash in it.

EdSchouten:
> Of note, the language actually doesn't require those things NOT in the omitted list to NOT be omitted (i.e. specified). It is within spec (as much as our comments are spec) for someone to omit DigestFunction.Value.BLAKE3 from a message with a relevant blake3 hash in it.

That is just being overly pedantic. The way the sentence is structured, the first "If" is supposed to be interpreted as "if and only if". Otherwise the sentence would be a tautology.

mostynb (approved):

EdSchouten:
Thanks for the review, @mostynb.

PR #313, opened 2024-10-01
publish-to-BCR setup

Wyverald:
@bergsieker: I volunteered you to be a module maintainer :) it should be very low maintenance, but let me know if you would rather not.

We can cut a release after this gets merged and watch it show up in BCR. (Note that we'll need to attach a release archive -- I can do this part.)

### DIFF
--- a/.bcr/config.yml
+++ b/.bcr/config.yml
@@ -1,0 +1,4 @@
+fixedReleaser:
+  login: "Wyverald"
+  email: "wyverald@gmail.com"
+
--- a/.bcr/metadata.template.json
+++ b/.bcr/metadata.template.json
@@ -1,0 +1,19 @@
+{
+  "homepage": "https://github.com/bazelbuild/remote-apis",
+  "maintainers": [
+    {
+      "email": "wyverald@gmail.com",
+      "github": "Wyverald",
+      "name": "Xudong Yang"
+    },
+    {
+      "github": "berksieker",
+      "name": "Steven Bergsieker"
+    }
+  ],
+  "versions": [],
+  "yanked_versions": {},
+  "repository": [
+    "github:bazelbuild/remote-apis"
+  ]
+}
--- a/.bcr/presubmit.yml
+++ b/.bcr/presubmit.yml
@@ -1,0 +1,11 @@
+matrix:
+  platform: ["ubuntu2204", "macos", "macos_arm64", "windows"]
+  bazel: [7.x]
+tasks:
+  build_targets:
+    name: build targets
+    platform: ${{ platform }}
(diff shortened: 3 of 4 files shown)


.bcr/metadata.template.json
```diff
@@ -0,0 +1,19 @@
+{
+  "homepage": "https://github.com/bazelbuild/remote-apis",
+  "maintainers": [
+    {
+      "email": "wyverald@gmail.com",
+      "github": "Wyverald",
+      "name": "Xudong Yang"
+    },
+    {
+      "github": "berksieker",
```
bergsieker, line 10: s/bergksieker/bergsieker/

.bcr/metadata.template.json
```diff
@@ -0,0 +1,19 @@
...
+      "name": "Steven Bergsieker"
```
bergsieker, line 11: Should you add bergsieker@google.com as an email contact?

.bcr/config.yml
bergsieker, line 1: Where can I find the expected format and information for these files?

bergsieker (changes requested):

.bcr/config.yml
Wyverald: https://github.com/bazel-contrib/publish-to-bcr/tree/main/templates

bentekkie:
> hmm, maybe this is a good opportunity to actually split this module into several ones (one per language). For example, Bazel's own use of remote-apis requires the Java gRPC targets, which aren't currently in this repo. I'll probably send another change tomorrow.

Can this be done orthogonally to getting this into bcr? Reclient and remote-apis-sdks's usage of this module would be greatly simplified if we can get this deployed to bcr.

bergsieker:
@Wyverald can we pick this up again?

Wyverald:
Sorry, I've been busy with managing Bazel's 8.0 release. Are you blocked on this change? I think you mentioned a few action items from your side in a previous email thread (including setting a new release schedule, etc). Are those completed already?

bergsieker:
> Sorry, I've been busy with managing Bazel's 8.0 release. Are you blocked on this change? I think you mentioned a few action items from your side in a previous email thread (including setting a new release schedule, etc). Are those completed already?

We've agreed to move to a monthly release model, skipping months where there are no (meaningful) commits). We're also moving to date-based versioning. Ideally we'll cut the first such release around next Tuesday. Is it better to get this in place before that release, or do we need to wait for that release and use it as a baseline for the first BCR release the following month?

mbland:
FYI, just opened #370 to pick this up, with @Wyverald's blessing.

PR #320, opened 2024-11-21
Update the doc of ExecuteOperationMetadata

vanja-p:
It currently says that that ExecuteOperationMetadata is contained in the metadata field, but then it links to the response field.


vanja-p:
I just filled out the CLA

PR #325, opened 2025-01-20
Add max_blob_size_bytes to CacheCapabilities

mostynb:
Some CAS implementations place an upper limit on the size of blobs that they will accept in uploads. There is currently no standard way for clients to query cache servers for this limit, and rejected large uploads can be slow.

So let's add this to the CacheCapabilities API. For backwards compatibility, a value of 0 means "no specified limit" (ie not "no uploads allowed").

* Bazel-remote has a `--max_blob_size` flag to specify an optional limit.
* Buildfarm appears to check against a blob size limit here: https://github.com/buildfarm/buildfarm/blob/4c539f6655e70b571f0604a31f8a15a8fb0d5d99/src/main/java/build/buildfarm/instance/shard/ServerInstance.java#L1454


mostynb:
This seems to be a duplicate of #278, closing in favour of that.

Issue #297, opened 2024-05-30
Cache Capability to indicate that CAS is read-only


sluongng:
We need to signal to the client that in some configurations, the CAS will only accept read requests. All write requests will result in errors.

Given that we are currently advertising

```protobuf
// Describes the server/instance capabilities for updating the action cache.
message ActionCacheUpdateCapabilities {
  bool update_enabled = 1;
}
```

in `CacheCapabilities` proto. I was thinking that we could simply add another identical field for CASUpdateCapabilities.

However, such a field would default to `false`, which is troublesome for backward compatibility. And doing `bool is_readonly = 1;` feels out of spec. Looking for feedback on this design.

EdSchouten:
Use an enum where 0 indicates it’s writable?

bergsieker:
I think the right course is an enum with UNKNOWN=0, WRITEABLE, READ-ONLY as the values. It's possible there are other CAS read-only CAS implementations (e.g., local caches), and we shouldn't extend the API in a way that invalidates their position by default.

PR #328, opened 2025-02-12
Add max_cas_blob_size_bytes to CacheCapabilities message (continuation)

axeluhlig:
Continuing the work from https://github.com/bazelbuild/remote-apis/pull/278


build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -2056,6 +2056,17 @@ message CacheCapabilities {
   // [BatchUpdateBlobs][build.bazel.remote.execution.v2.ContentAddressableStorage.BatchUpdateBlobs]
   // requests.
   repeated Compressor.Value supported_batch_update_compressors = 7;
+
+  // Maximum size a single CAS blob can have in the remote cache system.
```
mostynb, line 2060: This sentence contradicts the last sentence in this block. How about:
```suggestion
  // The maximum size that the server will accept for CAS blob uploads.
```

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -2056,6 +2056,17 @@ message CacheCapabilities {
...
+  // - If it is 0, it means there is no limit is set. A client may assume
```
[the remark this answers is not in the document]
axeluhlig: Thanks!

build/bazel/remote/execution/v2/remote_execution.proto
axeluhlig: Thanks!

bergsieker (approved):
Other than one nit this LGTM. I'd like to get one other person to look at it before merging.

mostynb (approved):
lgtm if @bergsieker's nit is fixed

sluongng:
I wonder if there is a value that can be added to the spec that would have solved my issue as well https://github.com/bazelbuild/remote-apis/issues/297

Some context: our server has a fine-grained access level for each API key, and one level could grant CAS-readonly and reject all uploads. So with this change, I can see a quick hack where we set the value to 1 byte to signal the client to skip all uploads. But if there is a `-1` value added to the spec, I think it would be better.

bergsieker:
> I wonder if there is a value that can be added to the spec that would have solved my issue as well #297
> 
> Some context: our server has a fine-grained access level for each API key, and one level could grant CAS-readonly and reject all uploads. So with this change, I can see a quick hack where we set the value to 1 byte to signal the client to skip all uploads. But if there is a `-1` value added to the spec, I think it would be better.

I think it would be better to keep "permission" separate from "limits." Presumably you want to have some client-side logic that says something like "These blobs should be uploaded if possible, but it's not an error if I can't upload them because I don't have permission." That's pretty different from "These blobs should be uploaded if possible, but if the CAS can't support them don't trigger an error."

PR #330, opened 2025-03-11
[remote asset api] Clarify encoding of checksum.sri

malt3:
The SRI specification specifically allows for more than one checksum to be part of integrity metadata (https://www.w3.org/TR/SRI/). The possible presence of more than one checksum is not immediately obvious without reading the specification, so we add a clarification.

### DIFF
--- a/build/bazel/remote/asset/v1/qualifiers.md
+++ b/build/bazel/remote/asset/v1/qualifiers.md
@@ -28,7 +28,9 @@ The following standard qualifier `name`s are defined:
   ```   
 
 * `checksum.sri`: The value represents a [Subresource Integrity](https://www.w3.org/TR/SRI/)
-  checksum of the content.
+  checksum of the content. Multiple checksums may be specified, separated by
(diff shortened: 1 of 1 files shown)


build/bazel/remote/asset/v1/qualifiers.md
```diff
@@ -28,7 +28,8 @@ The following standard qualifier `name`s are defined:
   ```   
 
 * `checksum.sri`: The value represents a [Subresource Integrity](https://www.w3.org/TR/SRI/)
-  checksum of the content.
+  checksum of the content. Multiple checksums may be specified, separated by
+  whitespace.
```
bergsieker, line 32: Because that spec is obtuse enough, it's probably worth adding something like "The Qualifier is satisfied if the server validates at least one of the provided checksums, although it may choose to validate more than one."

malt3: Sounds good to me. Added.

bergsieker (approved):

PR #333, opened 2025-04-11
platform.md: Define RemoteApisSocketPath for nested remote execution

juergbi:
This defines `RemoteApisSocketPath` as a new platform property to optionally provide remote commands access to the Remote APIs via a UNIX socket at a specified path.

This is intended to e.g. support running Bazel as a remote command while allowing the remote Bazel to then use nested remote execution or remote caching for individual actions. Or to use recc to speed up compilations within a package build that is also using remote execution (e.g. recc in BuildStream).

### DIFF
--- a/build/bazel/remote/execution/v2/platform.md
+++ b/build/bazel/remote/execution/v2/platform.md
@@ -124,3 +124,12 @@ The following standard property `name`s are defined:
     ]
   }
   ```
+
+* `RemoteApisSocketPath`: This requests Remote APIs to be made available in
+  the execution environment via a UNIX socket at the specified path. The
+  path is relative to the input root.
+
+  The APIs and capabilities offered is implementation-specific and servers
+  MAY impose additional restrictions and/or resource limits on nested remote
+  execution clients. The Capabilities API may be used by remote commands to
+  query server properties.


EdSchouten:
Why permit the client to select the path of the UNIX socket? Wouldn’t it make more sense for the worker to create one and subsequently set it as part of some envvar?

Note that binding UNIX sockets to paths in excess of 108 bytes in size is problematic, due to size limits on sockaddr_un.sun_path. It may therefore make more sense to let the server pick a path, which may pick a path outside the input root that is shorter.

bergsieker:
My first instinct is the same as Ed's, that the worker should choose the socket location and pass it to the job via something like an environment variable. On reflection, however, I think I agree with your reasoning that it's reasonable for the client to choose the location such that it won't conflict with the input tree.

I think my bigger question is whether this truly belongs in the lexicon. It's only applicable to Unix-like systems, and it's a fairly specific capability. Of course, the lexicon is just a guideline, not a true API specification, so maybe the bar for inclusion is lower. What's the benefit of including this in the lexicon?

juergbi:
> What's the benefit of including this in the lexicon?

The benefit would be for clients that need/want to use nested remote execution to be compatible with all worker implementations that support nested remote execution. I.e., avoid the need for clients to have logic or configuration specific to a particular server implementation.

I have a draft branch that implements this in BuildBox and the plan is to add client-side support for this in BuildStream. It would be nice if BuildStream projects using this feature could also use other RE workers that support nested remote execution. And I'm aware of nested remote execution being used without BuildStream, i.e., this is not something that is useful only to BuildStream.

I don't know whether other implementations are interested in adding support for this (or already have support for nested remote execution with a non-standard configuration), but I think it would be unfortunate if multiple implementations support the same basic functionality yet are incompatible without good reason.

It would be great to hear from interested developers of other worker implementations.

sluongng:
IMO: this is a bit too niche. A more generic approach would be a platform property that signals the sandbox to enable/disable networking. That way the action could simply reach the remote build server via the official grpcs/https protocol.

The worker implementation could opt for additional networking restrictions. An example talk from BazelCon 2023 mentioning this here https://youtu.be/tNY7qiHOdoY?list=PLbzoR-pLrL6rUiqylH-kumoZCWntG1vjp&t=321.

juergbi:
> That way the action could simply reach the remote build server via the official grpcs/https protocol.

The regular remote execution endpoint may require authentication that the action command may not have access to.

And even embedding the remote execution endpoint itself may not be ideal as this would affect the cache key.

PR #336, opened 2025-07-08
Chunking Algorithms

sluongng:
Based on #282

Introduce ChunkingFunction which enum is a set of known chunking
algorithms that the server can recommend to the client.

Provide FastCDC_2020 as the first explicit chunking algorithm.

The server advertises these through a new chunking_configuration field in
CacheCapabilities message. There, the server may set the chunking
functions that it supports as well as the relevant configuration
parameters for that chunking algorithm.

---

I recommend reading https://joshleeb.com/posts/fastcdc.html to understand more about the available FastCDC configuration parameters.


build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -439,6 +439,115 @@ service ContentAddressableStorage {
...
+  // download network traffic, when downloading large blobs from the CAS.
+  // However, clients MUST first check the server capabilities, whether blob
+  // splitting is supported by the server.
+  //
+  // Clients SHOULD verify whether the digest of the blob assembled by the
+  // fetched chunks results in the requested blob digest.
+  //
+  // Since the generated chunks are stored as blobs, they underlie the same
+  // lifetimes as other blobs. In particular, the chunk lifetimes are
+  // independent from the lifetime of the original blob:
```
mostynb, line 478: This wording is a bit complex. How about "The lifetimes of the generated chunk blobs MAY be independent of the lifetime of the original blob. In particular:"

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -439,6 +439,115 @@ service ContentAddressableStorage {
...
+  //  * A blob and any chunk derived from it may be evicted from the CAS at
+  //    different times.
+  //  * A call to Split extends the lifetime of the original blob, and sets
```
mostynb, line 481: Isn't the api called SplitBlob, rather than Split? Might be good to add a link here too.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -439,6 +439,115 @@ service ContentAddressableStorage {
...
+  // However, clients MUST first check the server capabilities, whether blob
+  // splitting is supported by the server.
+  //
+  // Clients SHOULD verify whether the digest of the blob assembled by the
+  // fetched chunks results in the requested blob digest.
+  //
+  // Since the generated chunks are stored as blobs, they underlie the same
+  // lifetimes as other blobs. In particular, the chunk lifetimes are
+  // independent from the lifetime of the original blob:
+  //  * A blob and any chunk derived from it may be evicted from the CAS at
```
mostynb, line 479: s/may/MAY/

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -439,6 +439,115 @@ service ContentAddressableStorage {
...
+  //    different times.
+  //  * A call to Split extends the lifetime of the original blob, and sets
+  //    the lifetimes of the resulting chunks (or extends the lifetimes of
+  //    already-existing chunks).
+  //  * Touching a chunk extends its lifetime, but does not extend the
```
mostynb, line 484: s/but does not extend the/but the server MAY choose not to extend the/ ?

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -439,6 +439,115 @@ service ContentAddressableStorage {
...
+  //    lifetime of the original blob.
+  //  * Touching the original blob extends its lifetime, but does not extend
```
mostynb, line 486: Similarly to the point above, how about "but the server MAY choose not to extend" ?

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -439,6 +439,115 @@ service ContentAddressableStorage {
...
+  //
+  // If a client needs to upload a large blob and is able to split a blob into
+  // chunks in such a way that reusable chunks are obtained, e.g., by means of
+  // content-defined chunking, it can first determine which parts of the blob
+  // are already available in the remote CAS and upload the missing chunks, and
+  // then use this API to instruct the server to splice the original blob from
+  // the remotely available blob chunks.
+  //
+  // Servers MAY implement this functionality, but MUST declare whether they
+  // support it or not by setting the
```
mostynb, line 518: For backwards compatibility, I think we can only require that servers which support this MUST declare it in the capabilities API.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -439,6 +439,115 @@ service ContentAddressableStorage {
...
+  // [CacheCapabilities.blob_splice_support][build.bazel.remote.execution.v2.CacheCapabilities.blob_splice_support]
+  // field accordingly.
+  //
+  // Clients MAY use this functionality, it is just an optimization to reduce
+  // upload traffic, when uploading large blobs to the CAS. However, clients
+  // MUST first check the server capabilities, whether blob splicing is
+  // supported by the server.
```
mostynb, line 525: How about "Clients MUST check that the server supports this capability, before deciding to use it" ?

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -439,6 +439,115 @@ service ContentAddressableStorage {
...
+  //
+  // In order to ensure data consistency of the CAS, the server MUST only add
+  // entries to the CAS under a hash the server verified itself. In particular,
+  // it MUST NOT trust the result hash provided by the client. The server MAY
```
mostynb, line 529: s/it/servers/

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -439,6 +439,115 @@ service ContentAddressableStorage {
...
+  // is obtained.
+  //
+  // When blob splitting and splicing is used at the same time, the clients and
+  // the server SHOULD agree out-of-band upon a chunking algorithm used by both
+  // parties to benefit from each others chunk data and avoid unnecessary data
+  // duplication.
+  //
+  // Errors:
+  //
+  // * `NOT_FOUND`: At least one of the blob chunks is not present in the CAS.
```
mostynb, line 543: Should the response somehow specify which blobs were not present?

sluongng:
@mostynb I think most of your comments are meant for https://github.com/bazelbuild/remote-apis/pull/282, which is what this PR is based on. Since #282 is merged, I have rebased this PR on top of the latest changes.

I would recommend creating a separate PR with the suggestions above.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1976,6 +1976,12 @@ message SplitBlobRequest {
   // length of the blob digest hashes and the digest functions announced
   // in the server's capabilities.
   DigestFunction.Value digest_function = 3;
+
+  // The chunking function that the client prefers to use.
+  //
+  // The server MAY use a different chunking function. The client MUST check
+  // the chunking function used in the response.
+  ChunkingFunction.Value chunking_function = 4;
```
tjgq, line 1984: 1. Is this field intended to be mandatory or optional (with the latter giving the server complete leeway in choosing a function)? If optional, can we document it?
2. When the field is present, should it be required to match one of the functions declared in the server capabilities?

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -2023,6 +2032,9 @@ message SpliceBlobRequest {
   // server SHOULD infer the digest function using the length of the blob digest
   // hashes and the digest functions announced in the server's capabilities.
   DigestFunction.Value digest_function = 4;
+
+  // The chunking function that the client used to split the blob.
+  ChunkingFunction.Value chunking_function = 5;
```
tjgq, line 2037: Why is this necessary? Isn't the result of a splice completely independent from the function originally used to do the splitting?

It also imposes a requirement that the chunks must have necessarily originated from a split operation, which seems counter to the spirit of the original proposal (that split and splice are independent optimizations).

tjgq (changes requested):

build/bazel/remote/execution/v2/remote_execution.proto
tjgq: Ah, it defaults to UNKNOWN if unset, because this is proto3. (But perhaps it's still worth spelling it out - up to you.)

build/bazel/remote/execution/v2/remote_execution.proto
tyler-french: A lot of the requirements for this come from the desire for the server to be able to store and retrieve information on how to convert between a blob and chunks, and act as a cache for that information.

I think this is important if a client is doing the chunking, and is telling the server to store information on how a certain blob can be reconstructed from chunks. Then, if someone calls Split on the same blob, the server can then return this information to the client, without needing to actually split the blob, given that the chunking operation is the same. Caches can also share these entries and do chunking async, etc, if the server and client are aligned on a reproducible way to create these chunks.

If the client calls Splice twice, with two different sets of digests, and two different chunking functions, the server can also make decisions on how it handles this. A server could choose to be very explicit on the types of chunking it supports, for example, if it needs to be able to reproduce them somewhere else, and so that it chooses to store chunks that are more likely to de-duplicate storage of content.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -2279,6 +2334,50 @@ message CacheCapabilities {
...
+  // The chunking configuration of the server.
+  ChunkingConfiguration chunking_configuration = 11;
+}
+
+// The chunking configuration of the server.
+message ChunkingConfiguration {
+  // If any of the advertised parameters are not within the expected range,
+  // the client SHOULD ignore FastCDC chunking function support.
+  message FastCDCParams {
+    // The normalization level for the FastCDC chunking algorithm.
```
tyler-french: We should consider having just `avg_chunk_size_bytes`, and all the other parameters are set to defulats and documented.

PR #353, opened 2025-11-25
Update `Split` and `Splice` docs

tyler-french:
This PR better aligns the language of the REv2 API to describe how a client should expect to interact with the `Split` and `Splice` APIs. 

Generally speaking, when designing a Remote Cache service, the server is not always primarily responsible for doing splitting and splicing blobs. In fact, the `Split` and `Splice` APIs are extremely helpful from a client's context to store and retrieve this manifest for how content defined chunking can compose a blob.

For example, if a client calls `Splice` that maps blob digest `A` to chunks `A1` and `A2`, this instructs the server to store this information. Later, if a client that is not chunking-aware calls `Read` on `A`, the server can use this stored state to compose `A` from `A1` and `A2` stored in the CAS, and serve it to the client.

Similarly, if a user calls `Split` on blob `B` (which could be some Action Result), the server would respond with its stored manifest: `B1` and `B2`. A chunking aware client can then skip downloading `B1` if it's available locally from some other file's chunks, and download only `B2` without ever needing to download the entirety of `B`.

### DIFF
--- a/build/bazel/remote/execution/v2/remote_execution.proto
+++ b/build/bazel/remote/execution/v2/remote_execution.proto
@@ -440,19 +440,19 @@ service ContentAddressableStorage {
     option (google.api.http) = { get: "/v2/{instance_name=**}/blobs/{root_digest.hash}/{root_digest.size_bytes}:getTree" };
   }
 
-  // Split a blob into chunks.
+  // SplitBlob retrieves information about how a blob is split into chunks.
   //
-  // This call splits a blob into chunks, stores the chunks in the CAS, and
-  // returns a list of the chunk digests. Using this list, a client can check
-  // which chunks are locally available and just fetch the missing ones. The
-  // desired blob can be assembled by concatenating the fetched chunks in the
-  // order of the digests in the list.
+  // This call returns information about how a blob is split into chunks, and
+  // returns a list of the chunk digests. Using the returned list of chunk digests,
+  // a client can check which chunks are locally available and only fetch the
+  // missing ones. The desired blob can be assembled by concatenating the fetched
+  // chunks in the order of the digests in the list. The chunks SHOULD all be
+  // available in the CAS.
   //
-  // This rpc can be used to reduce the required data to download a large blob
-  // from CAS if chunks from earlier downloads of a different version of this
-  // blob are locally available. For this procedure to work properly, blobs
-  // SHOULD be split in a content-defined way, rather than with fixed-sized
-  // chunking.
+  // This API can be used to reduce the required data to download a large blob
+  // from CAS if some chunks from similar blobs are locally available. For this
+  // procedure to work properly, blobs SHOULD be split in a content-defined way,
+  // rather than with fixed-sized chunking.
   //
   // If a split request is answered successfully, a client can expect the
   // following guarantees from the server:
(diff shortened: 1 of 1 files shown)


build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -440,13 +440,21 @@ service ContentAddressableStorage {
...
+  // server MAY compute a new one by performing chunking on-demand if the blob
+  // exists in the CAS, though this should be rare as manifests are typically
+  // created when blobs are written. The original blob digest does not need to
+  // be present in the CAS for this call to succeed if a manifest already exists.
+  //
+  // Using the returned list of chunk digests, a client can check which chunks
+  // are locally available and just fetch the missing ones. The desired blob can
+  // be assembled by concatenating the fetched chunks in the order of the
+  // digests in the list. ZSTD-compressed chunks can be concatenated without
+  // decompression.
```
sluongng: > ZSTD-compressed chunks can be concatenated without decompression.

I don't think this is correct.

There are a few ways compression can be applied here:

a. Compress the original large blob before chunking do not compress chunks
b. Compress the original large blob, chunk, then compress the chunks additionally
c. Chunk the uncompressed large blob, then compress the chunks individually

From the previous discussions, I think we do not care about (a) and (b) and will mostly be doing (c). Theoretically, (c) should give us a better chunk "hit rate" when using a consistent chunking algorithm (i.e. FastCDC) across different versions of the same large blob, as the compression algorithm can shuffle contents around.

Under (c), each chunk will be compressed individually. To get back the original large blob, one would need to decompress each chunk before concatenating them based on the order in the response.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -440,13 +440,21 @@ service ContentAddressableStorage {
...
   }
 
-  // Split a blob into chunks.
-  //
-  // This call splits a blob into chunks, stores the chunks in the CAS, and
-  // returns a list of the chunk digests. Using this list, a client can check
-  // which chunks are locally available and just fetch the missing ones. The
-  // desired blob can be assembled by concatenating the fetched chunks in the
-  // order of the digests in the list.
+  // Split retrieves information about how a blob is split into chunks.
```
sluongng: Compared to the original doc, this new version missed the part where the chunks MUST/SHOULD be stored in CAS. It's something obvious to an experienced maintainer, but should be documented to help newer implementations onboard the spec.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -491,26 +492,36 @@ service ContentAddressableStorage {
...
   // This is the complementary operation to the
   // [ContentAddressableStorage.SplitBlob][build.bazel.remote.execution.v2.ContentAddressableStorage.SplitBlob]
   // function to handle the chunked upload of large blobs to save upload
   // traffic.
   //
+  // When uploading a large blob using chunked upload, clients MUST first upload
+  // all chunks to the CAS, then call this RPC to store how those chunks compose
+  // the original blob. The chunks referenced in the manifest SHOULD be available
+  // in the CAS before calling this RPC. The original blob does not need to be
+  // available: the correctness of the manifest can be validated from manifest
```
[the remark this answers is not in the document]
tyler-french: Ah I missed this part, agree: cleaned up any implementation-specific bits. Just recommend that they must write the chunks first, and then Splice assumes that the CAS _should_ have access to all the chunks.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -2004,6 +2013,8 @@ message SpliceBlobRequest {
   //  2. It allows servers with different storage backends to dispatch the
   //     request to the correct storage backend based on the size and/or the
   //     hash of the blob.
+  //  3. It allows the server to check if existing chunking information is 
+  //     available for the blob.
```
sluongng: This is essentially a rewording of point (1), right?

It's worth keeping in mind that this field is still optional and not required, which makes it a SHOULD, not a MUST. If the client is CPU-constrained and prefers to skip hashing the giant file, it should be able to send all the chunks to the server and trust the digest the server returns.

If you are interested in implementing a server that enforces that this field is mandatory, we can try creating a new capability field, telling clients that they MUST set this field or expect an error otherwise.

tyler-french: See https://github.com/bazelbuild/remote-apis/pull/356. I think starting stricter is better, and this should be required. Its still an unreleased API

tyler-french:
@tjgq Do you have time to take another look here also? Thanks!

PR #356, opened 2025-12-18
Make SpliceBlobRequest.blob_digest mandatory

tyler-french:
For reasons listed in the API, this field should be required


fmeum (approved):

tyler-french:
> > 7\. If chunking information already exists for the blob, it allows
> >     the server to keep or replace with new chunking information.
> 
> how is this OR decided?

@sluongng I think the server should be at liberty to decide whether to keep the previous entry, or replace it. There is a risk of a faulty client poisoning or maliciously populating these.

PR #357, opened 2026-01-07
feat(SplitBlob/SpliceBlob): add chunking algorithm

tyler-french:
For CDC (Content-Defined Chunking), having the client and server agree upon a chunking algorithm unlocks a new level of possible improvements, where we can have distributed, deterministic, and reproducible chunking. 

This PR adds a chunking algorithm negotiation to GetCapabilities. Most notably, it:
- Defaults to no algorithm (server is not expected to be able to chunk, but can verify/store chunking info if it has Split/Splice capabilities)
- Adds FastCDC 2020 and RepMaxCDC as a supported algorithm with suggested defaults. 

If a chunking algorithm is announced by the server, the client will expect Split responses to be chunked using one of these algorithms. The client should also try to chunk using the same algorithm so that the server will accept new chunking information via Splice, and so many clients can de-duplicate storage with shared chunks.

## Why FastCDC 2020?

FastCDC 2020 is fast (~5GB/s on AMD Ryzen 9950), and is backed by a clear spec described in the paper: [IEEE paper](https://ieeexplore.ieee.org/document/9055082) [pdf](https://ranger.uta.edu/~jiang/publication/Journals/2020/2020-IEEE-TPDS(Wen%20Xia).pdf). It is popular, with https://github.com/nlfiedler/fastcdc-rs mirroring the paper's implementation.

This algorithm is:
- simple to implement: https://github.com/bazelbuild/bazel/pull/28438
- deterministic and reproducible
- fast and well-known

## Why RepMaxCDC?

RepMaxCDC is another fast chunking algorithm that has many benefits, and is used in https://github.com/buildbarn/bonanza. This algorithm is also included to provide users with multiple options for configuration of the chunking.

### Why the FastCDC 2020 threshold = Max Blob Size

We only chunk blobs > the threshold size. Having a threshold be >= the max chunk size is an important design decision for the first iterations of implementation here. Mainly, there's a cyclic re-chunking possibility that could happen if it's not upheld. For example, if the chunking threshold is only 1MB, and a chunker (avg size 512k and max 2MB) produces a 1.5MB chunk, the server is going to try to re-chunk this again. It also means the CAS doesn't know if a chunk is a full blob, or is a chunk, or is a file that was chunked into a single chunk. For FastCDC2020, the threshold will be set to the max chunk size, or 4*avg_size. This also has the benefit of **guaranteeing that we'll always get >1 chunk**, which simplifies implementations.

### Why the default chunk size?

512kB hits the sweet spot with Bazel artifacts for medium-side repos like https://github.com/buildbuddy-io/buildbuddy with some overlapping GoLink objects.

Using my `--disk_cache` from the past 1-2 months of development, I ran some benchmarking using different sizes, and got the following results:

| Avg Size | % Chunked | % Reused | % Unique | Dedup % | Saved |
|----------|-----------:|----------:|----------:|---------:|-------:|
| 16 KB | 18.1% | 48.0% | 52.0% | 50.0% | 141.91 GB |
| 32 KB | 15.8% | 46.0% | 54.0% | 48.2% | 136.90 GB |
| 64 KB | 13.2% | 42.3% | 57.7% | 45.8% | 130.18 GB |
| 128 KB | 7.5% | 38.4% | 61.6% | 43.1% | 122.24 GB |
| 256 KB | 6.1% | 32.5% | 67.5% | 39.8% | 112.96 GB |
| **512 KB** | **4.2%** | **24.0%** | **76.0%** | **35.8%** | **101.52 GB** |
| 1 MB | 1.9% | 16.4% | 83.6% | 31.0% | 88.00 GB |
| 2 MB | 1.6% | 8.8% | 91.2% | 25.9% | 73.47 GB |

De-duplication is strong at 512kiB (at 35%), and this only affects 4% of files.

We could drop to 64k, but we'd only get ~10% more de-duplication savings and still need to chunk 3x the number of files.

The sizes selected for RepMaxCDC are selected to similarly produce an average chunk size around 512kiB

Thank you to @sluongng for much of the initial version of this PR


build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -2284,6 +2331,98 @@ message CacheCapabilities {
...
+  // Blobs with size less than or equal to this threshold SHOULD be uploaded
+  // as single blobs without chunking. Blobs with size greater than this
+  // threshold SHOULD be chunked.
+  //
+  // This threshold SHOULD be greater than or equal to the maximum chunk size
+  // produced by the chunking function to avoid re-chunking blobs that would
+  // result in a single chunk.
+  //
+  // If unset, clients SHOULD use 2097152 (2 MiB) as the default.
+  uint64 chunking_threshold_bytes = 2;
```
sluongng: Nit: Move these 2 fields to the top of ChunkingConfig message so FastCDCParams is closer to the field. In the future, if new chunking function is added, then the message will also be closer to the field

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -2153,13 +2162,51 @@ message DigestFunction {
...
+    //
+    // Supported implementations:
+    //   - Rust: https://docs.rs/fastcdc/3.2.1/fastcdc/v2020/index.html
+    //   - Go: https://github.com/buildbuddy-io/fastcdc2020
+    //
+    // Servers which support this chunking function SHOULD advertise the following
+    // configuration parameters through the CacheCapabilities message:
+    //   - avg_chunk_size_bytes
+    //   - normalization_level
+    //   - seed
```
sluongng: we could make it s/SHOULD/MUST/ here to force servers that support fastcdc to always advertise relevant chunking config.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -2284,6 +2331,98 @@ message CacheCapabilities {
...
+  // The minimum and maximum chunk sizes are derived from the average:
+  //   - min_chunk_size = avg_chunk_size_bytes / 4
+  //   - max_chunk_size = avg_chunk_size_bytes * 4
+  //
+  // If any of the advertised parameters are not within the expected range,
+  // the client SHOULD ignore FastCDC chunking function support.
+  message FastCDCParams {
+    // The average (expected) chunk size for the FastCDC chunking algorithm.
+    // The value MUST be between 1 KiB and 1 MiB.
+    // If unset, clients SHOULD use 524288 (512 KiB) as the default.
```
sluongng: There are average_max and average_min values in the Rust implementation. Might be worth mentioning them here?

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -2284,6 +2331,98 @@ message CacheCapabilities {
...
+  // is a reference implementation that follows the paper specification.
+  //
+  // Key algorithm components from the paper:
+  //
+  // GEAR table: 256 64-bit integers for the rolling hash, computed as:
+  //   GEAR[i] = high_64_bits(MD5(byte(i))) for i in 0..255
+  //
+  // MASKS table: Bit patterns for chunk boundary detection, derived from
+  // the C reference implementation. The mask selection based on average
+  // chunk size and normalization level should match the paper.
```
sluongng: A bit more information about this would be nice. That would save folks from having to chew through the paper just to figure out how the mask works.

In actual rust implementation, we know that this is not just a simple table:

```rust
pub struct FastCDC<'a> {
    ...
    mask_s: u64,
    mask_l: u64,
    mask_s_ls: u64,
    mask_l_ls: u64,
    gear: Box<[u64; 256]>,
    gear_ls: Box<[u64; 256]>,
}

    ///
    /// Create a new [`FastCDC`] with the given normalization level and seed to
    /// be XOR'd with the values in the gear tables.
    ///
    pub fn with_level_and_seed(
        source: &'a [u8],
        min_size: u32,
        avg_size: u32,
        max_size: u32,
        level: Normalization,
        seed: u64,
    ) -> Self {
        ...
        let bits = logarithm2(avg_size);
        let normalization = level.bits();
        let mask_s = MASKS[(bits + normalization) as usize];
        let mask_l = MASKS[(bits - normalization) as usize];
        let (gear, gear_ls) = get_gear_with_seed(seed);
        Self {
            ...
            mask_s,
            mask_l,
            mask_s_ls: mask_s << 1,
            mask_l_ls: mask_l << 1,
            gear,
            gear_ls,
        }
    }
```

but a set of configs picked by deriving from the avg_size, level, and seed values based on some heuristic.

If we don't want to describe the heuristic, might as well just link people directly to the referenced implementation.

sluongng (commented):
Im a bit occupied this week, so I plan to give this a better read next week. Got some small nits but lgtm overall.

I think it would be nice if we could provide a test vector for this, similar to https://github.com/bazelbuild/remote-apis/blob/main/build/bazel/remote/execution/v2/sha256tree_test_vectors.txt, which was added for sha256tree. This way, folks can test their implementation against the test vector to verify that the generated chunks are identical across implementations.

build/bazel/remote/execution/v2/remote_execution.proto
tyler-french: This was why I added defaults, so if the server uses the defaults, they can leave it unset, we can assume the defaults. Do you think it's necessary to require even with the defaults?

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -2284,6 +2331,98 @@ message CacheCapabilities {
...
+    // The value MUST be between 0 and 3.
+    //
+    // Higher normalization levels produce chunks closer to the average size
+    // but may reduce deduplication effectiveness:
+    //   - Level 0: No normalization, widest size distribution
+    //   - Level 1: Fewer chunks outside desired range
+    //   - Level 2: Most chunks match desired size (recommended)
+    //   - Level 3: Nearly all chunks are the desired size
+    //
+    // If unset, clients SHOULD use 2 as the default.
```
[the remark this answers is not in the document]
sluongng: Personally I think we should provide concrete benchmark setup so that folks can verify against their own data set to see what's best.

If we think 2 is definitively best for RBE use case, let's just document that we should default use 2 and remove the normalization level config knob here. That would help us simplify the setup a bit more. Whoever is interested in using a different level could add it back later.

WDYT?

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -2284,6 +2331,98 @@ message CacheCapabilities {
...
+  //
+  // If any of the advertised parameters are not within the expected range,
+  // the client SHOULD ignore FastCDC chunking function support.
+  message FastCDCParams {
+    // The average (expected) chunk size for the FastCDC chunking algorithm.
+    // The value MUST be between 1 KiB and 1 MiB.
+    // If unset, clients SHOULD use 524288 (512 KiB) as the default.
+    //
+    // The minimum chunk size will be avg_chunk_size_bytes / 4.
+    // The maximum chunk size will be avg_chunk_size_bytes * 4.
```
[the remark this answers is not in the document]
tyler-french: Maybe MUST? then we can relax in the future.

tyler-french:
> Yeah, that would be great!

When you get the chance, please let me know how the wording of the MaxCDC params look. I rounded the recommended params from my benchmarking to the nearest power of 2.

tyler-french:
> > Yeah, that would be great!
> 
> When you get the chance, please let me know how the wording of the MaxCDC params look. I rounded the recommended params from my benchmarking to the nearest power of 2.

@EdSchouten Oops, I totally missed the comments on https://github.com/tyler-french/remote-apis/pull/1. I'll go ahead on addressing those here, thanks for the feedback!

EdSchouten (approved):

fmeum (approved):

EdSchouten (approved):

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -2284,6 +2331,112 @@ message CacheCapabilities {
...
+  }
+
+  // The size threshold for chunking blobs.
+  // Blobs with size less than or equal to this threshold SHOULD be uploaded
+  // as single blobs without chunking. Blobs with size greater than this
+  // threshold SHOULD be chunked.
+  //
+  // This threshold SHOULD be greater than or equal to the maximum chunk size
+  // produced by the chunking function to avoid re-chunking blobs that would
+  // result in a single chunk.
```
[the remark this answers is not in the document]
tyler-french: Maybe I can make this more of a recommendation. The intention here is that with multiple servers, and some clients that support chunking and some that don't, it's difficult to distinguish the difference between a blob that was chunked and results in a single chunk (a singleton), and a blob that hasn't yet been chunked.

To begin, keeping this threshold > the max blob size makes it absolutely clear that a blob hasn't been chunked, and needs to be. 

More info here: https://github.com/bazelbuild/remote-apis/pull/358, but I think we should recommend this for now: which is why it's a SHOULD and not MUST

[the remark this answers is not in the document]
tyler-french: The threshold is also separate from the algorithm. Anything below the threshold can bypass any chunking-related logic, and the client can avoid any split/splice interactions. How this compares with the max blob size is more of a suggestion.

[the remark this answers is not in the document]
EdSchouten: Gotcha! Thanks for clarifying.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1971,6 +1971,12 @@ message SplitBlobRequest {
   // length of the blob digest hashes and the digest functions announced
   // in the server's capabilities.
   DigestFunction.Value digest_function = 3;
+
+  // The chunking function that the client prefers to use.
+  //
+  // The server MAY use a different chunking function. The client MUST check
+  // the chunking function used in the response.
+  ChunkingFunction.Value chunking_function = 4;
```
sluongng: Hmm why MUST? The client can simply not care and just concat the chunks, no?

If there are specific actions the client should take when the response's chunking function differs from the request, I would like them to be listed out here.

For example, we can say that the server MAY use the chunking function as part of the resource name (cache key, cache path,etc..) to store the large blob manifest. But even then, I'm not sure if it's reasonable for us to expect the client to automatically reconfigure itself according to the new chunking function. At best, all I'm seeing is the client outputting a message to warn about poor CAS hit rates?

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -1971,6 +1971,12 @@ message SplitBlobRequest {
   // length of the blob digest hashes and the digest functions announced
   // in the server's capabilities.
   DigestFunction.Value digest_function = 3;
+
+  // The chunking function that the client prefers to use.
```
sluongng, line 1990: MUST be one of the advertised in `CacheCapabilities.chunking_configuration`.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -2153,13 +2165,48 @@ message DigestFunction {
...
+
+// The chunking function is used to split a blob into chunks.
+//
+// The server advertises support for a chunking function by setting the
+// corresponding params field in
+// [ChunkingConfiguration][build.bazel.remote.execution.v2.ChunkingConfiguration].
+// For example, if fast_cdc_2020_params is set, the server supports FAST_CDC_2020.
+//
+// For optimal deduplication, clients SHOULD use an advertised chunking function.
+// Servers MUST always accept UNKNOWN.
```
sluongng: We can reword this to be a bit more descriptive, especially around how the server handles Split requests and Splice requests when the chunking algo was unknown.

Does it pick a random chunking algorithm to do the split, then write the manifest into an "chunking_function.unknown" namespace? 

In another PR, we tried to characterize Splice as not concatenating the chunks live, but only to look up the manifest. So what would happen if a Splice request came in with an unknown chunking function? Should the server look up the manifest in all supported chunking function namespaces? Or is this too deep into the implementation details for the spec?

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -2153,13 +2165,48 @@ message DigestFunction {
...
+    //
+    // Cutting points are never placed within the first or last min_chunk_size_bytes
+    // of a file. This differs from FastCDC, where the final chunk may be as small
+    // as a single byte. This property improves semantic security when encrypting
+    // objects, reduces data processing overhead, and ensures more evenly sized chunks.
+    //
+    // This approach results in a more uniform chunk size distribution compared
+    // to FastCDC, with the spread in chunk size being significantly smaller.
+    // The algorithm is expected to converge more quickly after differing parts
+    // between similar files have finished processing.
```
sluongng: I would consider this paragraph to be the implementation "details" that folks could go to https://github.com/buildbarn/go-cdc to read more about?

sluongng (commented):
I think this is clean and nice. A few nits to make the spec stricter.

Given the context of other PR, I do get why you would want to add chunking_function into the SpliceRequest and SplitResponse. 

If the server were doing all the Chunking and Concatting live, I don't think those fields are needed. But if they are simply looking up existing "big blob manifests", then those values might be useful for the client/server to store and look up the manifests? Is that the intention?

sluongng:
@EdSchouten @tjgq, since the current 2 algorithms all have their respective open source implementation available. Do we still want a test vector in this repo, similar to the sha256tree test vector, to help others validate their implementation?

I know we asked for test vectors in a meeting many moons ago when we discussed the chunking algorithms. Not entirely sure if they are still necessary.

I suspect if the 2 maintainers just provide released binaries for each algorithm in their repo, perhaps with the config knobs as flags, then we can just refer to those CLIs as test vectors?

tyler-french:
> @EdSchouten @tjgq, since the current 2 algorithms all have their respective open source implementation available. Do we still want a test vector in this repo, similar to the sha256tree test vector, to help others validate their implementation?
> 
> I know we asked for test vectors in a meeting many moons ago when we discussed the chunking algorithms. Not entirely sure if they are still necessary.
> 
> I suspect if the 2 maintainers just provide released binaries for each algorithm in their repo, perhaps with the config knobs as flags, then we can just refer to those CLIs as test vectors?

I have been using this as a test vector: https://github.com/nlfiedler/fastcdc-rs/blob/master/test/fixtures/SekienAkashita.jpg
The issue is I don't know if we want to check in an image to this repo, https://github.com/nlfiedler/fastcdc-rs/blob/master/src/v2020/mod.rs#L902-L925

I added the same test here: https://github.com/buildbuddy-io/fastcdc2020/blob/main/fastcdc/fastcdc_test.go#L11-L36

The resulting chunks are:
```
		{17968276318003433923, 21325},
		{8197189939299398838, 17140},
		{13019990849178155730, 28084},
		{4509236223063678303, 18217},
		{2504464741100432583, 24700},
```
Should I add this?

sluongng:
Perhaps we can use the perma link https://github.com/nlfiedler/fastcdc-rs/blob/49c3d0b8043a7c1c2d9aca75e868d3791ffedcf3/test/fixtures/SekienAkashita.jpg with the SHA256 fingerprint to identify the blob?

The file is 107KB, so it's not too bad.

EdSchouten:
Tyler and I had a private chat via Slack, but I thought I should share this here as well.

I got nerd sniped by the discussion we had about the addition of `chunking_threshold_bytes` and PR #358. Namely, algorithms like FastCDC and MaxCDC don't really allow you to easily distinguish between chunked and unchunked blobs. For example, you can't just look at a blob's size to know whether repeated application of chunking yields anything different.

This got me thinking: would it be possible to design a somewhat decent chunking algorithm that always yields chunks of size $[n, 2n)$? Because if you had an algorithm like that, you know that anything smaller than `2n` cannot be chunked further, while anything at least as large as `2n` can definitely be chunked. So it turns out we can. Namely, by applying MaxCDC repeatedly!

I just added such an algorithm to [the go-cdc repository](https://github.com/buildbarn/go-cdc/tree/main) under the name RepMaxCDC. Benchmark results look very promising. In fact, I wasn't able to find any measurable difference between MaxCDC and RepMaxCDC, even though the latter has far tighter bounds on object size. @tyler-french (and others), I would really like to invite you to give this a try.

For this PR my recommendation would be to:

- Change references of MaxCDC to RepMaxCDC.
- Rename RepMaxCDC's `max_chunk_size_bytes` to `horizon_size_bytes`.
- Move `chunking_threshold_bytes` into `FastCdc2020Params`, as RepMaxCDC does not depend on it.
- Maybe eliminate `ChunkingConfiguration` and inline its fields into `CacheCapabilities`. The reason being that its presence/absence does not imply anything.

tyler-french:
> Perhaps we can use the perma link [nlfiedler/fastcdc-rs@`49c3d0b`/test/fixtures/SekienAkashita.jpg](https://github.com/nlfiedler/fastcdc-rs/blob/49c3d0b8043a7c1c2d9aca75e868d3791ffedcf3/test/fixtures/SekienAkashita.jpg) with the SHA256 fingerprint to identify the blob?
> 
> The file is 107KB, so it's not too bad.

Added test vectors, and referenced them here: https://github.com/buildbuddy-io/fastcdc2020/pull/4, which can be used to verify

build/bazel/remote/execution/v2/remote_execution.proto
tyler-french: I think this could be an implementation detail of the server, similar to the one above. Just because a server doesn't have a chunker that uses an algo, doesn't mean it might not have information on that stored, and the server can still respond with a different algorithm.

I just simplified this to be:
```
  // The chunking function that the client prefers to use.
  //
  // The server MAY use a different chunking function.
```
The client can concat them together for the Read, and then decide to keep/discard the chunks for later based on the function

build/bazel/remote/execution/v2/remote_execution.proto
tyler-french: See response to other one: I think the only guidelines we need are:
```
  // The chunking function that the client prefers to use.
  //
  // The server MAY use a different chunking function.
```
How the server and client handle these things are implementation details. I would probably say clients that use fastcdc2020 should only keep chunks from fastcdc2020 in their local disk cache, or keep chunked mappings locally for these, but it's not a requirement.

build/bazel/remote/execution/v2/remote_execution.proto
tyler-french: I think the description is mostly sufficient, and this is also server implementation details. I updated it to be a bit more clear that for `Split` the server can pick, if Unknown is passed, and for `Splice` the server is only responsible for validating concatenation.

tyler-french:
@EdSchouten @fmeum @sluongng made some changes, please take another look when you get the chance!

Summary is:
- Removed threshold: rep max doesn't need one, fastcdc2020 the threshold will be == max chunk size
- Added test vectors for fastcdc2020

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -2299,6 +2338,101 @@ message CacheCapabilities {
...
+    // The minimum chunk size for the RepMaxCDC chunking algorithm.
+    // The value MUST be at least 64 bytes (the Gear hash window size).
+    // All chunks will be in the range [min_chunk_size_bytes, 2*min_chunk_size_bytes).
+    // The recommended value is 262144 (256 KiB).
+    uint64 min_chunk_size_bytes = 1;
+
+    // The lookahead window for finding optimal cutting points.
+    // Larger values improve deduplication quality with diminishing returns.
+    // Setting to 0 produces uniform chunks of min_chunk_size_bytes.
+    // The recommended value is 262144 (256 KiB).
```
EdSchouten: Measurements seem to indicate that the algorithm works best if this is `8 * min_chunk_size_bytes`.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -2299,6 +2338,101 @@ message CacheCapabilities {
   // [ContentAddressableStorage.SpliceBlob][build.bazel.remote.execution.v2.ContentAddressableStorage.SpliceBlob]
   // operation.
   bool splice_blob_support = 10;
+
+  // The chunking configuration of the server.
+  ChunkingConfiguration chunking_configuration = 11;
```
EdSchouten: Do we need this to be a separate message? Is there anything special that a client can do differently if this field is set, but it does not contain any `*_params` for algorithms supported by the client? There's also no reuse of this message in other places. I would therefore recommend not doing this, as it only introduces ambiguity.

tyler-french: I would personally prefer to keep it, so if we need to add any algorithm agnostic chunking config, we don't have to make backward incompatible changes or add a new structure. But if we would prefer to change it, I'm happy to!

tyler-french: Actually, best practice is to avoid speculative structure. I'll fix this.

EdSchouten (approved):

fmeum (approved):

sluongng (approved):

tyler-french:
@tjgq Please take another look when you get the chance! Once this gets merged, we should cut a new release.

EdSchouten:
Even though I already approved this PR a couple of days ago, just wanted to say that this is good to land w.r.t. RepMaxCDC. The reference implementation available at https://github.com/buildbarn/go-cdc works reliably now, and performance is on par with FastCDC.

PR #377, opened 2026-05-18
Support streaming split/splice APIs for large blobs

tyler-french:
**TL;DR: Unary split/splice forces large chunk manifests into a single size-limited message and makes the receiver wait for the complete manifest before processing it; streaming keeps messages bounded and enables incremental processing, improving scalability and performance for large blobs.**

This PR adds `GetChunkMapping` and `RegisterChunkMapping`, streaming replacements for split and splice. This lets clients send the ordered chunk list across multiple messages instead of fitting it all into one unary request or response.

The names make the direction of each operation explicit and describe what the RPCs exchange: an ordered blob-to-chunk digest mapping. They do not transfer chunk data, so names such as `GetChunks` or `PutChunks` would be misleading.

Streaming also gives servers a chance to process the mapping as it arrives. For splice, this can avoid doing all of the work in one final RPC after every chunk has already been uploaded. Servers can still choose to wait until the stream is complete and process it like the unary path.

This solves the issues described in #376.


EdSchouten:
I got added as a reviewer, so here's my take: just like how chunking has been added to the protocol in its current form, I'm not a fan of this. Not having streaming, but requiring a recursive construction of very large files using Prolly trees is in my opinion the way to go.

That said, if people actually on the working group think that this is the way to go, go for it! :-)

tyler-french:
> I got added as a reviewer, so here's my take: just like how chunking has been added to the protocol in its current form, I'm not a fan of this. Not having streaming, but requiring a recursive construction of very large files using Prolly trees is in my opinion the way to go.
> 
> That said, if people actually on the working group think that this is the way to go, go for it! :-)

@EdSchouten Thanks, that makes sense. I agree that a recursive/Merkle-style representation such as Prolly trees is probably the better long-term model for representing extremely large blobs, since it bounds individual node sizes and avoids one giant flat chunk list.

My goal with this PR is narrower: provide a transport-level alternative to https://github.com/bazelbuild/remote-apis/pull/376 without making clients materialize or dereference a chunk-list object in CAS. Streaming lets clients send the existing ordered chunk mapping in bounded batches, and it lets servers begin verification as chunks become available instead of waiting until every chunk has been uploaded and then receiving the whole mapping in one unary `SpliceBlob` request.

I also see this as compatible with a future recursive representation. A server could still store the received mapping internally as a Prolly tree or some other recursive structure. The protocol question here is whether clients need to understand and construct that representation, or whether they can just stream the chunk references they already have.

Longer term, I would prefer for the streaming APIs to become the primary split/splice shape for clients that support them, with the existing unary APIs remaining as the compatibility/simple-case path. For cases where the chunk list fits in one request, sending it as a single stream message should be performance-equivalent to a unary gRPC request in practice, while still leaving the same API able to scale when the list grows. A recursive representation also fits into that model: the stream does not require the server to store a flat list internally, it just gives the server live visibility into the composition progress. The server can choose to persist that mapping as a Prolly tree, a flat manifest, or another internal structure, while still getting the benefits of streaming transport: bounded message sizes, earlier validation, and the ability to pipeline verification while the client is still uploading or discovering chunks, since chunking is inherently sequential.

Some recursive composition can be approximated with the current API today by splicing subranges and then splicing those intermediate blobs, but standardizing that as the preferred model seems like it would need more protocol work: tree node encoding, fanout rules, digest/validation semantics, and how `chunking_function` applies at each level. Streaming is a smaller additive step that addresses the immediate unary message-size and final-RPC verification problem while leaving room for that design.

tyler-french:
@EdSchouten One open question I have for a Prolly tree design is how it preserves the existing full-blob digest semantics. In REAPI, the digest in an `ActionResult` is the digest of the output file contents, and CAS lookups are keyed by that full content digest. So even if the server stores the blob as a recursive tree internally, both the client and server eventually need to know that the tree represents exactly the bytes for the full blob digest. 

If the Prolly root digest is a digest of tree-node bytes, it is not a substitute for the file's content digest. If the key remains the full blob digest, the server still has to verify the leaves in order by hashing the reconstructed content, or trust a mapping it has already validated. And if a server only has chunks/tree nodes under one chunking strategy, it is not obvious how it can know that a different client-provided tree already represents the same blob without either having a stored full-digest-to-tree mapping or walking/rehashing the content.

moroten:
> The streaming approach keeps chunk-list transport in the RPC layer instead. That has a few advantages:
> 
> - It avoids using CAS as a protocol transport for RPC metadata.

That is true for this specific RPC, but many other parts of the protocol already work this way.

> - It does not require clients to add extra read/write logic for chunk-list objects.

I agree that it would probably be easier to adapt most servers and clients.

> - It lets servers receive the ordered chunk mapping incrementally instead of only at the final splice call.

The speed improvement I can see is that splicing can finish faster after the last chunk has been uploaded. The work still has to be done, so it is a matter of cutting latency, instead of loading GB of data at the end, it can be loaded while the client is uploading the chunks. @tyler-french Do you see the latency as a problem in your current clusters?

My biggest concern is that the servers may have to hold an infinite number of connections open, waiting for more chunks to arrive.
* What timeout should be allowed? For Bazel, we will still have the be quicker than the global [--remote_timeout](https://bazel.build/reference/command-line-reference#common_options-flag--remote_timeout).
* If a splice call times out, the server has to throw away all work. With a unary call with a chunk-list the server can process the whole splicing and store the result, so timeout should not be any problem (unless server internal networking is slow).
* Minor note: For clients to handle timeouts, they need to keep the full list of chunks until splicing is done. There will not be any benefit in lower memory consumption on client side.
* With a unary call, the splicing can be deduplicated on server side.

To address my concerns, I might implement the streaming splicing by waiting for the last message and then do the splicing anyway, effectively converting it into the chunk-list version.

> - It keeps the old unary APIs simple and fully usable by older clients.

The chunk-list approach also keeps the old API fully usable by older clients.

> - It gives clients a capability-gated path for large blobs without changing the semantics of existing fields.

Either we have the feature of handling large blobs capability gated or we can put it as requirement on Split/Splice API from a certain version of the REv2 API.

tyler-french:
@moroten (and cc @meroton-benjamin) Thanks, these are good concerns. I am also happy to support https://github.com/bazelbuild/remote-apis/pull/376 if that is where we land. I needed to add streaming support on our side anyway, so this felt like a good time to see whether it could also solve the large chunk-list problem.

I think the main point is that streaming does not remove the splice work. It gives clients a bounded-message way to send the chunk list, and it lets servers start seeing the mapping earlier if they want to. A client does not have to keep the stream open while chunks are still uploading. It can upload the chunks first, then call `SpliceChunks` with the full chunk list in one or a few messages. That makes this usable as an alternative to the indirect chunk-list approach, without requiring the chunk list itself to be stored and fetched through CAS.

> The speed improvement I can see is that splicing can finish faster after the last chunk has been uploaded. The work still has to be done, so it is a matter of cutting latency, instead of loading GB of data at the end, it can be loaded while the client is uploading the chunks. @tyler-french Do you see the latency as a problem in your current clusters?

Yes. We have hit issues with blobs around 1 GB. Today the client uploads chunks with `BatchUpdateBlobs`, then calls `SpliceBlob` after all chunks are available. Since the server only sees the mapping at the end, all splice work is deferred into one final RPC, and we have seen that call time out.

I agree that the work still has to be done. The benefit is mostly message sizing and tail latency. With a fast chunker, I would expect an incremental `SpliceChunks` flow to take about as long as `Write` for the same blob.

> My biggest concern is that the servers may have to hold an infinite number of connections open, waiting for more chunks to arrive.

That risk is real, but it is similar to `Write`. Servers still need normal RPC controls, such as deadlines, idle timeouts, max concurrent streams, and per instance limits.

Also, clients that want to avoid long idle splice streams can wait until all chunks are uploaded, then send the mapping in one or a few `SpliceChunks` messages. That gives a similar call shape to PR #376, but keeps the chunk-list transport in the RPC layer.

> What timeout should be allowed? For Bazel, we will still have the be quicker than the global --remote_timeout.

I think this should follow the normal RPC deadline model. For Bazel, the call still needs to complete inside `--remote_timeout`. Probably something similar to whatever the Write call was for the blob.

> If a splice call times out, the server has to throw away all work. With a unary call with a chunk-list the server can process the whole splicing and store the result, so timeout should not be any problem (unless server internal networking is slow).

This is the main tradeoff. If the stream fails before the server commits the splice, the client has to retry. After the chunks are uploaded, that retry is much cheaper than retrying the original upload.

There is also a useful fallback. If the chunk list fits in one message, the client can retry with unary `SpliceBlob`. If it does not, it can retry `SpliceChunks` with smaller batches or only a few messages. So the client has a bit more control over getting things to succeed

> Minor note: For clients to handle timeouts, they need to keep the full list of chunks until splicing is done. There will not be any benefit in lower memory consumption on client side.

Agreed. Keeping a []digests is pretty cheap. You could even just keep []offsets to save even more memory.

> With a unary call, the splicing can be deduplicated on server side.

I got this feedback, and I think it makes sense the request requires `blob_digest`. It should deduplicate well now by routing by this.

> To address my concerns, I might implement the streaming splicing by waiting for the last message and then do the splicing anyway, effectively converting it into the chunk-list version.

That seems like a valid implementation. It would still avoid the unary message size limit and let clients use one API shape for larger blobs.

> The chunk-list approach also keeps the old API fully usable by older clients.

Agreed. I did not mean that the chunk-list approach breaks older clients.

The difference I see is that PR #376 adds another representation to the existing unary API shape. Newer clients then need logic to write, read, and interpret chunk-list objects. With streaming, the old unary APIs stay as the simple path, and the large-blob path is a separate capability-gated RPC. A client that is not aware of indirection may fail if the SplitBlob response sets it.

moroten:
@tyler-french I would be happy with the streaming approach as well. Both approaches have their pros and cons, but the main problem is that the whole API refers to blobs as whole files, not a Prolly tree.

Maybe set the decision next monthly? I'll add some code comments as well.

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -2008,6 +2088,33 @@ message SplitBlobResponse {
...
+  // The server MUST use the same digest function as the one explicitly or
+  // implicitly (through hash length) specified in the split request.
+  repeated Digest chunk_digests = 1;
+
+  // If set, this is the final response in the stream. Servers MUST set this on
+  // exactly one response and MUST close the stream after sending that response.
+  // If the final response also contains `chunk_digests`, clients MUST include
+  // those digests in the assembled chunk list before treating the stream as
+  // complete.
+  bool finish_split = 2;
```
moroten: Why use a `finish_split` boolean instead of just closing the stream? (Also applies to splice.)

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -2008,6 +2088,33 @@ message SplitBlobResponse {
...
+
+  // The chunking function used to split the blob. The server MUST send the same
+  // value in all responses.
+  ChunkingFunction.Value chunking_function = 3;
```
moroten: Can we move this into a `header` or `first_message` structure that MUST be set only for the first message in the stream? (Also applies to splice.)

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -2346,6 +2500,22 @@ message CacheCapabilities {
...
+  // operation.
+  bool split_chunks_support = 13;
+
+  // Whether streamed blob splicing is supported for the particular
+  // server/instance. If yes, the server/instance implements the specified
+  // behavior for streamed blob splicing and a meaningful result can be expected
+  // from the
+  // [ContentAddressableStorage.SpliceChunks][build.bazel.remote.execution.v2.ContentAddressableStorage.SpliceChunks]
+  // operation.
+  bool splice_chunks_support = 14;
```
moroten: As the server implementation will be very similar to the split/splice blob API, would it make sense to just say that from version X, implementing split/splice blob API also requires the split/splice chunks implementation? This way, clients get less amount of if cases to support all permutations of capabilities.

build/bazel/remote/execution/v2/remote_execution.proto
armandomontanez: This exists on `ByteStream.Write()` (and NOT `Read()`), but I believe the sole reason for that is to support resumability, for which the directionality of the operation is relevant.

https://github.com/googleapis/googleapis/blob/29c6482ccb75b302c33665d8c7d7a38d3698be53/google/bytestream/bytestream.proto#L146-L149

https://github.com/googleapis/googleapis/blob/29c6482ccb75b302c33665d8c7d7a38d3698be53/google/bytestream/bytestream.proto#L66-L73

I think this is worth evaluating from the perspective of whether or not we'd like to fully support the same partial read/write semantics or not. If we do want that, though, there's more work to be done to finish this thought.

Edit: resumability is amusing, but complex enough that it's probably not worth thinking about right now.

tyler-french: @armandomontanez Yeah, I think I'll wait on making this resumable.

> instead of just closing the stream? 

I agree. I think that this is easier also - since in both directions, we know the size of the concatenated blob, so the client/server can just know if its finished if the sum of the sizes match.

build/bazel/remote/execution/v2/remote_execution.proto
tyler-french: I indicated this in the spec, per your recommendation. I don't see much value in adding additional structure to the message. Do any other fields apply to this? Currently `blob_digest` is required in all messages for routing. Assuming the connection isn't broken, since we don't do resuming, we probably should only put that in the first one too...

tyler-french: Updated

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -2329,13 +2469,17 @@ message CacheCapabilities {
   // yes, the server/instance implements the specified behavior for blob
   // splitting and a meaningful result can be expected from the
   // [ContentAddressableStorage.SplitBlob][build.bazel.remote.execution.v2.ContentAddressableStorage.SplitBlob]
+  // operation. Starting in RE API v2.13, this also includes the
+  // [ContentAddressableStorage.SplitChunks][build.bazel.remote.execution.v2.ContentAddressableStorage.SplitChunks]
```
moroten: Servers that stop supporting RE API v2.12 should be able to drop the old implementation.
```suggestion
  // operation for RE API v2.12 and from
  // [ContentAddressableStorage.SplitChunks][build.bazel.remote.execution.v2.ContentAddressableStorage.SplitChunks]
// operation for RE API v2.13 or higher.
```

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -2329,13 +2469,17 @@ message CacheCapabilities {
...
   // operation.
   bool split_blob_support = 9;
 
   // Whether blob splicing is supported for the particular server/instance. If
   // yes, the server/instance implements the specified behavior for blob
   // splicing and a meaningful result can be expected from the
   // [ContentAddressableStorage.SpliceBlob][build.bazel.remote.execution.v2.ContentAddressableStorage.SpliceBlob]
+  // operation. Starting in RE API v2.13, this also includes the
+  // [ContentAddressableStorage.SpliceChunks][build.bazel.remote.execution.v2.ContentAddressableStorage.SpliceChunks]
```
moroten: Servers that stop supporting RE API v2.12 should be able to drop the old implementation.
```suggestion
  // operation for RE API v2.12 and from
  // [ContentAddressableStorage.SpliceChunks][build.bazel.remote.execution.v2.ContentAddressableStorage.SpliceChunks]
  // operation for RE API v2.13 or higher.
```

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -2346,6 +2490,7 @@ message CacheCapabilities {
   // The parameters for the RepMaxCDC chunking algorithm.
   // If set, the server supports the RepMaxCDC chunking algorithm.
   RepMaxCdcParams rep_max_cdc_params = 12;
+
```
moroten: Remove this inserted line.
```suggestion
```

build/bazel/remote/execution/v2/remote_execution.proto
moroten, line 522:
```suggestion
  //   chunks.
  //
  // New in v2.12 and removed in v2.13, use [SplitChunks][build.bazel.remote.execution.v2.ContentAddressableStorage.SplitChunks] instead.
```

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -554,10 +592,49 @@ service ContentAddressableStorage {
   //   extend the lifetime of the chunks specified in the request, e.g. because
   //   it prefers a different chunking and extended those instead. Clients can
   //   call [SplitBlob][build.bazel.remote.execution.v2.ContentAddressableStorage.SplitBlob]
+  //   or [SplitChunks][build.bazel.remote.execution.v2.ContentAddressableStorage.SplitChunks]
   //   to check what chunk mapping the server is using.
```
moroten:
```suggestion
  //   to check what chunk mapping the server is using.
  //
  // New in v2.12 and removed in v2.13, use [SpliceChunks][build.bazel.remote.execution.v2.ContentAddressableStorage.SpliceChunks] instead.
```

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -2049,6 +2147,48 @@ message SpliceBlobRequest {
...
+  Digest blob_digest = 2;
+
+  // The ordered list of digests of the chunks which need to be concatenated to
+  // assemble the original blob. Chunk digests may be split across multiple
+  // stream requests. The original blob is assembled by concatenating chunks in
+  // the order of these digests across all requests in stream order.
+  //
+  // Clients SHOULD limit the number of digests in each request to remain below
+  // the maximum message size accepted by the client/server pair.
+  repeated Digest chunk_digests = 3;
```
moroten, line 2168: While we are here, what about swapping `Digest` for an adapted variant that is smaller (`bytes` instead of `string`) and can be used for binary search to find a specific offset.
```proto
message ChunkDigest {
{
  // The hash padded with leading zeroes up to the hash function length.
  bytes hash = 1;

  // The sum of the sizes of this blob and all previous blobs, in bytes.
  int64 accumulated_size_bytes = 2;
}
```
The same applies to `SplitChunksResponse`.

EdSchouten: I don't think that would really be of any use. You're already paying the price of deserializing the Protobuf message (which requires at least a linear amount of time). If you're doing that, there's virtually no cost in converting the data to some internal representation that supports binary searching.

moroten: True, variable encoding means decoding is needed before binary search.

tyler-french: Keeping as digest also creates good continuity with the other APIs in v2

sluongng:
Thanks to the BuildBarn folks for the thorough reviews.

From my understanding, this PR seems to be in a good state for a final round of readings and discussion in tomorrow's meeting. If you have any questions / comments / amendments / objections, make sure to add them async here (preferred) and/or during the WG meeting tomorrow.

cc other active maintainers: @tjgq @ulfjack @werkt

tyler-french:
@tjgq @sluongng @fmeum let me know if you have time to take a pass. Thanks!

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -2049,6 +2151,48 @@ message SpliceBlobRequest {
...
```
armandomontanez, line 2168: It would be good to specify the expectations around an empty `chunk_digests` list now that clients/servers may send multiple messages.

tyler-french:
@tjgq @sluongng @fmeum friendly bump on this one. Thanks!

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -2170,6 +2177,10 @@ message SpliceChunksRequest {
   //
   // Clients SHOULD limit the number of digests in each request to remain below
   // the maximum message size accepted by the client/server pair.
+  //
+  // An empty list is allowed in any request. It contributes no chunks to the
+  // assembled chunk list; servers MUST validate the final concatenated chunk
+  // sequence against `blob_digest`.
```
moroten: If `ALREADY_EXISTS` is returned, servers should not need to concatenate the chunks at all.

build/bazel/remote/execution/v2/remote_execution.proto
moroten: Is "extending the lifetime of the chunks" necessary? I think that should be up to the server to decide.

tjgq (approved):
LGTM, but let's make sure feedback from other reviewers has been addressed (please mark conversations resolved as appropriate).

tyler-french:
> LGTM, but let's make sure feedback from other reviewers has been addressed (please mark conversations resolved as appropriate).

Thanks @tjgq! I addressed all previous feedback. @sluongng @fmeum @vanja-p do you all have time to take a look?

build/bazel/remote/execution/v2/remote_execution.proto
moroten: Remove `SplitBlob` in v2.13? (Cannot reopen my previous comment.)
```suggestion
  // New in v2.12 and removed in v2.13, use [SplitChunks][build.bazel.remote.execution.v2.ContentAddressableStorage.SplitChunks] instead.
  rpc SplitBlob(SplitBlobRequest) returns (SplitBlobResponse) {
```

tyler-french: Oops I missed Split, thanks

fmeum (approved):

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -554,10 +596,51 @@ service ContentAddressableStorage {
...
+  // * `NOT_FOUND`: At least one of the blob chunks is not present in the CAS.
+  // * `RESOURCE_EXHAUSTED`: There is insufficient disk quota to store the
+  //   spliced blob.
+  // * `INVALID_ARGUMENT`: The digest of the spliced blob is different from the
+  //   provided expected digest, OR the stream contains an invalid sequence of
+  //   splice requests.
+  // * `ALREADY_EXISTS`: The blob already exists in CAS. Clients can
+  //   call [SplitBlob][build.bazel.remote.execution.v2.ContentAddressableStorage.SplitBlob]
+  //   or [SplitChunks][build.bazel.remote.execution.v2.ContentAddressableStorage.SplitChunks]
+  //   to check what chunk mapping the server is using.
```
sluongng, line 623: missing

```
  //   ... and the server did not
  //   extend the lifetime of the chunks specified in the request, e.g. because
  //   it prefers a different chunking and extended those instead
```

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -496,10 +501,42 @@ service ContentAddressableStorage {
...
 
+  // SplitChunks is a streaming variant of SplitBlob.
+  //
+  // This RPC is equivalent to
+  // [ContentAddressableStorage.SplitBlob][build.bazel.remote.execution.v2.ContentAddressableStorage.SplitBlob],
+  // except that the list of chunk digests is streamed across response messages
+  // to avoid exceeding protocol message size limits. The complete list of
+  // chunks is the concatenation of `chunk_digests` across all responses in
+  // stream order. The server indicates that there are no more chunks by closing
+  // the response stream.
```
sluongng:
```
  // stream order. The server indicates that there are no more chunks by closing
  // the response stream.
```

but in SplitChunkResponse, we have

```
  // ... Clients MUST use the value
  // from the first response and ignore values sent on subsequent responses.
```

So we need to adjust the language a bit here to allow special cases where the first response is not available. (i.e., calling SplitChunks on a zero-size digest). OR we should mention that there will always be at least 1 response message (because a zero-size blob will be returned as 1 chunk).

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -554,10 +596,51 @@ service ContentAddressableStorage {
...
+  // * `RESOURCE_EXHAUSTED`: There is insufficient disk quota to store the
+  //   spliced blob.
+  // * `INVALID_ARGUMENT`: The digest of the spliced blob is different from the
+  //   provided expected digest, OR the stream contains an invalid sequence of
+  //   splice requests.
+  // * `ALREADY_EXISTS`: The blob already exists in CAS. Clients can
+  //   call [SplitBlob][build.bazel.remote.execution.v2.ContentAddressableStorage.SplitBlob]
+  //   or [SplitChunks][build.bazel.remote.execution.v2.ContentAddressableStorage.SplitChunks]
+  //   to check what chunk mapping the server is using.
+  rpc SpliceChunks(stream SpliceChunksRequest) returns (SpliceBlobResponse);
```
sluongng: could you please add the option google.api.http to these new RPCs?

That way they stay consistent with other rpcs in the file

sluongng (commented):
small doc nits, but lgtm

build/bazel/remote/execution/v2/remote_execution.proto
armandomontanez: > OR we should mention that there will always be at least 1 response message (because a zero-size blob will be returned as 1 chunk).

I'd advocate to reject a zero-size blob as `NOT_FOUND`. It's assumed to always be available via `FindMissingBlobs`, but I don't see any valuable reason to make `SplitChunks` expose a split version of the blob for which there is no split version. Similarly, it'd make sense for `SplitChunks` to respond to _any_ blob that is a single chunk as `NOT_FOUND`, since it's by definition not split.

build/bazel/remote/execution/v2/remote_execution.proto
tyler-french: This was intentionally removed in response to another review comment:
https://github.com/bazelbuild/remote-apis/pull/377#discussion_r3565315485
https://github.com/bazelbuild/remote-apis/pull/377#discussion_r3565311013

build/bazel/remote/execution/v2/remote_execution.proto
tyler-french: A zero-size blob is always available, but it has no split information, so NOT_FOUND is already allowed by the existing error definition. Successful calls therefore have at least one response. I don’t think additional wording is necessary.

build/bazel/remote/execution/v2/remote_execution.proto
tyler-french: Yes, good call.

build/bazel/remote/execution/v2/remote_execution.proto
tyler-french: I think that this is removed because the server behavior here is entirely unrelated to how the client interacts, and therefore is not critical to the API spec.

tyler-french:
Thinking about this more: if a client wants to start uploading and hashing at the same time, the blob hash shouldn't be required in the first request - it just needs to be in some request, or the last one. 

Routing isn't important here because the chunks will exist on many shards, so only the final frame needs routing to know where to store the verified entry. 

Going to update to account for this

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -2049,6 +2160,51 @@ message SpliceBlobRequest {
...
+  // The instance of the execution system to operate against. A server may
+  // support multiple instances of the execution system (with their own workers,
+  // storage, caches, etc.). The server MAY require use of this field to select
+  // between them in an implementation-defined fashion, otherwise it can be
+  // omitted. Servers MUST use the value from the first request and ignore
+  // values sent on subsequent requests.
+  string instance_name = 1;
+
+  // Expected digest of the spliced blob. Clients MUST set this on the first
+  // request. Servers MUST use the value from the first request and ignore values
```
vanja-p, line 2154: why must clients set the blob digest on the first request? This prevents the client from chunking, uploading (maybe with a findmissing call first), and splicing a blob in a streaming fashion.

tyler-french:
Just made in important change, that will further make this more performant:

Updated `SpliceChunks` so `blob_digest` is required only on the final request. A client may not know the full blob digest when it begins chunking, so this allows chunking, uploading, and incremental splice verification to proceed concurrently.

This does not weaken integrity: each chunk is content-addressed, and the server incrementally hashes the concatenated chunk contents in order, then compares the result with `blob_digest` when committing.

Earlier requests do not need the blob digest for routing because each chunk is routed independently by its own digest. Only the final commit needs `blob_digest` to validate and route storage of the resulting manifest.

build/bazel/remote/execution/v2/remote_execution.proto
tyler-french: Great idea, updated. Requiring this in the first frame would leave a lot of savings on the table.

vanja-p (approved):

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -496,10 +501,44 @@ service ContentAddressableStorage {
...
   //   chunks.
+  //
+  // New in v2.12 and removed in v2.13, use
+  // [SplitChunks][build.bazel.remote.execution.v2.ContentAddressableStorage.SplitChunks]
+  // instead.
   rpc SplitBlob(SplitBlobRequest) returns (SplitBlobResponse) {
     option (google.api.http) = { get: "/v2/{instance_name=**}/blobs/{blob_digest.hash}/{blob_digest.size_bytes}:splitBlob" };
   }
 
+  // SplitChunks is a streaming variant of SplitBlob.
```
EdSchouten: I think this is documented the wrong way around. Given that SplitBlob is deprecated, "SplitBlob is the deprecated non-streaming variant of SplitChunks"

Can we please write this in such a way that **this RPC** is authoritative, and that we don't refer to anything deprecated?

moroten:
> Updated `SpliceChunks` so `blob_digest` is required only on the final request. A client may not know the full blob digest when it begins chunking, so this allows chunking, uploading, and incremental splice verification to proceed concurrently.

How does the client know that it does need to upload the blob at all? I thought `FindMissingBlobs` would be called on the full blob before calling `SpliceChunks` but this indicates that the recommended process is to split the blob and stream the chunk digests. Before sending a chunk digest, it first must go through `FindMissingBlobs` for potential upload.

My original though was to first run `FindMissingBlobs` on the digest of the full blob before starting `SpliceChunks`.

Does calling many `FindMissingBlobs` really increase the efficiency compared to #376?

> Earlier requests do not need the blob digest for routing because each chunk is routed independently by its own digest. Only the final commit needs `blob_digest` to validate and route storage of the resulting manifest.

Well, that depends on the implementation of your server. Some servers might want to shard out the splicing process depending on the final `blob_digest`.

With the `blob_digest` up front, concurrent uploads can be aborted with `ALREADY_EXISTS` once the first upload finished.

tyler-french:
@moroten good points. I moved `blob_digest` back to the first request. You're right that clients should generally call FMB for the full blob first, and sending the digest up front gives servers more control over routing and early deduplication. It could also short circuit if it it already has the blob but FMB never got called.

moroten (approved):

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -496,22 +520,44 @@ service ContentAddressableStorage {
   //   reconstruct the blob is missing from the CAS.
   // * `RESOURCE_EXHAUSTED`: There is insufficient disk quota to store the blob
   //   chunks.
-  rpc SplitBlob(SplitBlobRequest) returns (SplitBlobResponse) {
-    option (google.api.http) = { get: "/v2/{instance_name=**}/blobs/{blob_digest.hash}/{blob_digest.size_bytes}:splitBlob" };
+  rpc SplitChunks(SplitChunksRequest) returns (stream SplitChunksResponse) {
```
meroton-benjamin: Could we rename this into `SplitIntoChunks` as SplitChunks doesn't really make any sense.

meroton-benjamin (changes requested):
Having implemented a client/server pair with this RPC I think we need to add guidance towards how this should be used to to upload a large blob. The naïve implementation, where you calculate a chunk, check if it exists, uploads it if it doesn't exist and sends a message with the chunk to the streaming rpc is incredibly inefficient and splicing is very intensive server side.

In essence a well behaved client should do the following for a large blob:

Step 1:

 * Calculate the blob digest and determine if the blob exists or not

Step 2:

 * Calculate a N of chunks of the blob where N is as many as can be comfortably held by the client, if calculating from a situation where you can do random access (e.g. when uploading a blob from disk) this may be something like 1MB worth of digests.
 * Determine which if any of the chunks are missing from the CAS with FMB
 * Upload those missing chunks
 * Repeat until all chunks exist in CAS

Step 3:

 * From the chunk list calculated earlier split into safe message sizes (e.g. below 1MB) and stream the splice chunks call.

While the api formally will work for clients which don't follow the above recipe, in practice server implementations will struggle to be able to serve it. A naïve implementation from above would from a very modest 1GB blob of random data create the equivalent of 10000 unary rpc calls.

build/bazel/remote/execution/v2/remote_execution.proto
armandomontanez: On this note: everyone I've discussed these RPCs with gets confused about the directionality of split/splice.

Maybe we could be more clear with something like `PutSplitBlob`/`GetSplitBlob` so directionality is clearer. Even if we want to be pedantic, `RegisterSplitBlob`/`RetrieveSplitBlob` are much clearer.

tyler-french: We could even go as simple as Get and Set 🤔

tyler-french:
@armandomontanez @meroton-benjamin per-request, I renamed the APIs: 
	- `GetChunkMapping`
 	- `RegisterChunkMapping`
 I think this makes the intention much more clear how to use these. 

@tjgq @fmeum @sluongng if you all also have time to take another look!

build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -2049,6 +2139,51 @@ message SpliceBlobRequest {
...
+  // An empty list is allowed in any request. It contributes no chunks to the
+  // assembled chunk list.
+  repeated Digest chunk_digests = 3;
+
+  // The digest function of all chunks to be concatenated and of the blob to be
+  // spliced. The server MUST use the same digest function for both cases.
+  // Servers MUST use the value from the first request and ignore values sent on
+  // subsequent requests.
+  //
+  // If the digest function used is one of MD5, MURMUR3, SHA1, SHA256, SHA384,
```
fmeum, line 2175: Symmetry is an argument for this, but perhaps we shouldn't carry these defaults forward to new APIs?

fmeum (approved):

build/bazel/remote/execution/v2/remote_execution.proto
tyler-french: We can do that, but I'd like that to be separate from this PR. If we want to still do it, we can make this change before a release, and focus the discussion there. This API is flexible until we publish a release

tyler-french:
@tjgq @sluongng friendly ping here - 

I think we should go ahead and try to get this merged. Leaving it open is causing a follow-up churn as we continue to iterate on CDC, and I want to avoid making this PR a grab-bag CDC related changes. Its in a good shape now, but I'd like any additional follow-ups to be raised by in separate PRs, to keep this one scoped.

We can wait to do a version release until we're confident in the shape of the spec, so merging the PR doesn't prevent changes to these unreleased APIs.

armandomontanez (approved):

build/bazel/remote/execution/v2/remote_execution.proto
EdSchouten: Indeed; this sentence should not be added to new RPCs. Also notice how the exact list of digest functions differs per RPC/field. Take `ExecuteRequest`:
```
  // If the digest function used is one of MD5, MURMUR3, SHA1, SHA256,
  // SHA384, SHA512, or VSO, the client MAY leave this field unset.
```
Now look at `ExecuteOperationMetadata`:
```
  // If the digest function used is one of BLAKE3, MD5, MURMUR3, SHA1,
  // SHA256, SHA256TREE, SHA384, SHA512, or VSO, the server MAY leave
  // this field unset.
```
We only list digest functions here that could be used in combination with those messages whose use predates the introduction of the `digest_function` field. Because `BLAKE3` was added AFTER `ExecuteRequest.digest_function` was added, it's never valid to use `BLAKE3` and leave `ExecuteRequest.digest_function` unset.

So by copying the phrasing from just one of these places you're doing something totally arbitrary. For a new RPC this should simply not be added. Symmetry is not a valid argument.

tyler-french: Fair, updated

tjgq (approved):

PR #378, opened 2026-05-27
Unbreak go_proto_library() targets

EdSchouten:
These reference repositories like @com_google_cloud_go_longrunning, which are currently only available as a dev dependency. Make sure that the use_repo() is called against a module extension proxy that doesn't have dev_dependency = True. This also requires making Gazelle a proper dependency.


MODULE.bazel
```diff
@@ -9,17 +9,16 @@ bazel_dep(name = "grpc", version = "1.73.1")
 bazel_dep(name = "rules_proto", version = "7.0.2")
 bazel_dep(name = "protobuf", version = "30.0")
 bazel_dep(name = "googleapis", version = "0.0.0-20260130-c0fcb356")
+
```
EdSchouten, line 12: Whitespace change caused by buildifier.

MODULE.bazel
```diff
@@ -9,17 +9,16 @@ bazel_dep(name = "grpc", version = "1.73.1")
...
-go_deps = use_extension(
-    "@gazelle//:extensions.bzl",
-    "go_deps",
-    dev_dependency = True,
-)
-go_deps.from_file(go_mod = "//:go.mod")
+go_deps_dev = use_extension("@gazelle//:extensions.bzl", "go_deps", dev_dependency = True)
+go_deps_dev.from_file(go_mod = "//:go.mod")
+
+go_deps = use_extension("@gazelle//:extensions.bzl", "go_deps")
```
fmeum, line 18: This will blow up if used by another module: the repos are used unconditionally but only instantiated as a dev dependency. It looks like there should only be one usage, which is non-dev?

EdSchouten: It won't blow up, as demonstrated by the fact that Buildbarn `main` already has this patch applied and CI passes.

If another Go based project were to depend on this, then it ships its own `go.mod` file that contains all of these repositories as indirect dependencies. This causes the repositories listed here to resolve properly.

fmeum: This is not guaranteed. For example, Bazel projects could use `go_deps.module` to bring in Go deps without a `go.mod`. 

This is definitely a clever way to avoid specifying all deps while still working with most setups. I don't think that's worth the cost though - couldn't we just specify all deps with minimal extra cost (a few transitive bazel_deps)? 

The proper fix would be to migrate all rulesets, in particular rules_go, to aspect-based proto rules. We can always make the `go_proto_library` targets aliases to the `proto_library` targets and drop all Go deps, leaving an appropriate nodep behind, when that's been done.

EdSchouten: Yeah, I agree that letting rules_go use aspect-based proto rules is the way to go. It's weird that it isn't already designed to work like that, given that Protobuf is the prime example of why you'd want to use aspects. Oh, well.

I've just changed this PR to also call `go_deps.from_file()` against a module extension proxy that doesn't have `dev_dependency = True`.

fmeum (approved):

PR #383, opened 2026-08-18
Update buildfarm links in README.md

Uthpal-p:
The current link for the Buildfarm Operation Queues doesnt exist and returns a 404. Fixed it with the right link.


sluongng (approved):
LGTM

cc: @werkt

